PR #4, opened 2017-01-16
Removed post type and other functions

ChrisChasm:


ChrisChasm:
Because I moved all the post types and other features to the plugin, if you run them both at the same time, you'll get a conflict. I believe this is a process of getting clarity regarding what needs to come from the plugin and what needs to come from the theme.

PR #34, opened 2017-08-15
Contacts List: Show name for dispatch || user || group

corsacca:
looks good?


burntsun:
This will not affect filtering, but it will affect the display of the rows and the search function.

The "Assigned to" filter pane will only display values where `user_login` is set. They should all work correctly even after this change.

PR #36, opened 2017-09-12
Group details

corsacca:


assets/js/group-details.js
```diff
@@ -0,0 +1,467 @@
+/* global jQuery:false, wpApiSettings:false */
+
+jQuery.ajaxSetup({
+  beforeSend: function(xhr) {
+    xhr.setRequestHeader('X-WP-Nonce', wpApiSettings.nonce);
+  },
+})
+
+jQuery(document).ready(function($) {
+function save_field_api(groupId, post_data){
```
burntsun: The lack of indentation here makes me nervous.

assets/js/group-details.js
```diff
@@ -0,0 +1,467 @@
...
+  })
+}
+
+function get_group(groupId){
+  return jQuery.ajax({
+    type:"GET",
+    contentType: "application/json; charset=utf-8",
+    dataType: "json",
+    url: wpApiSettings.root + 'dt-hooks/v1/group/'+ groupId
+  })
```
burntsun: I would change this to `return jQuery.ajax({ /* .... */ }).promise()` , just to make it clear that what is returned is a promise, although this should work as well.

assets/js/group-details.js
```diff
@@ -0,0 +1,467 @@
+/* global jQuery:false, wpApiSettings:false */
+
+jQuery.ajaxSetup({
+  beforeSend: function(xhr) {
+    xhr.setRequestHeader('X-WP-Nonce', wpApiSettings.nonce);
+  },
+})
```
burntsun: Note this has implications for all Javascript globally, even Javascript outside this file. This may or may not be a good thing.

assets/js/group-details.js
```diff
@@ -0,0 +1,467 @@
...
+ * Typeahead functions
+ */
+
+function add_typeahead_item(groupId, fieldId, val, name) {
+  add_item_to_field(groupId, { [fieldId]: val }).done(function (addedItem){
+    jQuery(`.${fieldId}-list`).append(`<li class="${addedItem.ID}">
+    <a href="${addedItem.permalink}">${_.escape(addedItem.post_title)}</a>
+    <button class="details-remove-button details-edit"
+            data-field="locations" data-id="${val}"
+            data-name="${name}"
```
burntsun: This needs escaping, you don't want to fail if `name` is `"><script src="http://example.com/malicious.js"></script>`

assets/js/group-details.js
```diff
@@ -0,0 +1,467 @@
...
+    editingAll = !editingAll
+  }
+  $('#edit-details').on('click', function () {
+    toggleEditAll()
+  })
+
+  $(document)
+    .on('click', '.details-remove-button', function () {
+      console.log("remove")
+    let fieldId = $(this).data('field')
```
burntsun: What's up with this indentation?

assets/js/group-details.js
```diff
@@ -0,0 +1,467 @@
...
+        commentData.comment.date = new Date(data.comment.comment_date)
+        comments.push(data.comment)
+        display_activity_comment()
+      })
+    })
+
+  $.when(
+
+
+  )
```
burntsun: `$.when()` returns a resolved `Promise`, but what's the point of this code?

assets/js/group-details.js
corsacca: @burntsun this file is only loaded on the group details page. I have the same code in contacts details and it did not apply to the group details.

corsacca: @burntsun  this file is only loaded on the group details page. I have the same code in contacts details and it did not apply to the group details.

assets/js/group-details.js
corsacca: @burntsun Kinda like falling off a cliff?

assets/js/group-details.js
burntsun: Yes, this file is only loaded on one page, but the code `xhr.setRequestHeader` applies to all jQuery AJAX calls made on that page, even in other scripts. It has global effects. I would prefer it if it only affected code within the file `group-details.js`.

assets/js/group-details.js
burntsun: Heh.

PR #37, opened 2017-09-13
Navigation & Page Additions

ChrisChasm:
Navigation update
Page template additions
Archive additions
Single additions
Logic addition for core pages
Developer README files added


corsacca:
image

@ChrisChasm I get this error on the locations page

corsacca:
Is each page in the menu supposed to be a placeholder? If they don't have any contact yet (mvp) can we put a message that says something like: `coming soon`. I'm getting 404 messages for some of them.

archive-locations.php
```diff
@@ -4,25 +4,101 @@
...
+                    'include' => '',
+                    'exclude' => '',
+                    'meta_key' => '',
+                    'meta_value' =>'',
+                    'post_type' => 'locations',
+                    'post_status' => 'draft, publish, future, pending, private',
+                    'suppress_filters' => true
+                );
+                
+                $recent_posts = wp_get_recent_posts( $args, ARRAY_A );
```
corsacca: What is the purpose of recent locations?
Locations are probably set up all at the same time. So should all have the post_date

archive-locations.php
```diff
@@ -4,25 +4,101 @@
...
+            $args = array(
+                'post_type' => 'locations',
+            
+            );
+            $query1 = new WP_Query( $args );
+            ?>
+            <?php if ( $query1->have_posts() ) : while ( $query1->have_posts() ) : $query1->the_post(); ?>
+                
+                <!-- To see additional archive styles, visit the /parts directory -->
+                <?php get_template_part( 'parts/loop', 'prayer' ); ?>
```
corsacca: It looks like this prayer loop inside locations needs to be renamed.

archive-locations.php
```diff
@@ -4,25 +4,101 @@
...
+    <div id="inner-content" class="grid-x grid-margin-x">
+        
+        <main id="main" class="large-8 medium-8 cell" role="main">
+            
+            <?php
+            $args = array(
+                'post_type' => 'locations',
+            
+            );
+            $query1 = new WP_Query( $args );
```
corsacca: Following contacts and groups this code would be in the locations.php.
example `$locations = Disciple_Tools_Locations::get_locations();`
Note: the current get_locations does not support paging. For conversation: if we are expecting 100s or more locations we might need to update code in contacts-list

archive-locations.php
```diff
@@ -4,25 +4,101 @@
...
+                    'format'          => 'html',
+                    'before'          => '',
+                    'after'           => '',
+                    'show_post_count' => false,
+                    'echo'            => 1,
+                    'order'           => 'DESC',
+                    'post_type'     => 'locations'
+                );
+                wp_get_archives( $args );
+
```
corsacca: same here

archive-peoplegroups.php
```diff
@@ -0,0 +1,106 @@
...
+                    'suppress_filters' => true
+                );
+                
+                $recent_posts = wp_get_recent_posts( $args, ARRAY_A );
+                
+                echo '<ul>';
+                ?>
+                <?php foreach ($recent_posts as $recent_post): ?>
+                    <li><a href="<?php echo $recent_post['guid'] ?>"><?php echo $recent_post['post_title'] ?></a></li>
+                <?php endforeach; ?>
```
corsacca: the guid should not be used as a link to the post. 
https://developer.wordpress.org/reference/functions/the_guid/
try get_permalink() instead

ChrisChasm: Actually, that was too much of a copy paste. I'll clean it up.

ChrisChasm:
I'll pull this back and work on it a bit more.

burntsun:
I've removed myself from reviewers, since @corsacca has reviewed it and @ChrisChasm closed this pull request.

PR #59, opened 2017-09-27
Escape Corrections, Added Locations, Added Teams/Members List

ChrisChasm:
Escape Corrections, Added Locations, Added Teams/Members List to the template-settings.php file


ChrisChasm:
These edits bring to completion the notifications section of the settings page. It includes site options, api interaction, section template, and endpoints/controller functions.

PR #60, opened 2017-09-28
Added user availability option

ChrisChasm:
template, js api, endpoints, controller, option checker/creator.

### DIFF
--- a/template-settings.php
+++ b/template-settings.php
@@ -351,13 +351,12 @@ dt_print_breadcrumbs(
             
             </div> <!-- End Notifications -->
             
-            
+            <?php
+            $dt_availability = dt_get_user_option_availability( get_current_user_id() );
+            ?>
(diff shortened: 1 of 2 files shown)


template-settings.php
```diff
@@ -376,7 +375,8 @@
                 </div>
                 
                 
-                <!-- List of past, present, and future vacations scheduled -->
+                <?php // TODO: Add scheduling and history of availability ?>
+                <!-- List of past, present, and future vacations scheduled
```
burntsun: This comment tag is not closed.

ChrisChasm: Yes. I commented out this section for future development. This is part of the conversation this morning, where we are pausing on the scheduling and history part of the availability feature, for possibly after Nov.

burntsun: That makes sense. It just seemed odd to me that the comment tag is not closed.

PR #437, opened 2018-02-28
Remove some borders on the contact view page

grantholle:
Removes the borders in the comment and activity section, as well as the border around show more/show less on the details card.


corsacca:
@grantholle Thanks! looks awesome!

PR #438, opened 2018-02-28
Fix #433

### LINKED ISSUE #433: Contacts: status is not updating when the contact assignment changes
When assigning a contact the status should automatically change to assigned without needing to refresh the page.

grantholle:
I hate these types of bugs 😡


corsacca:
Great catch!

PR #441, opened 2018-03-02
Limit the active navigation to only the left nav, involves PR #440

grantholle:
Qualify the navigation only to the left side, we don't want the settings getting those active styles


corsacca:
Looks good!

PR #443, opened 2018-03-02
Fix #435, refactor how the location endpoints are generated

### LINKED ISSUE #435: Locations rest endpoint url is constructed in two places.

grantholle:


corsacca:
@grantholle great array idea

PR #448, opened 2018-03-06
Replace google hosted resources with cdnjs, fix #444

### LINKED ISSUE #444: Consider using resources not hosted by Google
While there are obvious benefits of using resources from a CDN, where I am, I cannot load the page because it's trying to load jquery from Google. Something to consider

grantholle:


corsacca:
That was quick :)

PR #453, opened 2018-03-15
Major refactoring of contact details partial

grantholle:
I did a lot of cleaning up in the php partial and the corresponding js for the contact details partial.

I noticed the UI was really inconsistent and the template code needed a little love. I also refactored the js that saves the contact details, making things a little more consistent. I think everything is working, but I may have missed something. Let me know. It's a big one 😛


corsacca:
@grantholle great improvements!
Can you have a look at the travis checks? Some edits aren't making our phpcs rules happy.

grantholle:
This is good now

corsacca:
@grantholle thank you.
what is you reasoning for the single column in the show more?
image

grantholle:
I think that was by accident.

grantholle:
The catch is that the industry uses flexbox for grids (`display: flex`). When hiding/showing things with jquery (`.toggle()`, `.hide()`, `.show()`), it toggles between `display: none` and `display: block`. When it's `block` display, it breaks the grid. Which is what happened here.

PR #582, opened 2018-06-20
SETTINGS: took out delete button for default seeker path options

TheEvilTopHat:


dt-core/admin/menu/tabs/tab-custom-lists.php
```diff
@@ -531,6 +531,15 @@ public function process_seeker_path_box()
      */
     public function seeker_path_box()
     {
+        $default = [
```
corsacca: user the default list in `config-site-defaults`

PR #603, opened 2018-07-18
Change email subject to always be "Update..."

burntsun:
This will make the emails group better in Gmail's web interface.

### DIFF
--- a/dt-groups/groups.php
+++ b/dt-groups/groups.php
@@ -212,7 +212,7 @@ class Disciple_Tools_Groups extends Disciple_Tools_Posts
                                 $user = get_user_by( 'id', $id );
                                 if ( $user ) {
                                     $fields[ $key ] = [
-                                        "ID"          => (int) $id,
+                                        "id"          => (int) $id,
                                         "type"        => $type,
                                         "display"     => $user->display_name,
                                         "assigned-to" => $value[0],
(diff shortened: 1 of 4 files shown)


corsacca:
@burntsun do you suggest this just for comment notifications or for all type of notifications? 
Milestone updates, update needed, assignment changes etc all have their own subject line currently:
https://github.com/DiscipleTools/disciple-tools-theme/blob/master/dt-notifications/notifications.php

burntsun:
I think that everything related to a single contact record should be grouped in the same conversation on Gmail. The easiest way to do that is to give it the same subject line. I didn't see those other subject lines, I can go ahead and modify that code too if you approve.

corsacca:
Approved. Your suggestion is a good one. We might revisit it when we do some kind of notification grouping before we send out an email. 
We do loose a bit of information by making all subject lines for a contact the same.

burntsun:
@corsacca OK, in my new commits, I've created a new function `dt_send_email_about_contact`, which makes sure the same subject line is always used, and that a link is always included in the footer of the email body.

corsacca:
@burntsun thanks!

corsacca:
Is it the same for a group? I'm going to test the changes locally.

burntsun:
To be honest, I didn't think about how these changes affects groups.

burntsun:
Nice, I like this teamwork!

PR #620, opened 2018-08-24
Merge tools

buildittheywillcome:
Here is the updated work on Merge Tools.  We are mostly passing your standards test.  There are a few things that we may need some direction on if they do not pass properly.  Let me know what else you need from us.


corsacca:
Thanks @buildittheywillcome! I'm merging your changes and will check out what still is failing the standards test

PR #622, opened 2018-08-29
removing preg_grep

buildittheywillcome:
Fix to preg_grep error


corsacca:
@buildittheywillcome thanks

PR #644, opened 2018-10-24
Authorize API by site link token

cairocoder01:


corsacca:
@cairocoder01 the header authentication piece looks great!
For the extra capabilities i would suggest moving those into a plugin of your own.
There you can use the `site_link_type` and `site_link_type_capabilities` hooks to add an extra option in the site-to-site connection type list. This way you can modify the capabilities of the link as your project grows without having to update D.T itself.

cairocoder01:
@corsacca That makes sense about the added capabilities. Would you want any of the basic contact capabilities kept in there, since some of that was part of the public API access? Or will you just leave those public endpoints in place for those features?

cairocoder01:
@corsacca I updated the code to remove the added permissions. 

I'm not sure why the TravisCI build failed, and I can't trigger it to run again. Seems to be an issue with the gulps scripts task, but I didn't change any javascript.

corsacca:
@cairocoder01 Thanks!
This `contact sharing` option is meant for just sending contact to another system.
I'm thinking we add a default `contact access` option  with most of the contact permissions.

PR #645, opened 2018-10-31
Fix setting of activity user and gravatar

cairocoder01:
API for contact activity (/contacts/{id}/activity) was not correctly setting name and gravatar properties.

### DIFF
--- a/dt-core/posts.php
+++ b/dt-core/posts.php
@@ -517,8 +517,8 @@ class Disciple_Tools_Posts
             if ( isset( $a->user_id ) && $a->user_id > 0 ) {
                 $user = get_user_by( "id", $a->user_id );
                 if ( $user ){
-                    $a->name =$user->display_name;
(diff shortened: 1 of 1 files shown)


corsacca:
@cairocoder01 yeah the gravatars aren't always showing up. Tho this didn't fix it.
The $user fields are accessed directly all throughout the wp codebase.

corsacca:
Does this fix it for you? https://github.com/DiscipleTools/disciple-tools-theme/commit/f14adddadf4fc8541884a2839149ecb0ed70e75e

cairocoder01:
The bigger problem than the gravatar was the name. I thought they were both missing, but somehow I looked at it now after reverting code to an old point, and it seems to work fine. I think I got mixed up looking at system comments/activity that don't have name/gravatar. I did find an error in my own code that might have been the cause of me thinking something was broken. I'll close this PR since everything looks like it is working.

PR #653, opened 2018-11-26
Spanish translation update and replacement of english words

mikeallbutt:
Here is the current Spanish translation update and replacement of english words that were previously not being switched into other languages.


dt-core/admin/menu/tabs/tab-locations.php
```diff
@@ -268,20 +268,20 @@ public function import_simple_list_of_locations() {
         <form method="post" action="">
             <?php wp_nonce_field( 'dt_simple_import_levels'. get_current_user_id(), 'dt_simple_import_levels_nonce', false, true ) ?>
             <tr class="dt_simple_import_levels">
-                <td>
+                <td><!-- !@#$ text below needs replacing -->
```
corsacca: @mikeallbutt Here you can leave an "@todo" instead of !@#$

mikeallbutt:
> @mikeallbutt have a look at that failed travis test. There is one line that is off. Just a simple style edit. Can you fix it?

I would like to fix it but I don't know what the issue is. Please guide me as to what to do as I want to learn.

PR #672, opened 2019-03-07
Made the mobile top bar work more correctly

Noneofyall:
Changed the mobile search icon from black to white and added the "Create New" dropdown next to notifications and settings on mobile.


corsacca:
Thanks @Noneofyall!

PR #677, opened 2019-03-18
JS Cleanup

grantholle:
Mainly add a global function to make a request via jquery, `makeRequest()`, rather than calling it with the same general parameters. Refactors the `window.API` object and other places to implement this function.


corsacca:
@grantholle I like the clean look! Thanks.
There as still some instances of the misspelled `handelAjaxError` which are throwing errors.

grantholle:
They should all be replaced now.

corsacca:
Awesome!

PR #687, opened 2019-04-16
Seem to have fixed load more notifications bug, as well as the bug in…

Noneofyall:
… notifications where clicking on the page you are already in clears the page


corsacca:
Thanks @Noneofyall!

PR #692, opened 2019-05-08
change sources display name, added new sources for viber, whatsapp, a…

artenator:
Update "Sources" to be labelled as "How did you meet?" based on feedback from user interviews.

This is my first PR, and I'm pretty new to the codebase.

let me know what you guys think! 😎

### DIFF
--- a/dt-assets/parts/contact-details.php
+++ b/dt-assets/parts/contact-details.php
@@ -256,7 +256,7 @@
             <div class="xlarge-4 large-6 medium-6 small-12 cell">
                 <div class="section-subheader">
                     <img src="<?php echo esc_url( get_template_directory_uri() ) . '/dt-assets/images/source.svg' ?>">
-                    <?php esc_html_e( 'Source' ); ?>
+                    <?php esc_html_e( 'How did you meet?' ); ?>
                 </div>
                 <ul class="sources-list <?php echo esc_html( user_can( get_current_user_id(), 'view_any_contacts' ) ? 'details-list' : '' ) ?>">
                     <?php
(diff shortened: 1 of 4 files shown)


corsacca:
Hey @artenator thanks for your commit! We are exited to see more contributions.

Can you share a bit about the reason for the change? Sources is specifically a generic term. A contact can come many different sources like an in person encounter, a website, a facebook page or an event sign up sheet. "How did you meet?" might be confusing as it implies a meeting has already occurred.

For completeness of the label change you'd also want to update the field definition in the contacts-post-type.php file (line 411). Or through the hook: `dt_custom_fields_settings`. The name shows up in the filters on the contacts list and other places.

ChrisChasm:
Hey @artenator, great to have you contributing! We look forward to much more! 

On the name change of "Sources", I am with @corsacca . I think maybe your user feedback was not thinking widely enough about the multiple kinds of groups and outreaches that are collecting contacts from more than just "meetings". There are a number of media and event driven outreaches that would be confused by the "How did you meet?" question, rather than a more generic label for "Source". 

One other way to make term change, which we expected specific groups wanting to do, is to develop a new translation file. You can set the site to select a localization of english (or any other language). We don't have any active cases for this, but in our designing we expected there to be a bunch of language preferences between different organizations. But the potential for an organization to make a gettext (.po) file that customizes the phrasing and term choices to match their "insider" language for movement activity is definitely there. You could install it directly or though an organization specific "plugin".

This idea of working from an organization plugin is really a good step. This way you can cater to the organization your are serving for the customization, and then if you have core tweaks that benefit all system users, then those changes can be pushed onto the core upgrade track. 

When it comes to user focus groups and trying to build for the 80% majority, we have room to grow. Any research you are gaining, we would love to have access to. Our current understanding of user requirements is coming from our own focus groups, the known user community of groups implementing the software and actively interacting with us, and leadership from various movements who are speaking into the main things they need in multi-generational disciple making movements. 

One think I would encourage with your user interviews is to make sure you are interviewing leaders and teams running outreach coalitions and church planting movements. This is the core group the software is designed to serve. It is at its heart a disciple making movement software. If someone does not esteem this kind of multiplicative ministry vision, they will likely be confused over many parts of this system, and ultimately it won't be a good fit for them.

PR #699, opened 2019-05-14
Made contact get greyed-out on the drop down contacts menue for bapti…

Noneofyall:
…zed, baptized by, etc if it's status is closed


corsacca:
Cool!! Thanks @Noneofyall

corsacca:
image

Issue #669, opened 2019-02-20
Always available advanced global search feature.


corsacca:
Have a search icon available in the tap bar from anywhere on the site.
This will open a modal/overlay that lets you quickly search. You need to choose is you are searching in contacts or in groups. 
Needs to be available on desktop and on mobile.

corsacca:
Something like this:
![screenshot from 2019-02-20 15-32-39](https://user-images.githubusercontent.com/24901539/53099139-7127b300-3525-11e9-81b7-8854b2770c09.png)

corsacca:
Include where to search:
- everywhere
- comments
- texts fields 
- Contact info fields
etc

corsacca:
Search in code for  `dt_search_extra_post_meta_fields` filter for current search implementation

grantholle:
Will try to take this on

corsacca:
Awesome! :)

grantholle:
So you want the user to be able to specify the area in which to search?

An alternative I am thinking is just input that searches everything, then organizes the results into the different types (groups/contacts) based on results that were found anywhere

corsacca:
I like this idea. It could be more powerful. It might require we build a search api that returns the correct info.
It will be best if the organization distinguishes where the result came from. Something like this?
Contacts > Comments
- contact 1
- contact 2
Contacts > Custom Text Field
- contact 2
Groups > Comments.
- group 1

grantholle:
Yeah that's kind of what I was thinking

mikeallbutt:
Looking forward to having this available. Currently if I am viewing a group, I have to click to the lest of contacts then do a search to find a contact.

corsacca:
This issue has been started, but work has been paused. Anyone can continue from the pull request: https://github.com/DiscipleTools/disciple-tools-theme/pull/727

mikeallbutt:
This is a feature that we really need. It is beyond my current abilities, but would give it a go if I could. I really hope someone (@blachawk?) with the right skills can get this completed asap.

corsacca:
Icon opens the search modal.
The user types anything in the search modal
This could search the post names, post meta and comments for results
Have options to narrow the search on the side (exclude comments, only search contacts etc)
List out the results including name, post type and matched field.
Make it look awesome! :)

We'll need to build out a separate endpoint for this global search since the current ones rely on the post type being specified.

micahmills:
[...]

PR #727, opened 2019-06-25
Globally available search

grantholle:
#669, I wasn't able to complete the api side. The front end is ready to go. It needs to make the finished api call and then changing the boilerplate to be the real results.

Sorry to leave this half finished 😔. I would rather submit what I have currently have and hopefully someone can finish what I've started...


corsacca:
Merged changes into branch: https://github.com/DiscipleTools/disciple-tools-theme/tree/global_search

PR #735, opened 2019-07-11
Delete post metadata value along with details (from API)

cairocoder01:
When attempting to delete a phone number from a contact via the API, I found it was not being deleted. 

POST to `/wp-json/dt-posts/v2/contacts/{id}`
Body:
```
{
  "contact_phone": [{
    "delete": true,
    "key": "contact_phone_123"
  }]
}
```

**Expected**
Should delete phone number so that it is no longer returned from the API. In the database, it should delete metadata records for `contact_phone_123` and `contact_phone_123_details`

**Actual**
Phone number is still returned from the API. In the database, only `contact_phone_123_details` was deleted.

**Fix Description**
When deleting post metadata, this update includes the deletion of both metadata records so that the value is completely deleted.

### DIFF
--- a/dt-posts/posts.php
+++ b/dt-posts/posts.php
@@ -1123,7 +1123,14 @@ class Disciple_Tools_Posts
                         return new WP_Error( __FUNCTION__, "missing key on: " . $details_key );
                     }
                     //delete field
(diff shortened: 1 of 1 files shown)


corsacca:
Thanks @cairocoder01

PR #760, opened 2019-09-06
Zume mapping branch

ChrisChasm:
This branch re-conforms and refactors the dt-mapping folder:
- Adds a global variable for all the urls and path
- Centralizes the load in the loader.php, instead of having files required throughout the folder.
- Adds ipstack, mapbox, and google geocoding classes
- Relocates the location-grid-geocoder.php
- Upgrades the mapping.js with better URLs
- Adds optional caching transient layer to the REST api that can be called or not called by the application.


dt-core/global-functions.php
```diff
@@ -123,50 +127,68 @@ function dt_array_to_sql( $values ) {
...
+    if ( ! function_exists( 'dt_date_end_of_year' ) ) {
+        function dt_sanitize_array_html( $array) {
+            array_walk_recursive($array, function ( &$v) {
+                $v = filter_var( trim( $v ), FILTER_SANITIZE_STRING, FILTER_FLAG_NO_ENCODE_QUOTES );
+            });
+            return $array;
+        }
     }
 
-    /**
```
corsacca: @ChrisChasm Did you mean to delete this function? (dt_array_to_sql already is declared further up in this file.

PR #767, opened 2019-09-14
Fix enqueue scripts bug when using a child theme

bradwellsb:
get_stylesheet_directory_uri() uses the child theme directory if a child theme is active.  This is not the desired behavior when actually using a child theme.  It results in metrics-personal.js not being properly enqueued, and the Metrics page fails to load.  We should load metrics-personal.js from the parent theme by using get_template_directory_uri() instead to ensure compatibility.

### DIFF
--- a/dt-assets/js/shared-functions.js
+++ b/dt-assets/js/shared-functions.js
@@ -10,7 +10,7 @@ jQuery(document).ready(function($) {
     } else {
         ref = window.location.pathname
     }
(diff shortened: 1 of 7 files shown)


bradwellsb:
Issue is duplicated throughout Metrics pages.  Added commit to fix in other places

corsacca:
Thanks @bradwellsb, these commits are really helpful!

corsacca:
I suppose you are right about changing the rest of the `get_stylesheet_directory_uri()` occurrences. 
It looks like you have already worked on it some, shall we wait till you are able to commit those?

PR #769, opened 2019-09-17
Completing child-theme compatibility patch

bradwellsb:
Patch is for child theme compatibility. Replaces occurrences of get_stylesheet_directory_uri() with get_template_directory_uri() when used to reference critical scripts or files found in parent directory.


corsacca:
Thank you @bradwellsb!

PR #778, opened 2019-09-26
Mapping: Mapping permission fix

ChrisChasm:
Changed the capability from 'read_location' to 'view_mapping……'. Also, added a permission check that will be present even if loaded via REST.


corsacca:
Thanks @ChrisChasm !

Issue #730, opened 2019-06-28
Plugin to customize the theme's appearance and branding.


corsacca:
**Purpose**
Customize the logo, styles and colors of the theme. 

**Use case**
- A team or an org wants to make D.T feel and look more like other products they currently use.
- For more security aware contexts: changing the logo to make the theme less easily identifiable. A passer by seeing the name or logo and then googling it.

**Requirements**
- A plugin with settings pages in the wp-admin that allows a D.T admin to customize the appearance.
- customize the logo in the nav bar
- customize the logo on the login page (not displayed by default)
- change primary color and secondary colors.

**Code considerations**
In many cases the colors are currently hard coded. These need to be refactored to a css class or set of classes.

**Deeper security**
This issue only pertains to the surface appearance of the theme. A look at the html and css will reveal that it is a wordpress site using disciple-tools-theme. This is discussed more here: https://github.com/DiscipleTools/disciple-tools-theme/issues/615

mikeallbutt:
The developing/sponsoring of this plugin could be achieved with the help of other partners who want to user DT but with their own design/skin.

prykon:
@corsacca I can tackle this one if you want.

PR #780, opened 2019-10-02
Remove background fill from bell.svg

bradwellsb:
Updating bell.svg to be consistent with other vector graphics in .top-bar-right.  Removing hard-coded blue background fill.  Vector is now solid white, so it can be used without styling issues if .top-bar background-color is changed from the default #3f729b.

Previous behavior and new behavior demonstrated in attached image.
bell-bug


corsacca:
Thanks @bradwellsb, that does look better!
Are you working on something similar to https://github.com/DiscipleTools/disciple-tools-theme/issues/730? It would be of interest to several people.
Would you up for submitting a pull request without the changes to the 29 other files? It looks like you pulled in the changed code instead of the commits.

bradwellsb:
> Would you up for submitting a pull request without the changes to the 29 other files? It looks like you pulled in the changed code instead of the commits.
Fixed. PR should reflect changes to just the one file now.

corsacca:
Thanks @bradwellsb!

PR #791, opened 2019-10-20
add edit and help icons

mikeallbutt:


corsacca:
@mikeallbutt what are these icons for?
Why replace the current help icon with another one?

mikeallbutt:
> @mikeallbutt what are these icons for?
> Why replace the current help icon with another one?

The different color of the help icon is now the same as the general help icon and by having a different color to the normal blue that is used in many other places around DT, should help it stand out more and make it easier to tell users to look for this color icon.

mikeallbutt:
> @mikeallbutt what are these icons for?
These new icons can be used in the DT app and the website to help users know whether the add button is for a contact or a group.

corsacca:
@mikeallbutt we'd prefer to keep the icon blue as it is. Can you rename it back and the green one to -green.svg?

mikeallbutt:
> @mikeallbutt we'd prefer to keep the icon blue as it is. Can you rename it back and the green one to -green.svg?

Is there a way to just not use my changed/suggested file?

corsacca:
Pull requests are accept all or nothing. A solution is a separate PR for each change/feature.

PR #804, opened 2019-10-28
Icons & Help modals: Create New Contact and Group

mikeallbutt:
## A contact page with icons
Screenshot 2019-10-28 00 40 08

## A group page with the icons.
Screenshot 2019-10-28 00 37 43

Screenshot 2019-10-28 00 37 26


corsacca:
@mikeallbutt there are 3 big things happening in this request, which would be better as separate requests.
- Icon changes in the contacts and groups records
- Changes on the create group and contact pages
- Help modal improvements.

1. Contacts: I prefer the old home icon, and i'd like to see if we can get a better status icon
Group: I like the parent, child and peer group icons. Can we find a better group type icon? Could the Church start date have a clock next to a church and something similar for the group start date?

2. I like having the help icons and modals, though some don't load on the create contact.
image

3: Awesome. We'll want to make this dynamic down the road so it can reflex custom changes.

Thanks man!

mikeallbutt:
> @mikeallbutt there are 3 big things happening in this request, which would be better as separate requests.
> 
> * Icon changes in the contacts and groups records
> * Changes on the create group and contact pages
> * Help modal improvements.
> 
> 1. Contacts: I prefer the old home icon, 
> 
Ok. Address is different to location - so I tried to join the house with the place maker. By having a more specific icon for address, helps to advise that this is the exact address - which is different to location. 

> and I'd like to see if we can get a better status icon
> 
Ok. That one (I came up with is) is supposed to represent traffic lights.


>    Group: I like the parent, child and peer group icons. 
> 
Ok. (so do I.)


> Can we find a better group type icon? Could the Church start date have a clock next to a church and something similar for the group start date?
> 

Yes we could. I used the calendar icon instead of time one so that it could have other symbols placed over it (like the + and tick and -). 
The plus symbol because it was when the group was initially started; 
Church uses the tick because it reflects what they are trying to achieve. 
End date uses the X because that's when the group ended.

> 2. I like having the help icons and modals, though some don't load on the create contact.
>    image
> 

I have modified the modal-help.php file to add those 3 missing modals. (The text was there but they were in a different branch and would have appeared in the next PR that is awaiting approval.)

> 
> 3: Awesome. We'll want to make this dynamic down the road so it can reflex custom changes.
> 

I agree. There are other things I would like to automate and reduce replication ie. the add icon menu list is in 2 places - one for normal view and the other for mobile view. It would be great to have the 2 views use the same list.

> 
> Thanks man!

Your welcome. :)

corsacca:
@mikeallbutt Please revert the address icon. By adding the location symbol to the house it becomes less clear that it is different to the location field since the symbol is in both icons.
- The calendar icons for the group dates should be more visibly calendars (and consistent with each other)
- How about something like this for group type: https://www.iconfinder.com/icons/2578470/ask_group_question_team_teamwork_icon
- Status icon.

mikeallbutt:
> @mikeallbutt Please revert the address icon. By adding the location symbol to the house it becomes less clear that it is different to the location field since the symbol is in both icons.

I have modified the house by removing the door but keeping the location symbol. Will that work or do you still have the house (to be the icon for the address for where a group meets and the address for a contact)?

> 
> * The calendar icons for the group dates should be more visibly calendars (and consistent with each other)

Those other ones were actually what is used for the quick actions icons. And yes one of the them is slightly different - but that is what is in the figma file too.

> * How about something like this for group type: https://www.iconfinder.com/icons/2578470/ask_group_question_team_teamwork_icon

Not too keen on that because when the icons are small (15px) then fine details like what you suggested won't work.

> * Status icon.
This is a blend of a stop sign (inactive status) and tick (to represent active).

How about these versions? A cleaned up address, new status, 3 date and 4 group icons.

*Contact details - new status and modified address*
Screenshot 2019-10-31 00 50 40


*Group details  - new status and modified address*
Screenshot 2019-10-31 00 52 01


*Group icons*
Screenshot 2019-10-31 01 00 58

I think these are far clear as to the meanings and standout from each other significantly.

corsacca:
Thanks mike

Issue #773, opened 2019-09-20
Open or Close sidebar lists


mikeallbutt:
In the sidebar, have the ability to open or close the list by clicking the + or - symbol.

![image](https://user-images.githubusercontent.com/43966676/65358854-5226f780-dbc0-11e9-9e8b-ae4d9d5ecf9f.png)

mikeallbutt:
@corsacca if you give me some pointers on where to look in the code and what needs to be done, I would like to have a go at fixing this issue.

corsacca:
@mikeallbutt we use foundations accordion: https://foundation.zurb.com/sites/docs/accordion.html
I'm not confident there is a good solution without some fancy js or css.

blachawk:
This request appears to be working within my localhost environment. I can click on the + and - signs to open / close these panels within the Contact Filters section.

corsacca:
The end result for this on is the ability to close all the accordion list. Right now there is always one open

blachawk:
I can look into it if its not already assigned to someone.

corsacca:
@blachawk You are welcome to.

blachawk:
@corsacca requesting to close this issue, as the feature has been recently merged into the master.

corsacca:
Thanks @blachawk

corsacca:
@blachawk The hoped for behavior:
- When the page loads one of the accordions is open (and highlighted in blue).
- The user can minimize the open accordion so that all of them are closed.
- Only one accordion can be open at a time.
Currently all the accordions can't be close at the same time. @blachawk would you be up for looking at this again. Or thoughts on feasibility?

corsacca:
From Mike: https://github.com/DiscipleTools/disciple-tools-theme/commit/01d6ef0d3cd798f5ccc3593b27134a607e0b6864#diff-dd479bebdc6ce11aeda3b3e32b5468aeR1216

corsacca:
This is working now. Thanks guys

PR #824, opened 2019-11-15
Accordion improvement - Content Filter

blachawk:
Attempting to improve accordion flexibility related to issue #773


corsacca:
You could put the code in the list.js so it is only loaded on the list page.
The code expect there to just be one open at a time. Here the "All contacts" values get overwritten with 0s when i use filters in the subassigned section
image

Can the default opened section still be blue?

blachawk:
Will do....

blachawk:
The latest push to now have all of my custom code placed in list.js
It also has the multiExpand option set to false for this specific Accordion so that internal logic works properly.

Let me now if anything more is needed.

corsacca:
Hey, 
I am still getting the same issue with multiple lists open at the same time.
When the page loads, can we have the the active menu still be blue?

blachawk:
@corsacca, for cache purposes, please test this update in private mode with Firefox or Incognito mode with Chrome,  Thanks

corsacca:
Thanks @blachawk!

mikeallbutt:
Thanks for working on this @blachawk. The main reason for this issue was actually that it was desired to be able to allow all to be closed - which is especially helpful when using DT on mobile and when a user has a long list of Custom Filters. 

If that mod was implemented, that would be awesome.

PR #822, opened 2019-11-13
Corrected watch coding through gulp

blachawk:
Ready for testing my fix for watching (with and without BrowserSync) via gulpfile.js

PR #828, opened 2019-11-16
Browser sync

blachawk:
Attempting to improve the experience by... 
* Allowing developers to set their proxy from the command line with Browser-sync.
* Cleaned up some code in the gulp area
* This enhancement connects to #822


blachawk:
This browser-sync update now allows users to process the following from the command line. (i.e.)
```powershell
gulp browsersync-p --option http://my-local-url
```

Let me now if anything more is needed

corsacca:
@blachawk why are we ignoring the `dt-assets/build` folder in the .ignore?
When the .scss files are changed, they are minified into the build folded and then used in the theme.
See `dt-assets/functions/enqueue-scripts.php` line 87 and 90.

blachawk:
That was an overlook.  This issue is now corrected.

corsacca:
Thanks @blachawk

PR #834, opened 2019-11-22
Notifications page help & New contacts & New groups update

mikeallbutt:
Added a help icon to access a modal on the notifications page.
Screenshot 2019-11-22 10 45 24

And fixed the missing info for all these ? icons.
**Before**
Screenshot 2019-11-22 11 53 27
**After**
Screenshot 2019-11-22 11 54 02


dt-assets/scss/_main.scss
```diff
@@ -672,6 +668,10 @@ a.dt-green:hover {
     }
 }
 
+div.button-group {
```
corsacca: What is this edit for? It applies to the whole site

mikeallbutt: It is there to fix the spacing between the All and Unread buttons, and can prevent the same thing happening elsewhere.
Before:
Screenshot 2019-11-26 11 33 43
After:
Screenshot 2019-11-26 11 36 15

mikeallbutt: The same styling could be used to fix these grouped buttons too as they are touching. 
IMO Removing the inline style (that is coded into the html) and moving it to the css file would be an option.

Screenshot 2019-11-26 11 45 30

corsacca: The button-group class is used in several places in the site (ie milestones). Each place will need it's own css.
For the notification buttons please use specific css instead of general.

mikeallbutt: > The button-group class is used in several places in the site (ie milestones). Each place will need it's own css.
> For the notification buttons please use specific css instead of general.

Ok. So do you suggest I write CSS to be location specific - just to be used on the notification page? So adjust the CSS file would be the cleanest IMO.

corsacca: The scss code you wrote applies to tho whole theme, maybe in ways you won't expect.
Either you change it to work for every situation, or you make your selectors more specific.

template-view-duplicates.php
```diff
@@ -13,20 +13,23 @@
...
         <div id="inner-content" class="grid-x grid-margin-x">
 
             <main id="main" class="large-12 medium-12 cell" role="main">
                 <div class="bordered-box">
-                    <h3><?php esc_html_e( 'Duplicate Contacts', 'zume' ) ?></h3>
+                    <h3><?php esc_html_e( 'Duplicate Contacts', 'zume' ) ?><button class="help-button float-right" data-section="duplicates-template-help-text">
+                        <img class="help-icon" src="<?php echo esc_html( get_template_directory_uri() . '/dt-assets/images/help.svg' ) ?>"/>
+                    </button></h3>
 
+                    <?php //@// TODO: use an if statement to show a message if not duplicates were found. Else show the following table ?>
```
corsacca: Can you give this a try?

mikeallbutt: No. Doing that is beyond me at present. Sorry.

template-view-duplicates.php
```diff
@@ -13,20 +13,23 @@
...
                     <table id='table-duplicates'>
                         <?php
                         foreach ( $dt_duplicates as $dt_id => $dt_duplicate ) {
                             if ($dt_duplicate['count'] <= 0) { continue; }
                             echo "<form name='merge_".esc_html( $dt_id )."' method='POST' action='".esc_html( site_url() )."/contacts/".esc_html( $dt_id )."'><input type='hidden' name='dt_contact_nonce' value='".esc_attr( wp_create_nonce() )."'/></form><tr id='".esc_html( $dt_id )."'><td><a>".esc_html( $dt_duplicate['name'] )."</a></td><td>
-                            <a>" . esc_html( $dt_duplicate['count'] ). esc_html_e( 'duplicates' ) . "</a></td></tr>";
+                            <a>" . esc_html( $dt_duplicate['count'] ). /* esc_html_e( 'duplicates' ) . */ " duplicates</a></td></tr>";
```
corsacca: The line should be `<a>" . esc_html( $dt_duplicate['count'] ). esc_html__( 'duplicates', 'disciple_tools' ) . "</a></td>`

template-view-duplicates.php
corsacca: Ok. I don't disagree with you but please don't add TODOs for others to do in the code. 
We'll organize things to do in the issues.
So if you would please remove it here that would be great.

template-view-duplicates.php
mikeallbutt: Thanks - I spent way too long trying to try and resolve that.

template-view-duplicates.php
```diff
@@ -29,7 +29,7 @@
                         foreach ( $dt_duplicates as $dt_id => $dt_duplicate ) {
                             if ($dt_duplicate['count'] <= 0) { continue; }
                             echo "<form name='merge_".esc_html( $dt_id )."' method='POST' action='".esc_html( site_url() )."/contacts/".esc_html( $dt_id )."'><input type='hidden' name='dt_contact_nonce' value='".esc_attr( wp_create_nonce() )."'/></form><tr id='".esc_html( $dt_id )."'><td><a>".esc_html( $dt_duplicate['name'] )."</a></td><td>
-                            <a>" . esc_html( $dt_duplicate['count'] ). /* esc_html_e( 'duplicates' ) . */ " duplicates</a></td></tr>";
+                            <a>".esc_html( $dt_duplicate['count'] )." ".esc_html( 'duplicates' )."</a></td></tr>";
```
corsacca: esc_html__( 'duplicates', 'disciple_tools' ) ?

mikeallbutt: Yes - that's all that was needed - change the esc_html_e to esc_html because it was putting the word 'duplicates' before the for each. See any of the current live sites to confirm this issue.

corsacca: Your last commit is just escaping, not translating. Will you add that?

mikeallbutt: Sorry. Missed that. Fixed now. Changes have been committed.

template-view-duplicates.php
mikeallbutt: Todo comment was removed :)

Issue #815, opened 2019-10-29
Comments: Showing special characters i don&#039;t have &quot;42&quot;


corsacca:
Sometimes a comment is not display correctly.
Example: `My script: i don't have "42"`
shows:
`My script: i don&#039;t have &quot;42&quot;`

In this case the key word `script` is triggering an issue with the comment parser.

mikeallbutt:
If someone wanted to look at this issue, where would they start?

corsacca:
The comments.js

mikeallbutt:
> The comments.js

Thanks @corsacca. Maybe @blachawk can have a look at this?

blachawk:
Sure. @corsacca please feel free to assign this too me. I'll give it a shot.

blachawk:
Team, here is my theory.  This is intentional within the system.  I believe it is a security feature to prevent nasty JavaScript injections into the comments.  The JavaScript logic will look for certain patterns of strings to determine that.  Here is a note I found in comment.js. which controls the comment section:

```javascript
/* comment_content should be HTML. However, we want to make sure that
       * HTML like "<div>Hello" gets transformed to "<div>Hello</div>", that
       * is, that all tags are closed, so that the comment_content can be
       * included in HTML without any nasty surprises. This is one way to do
       * that. This is not sufficient for malicious input, but hopefully we
       * can trust the contents of the database to have been sanitized
       * thanks to wp_new_comment . 
*/
```

* If we choose to  un-escape these comment strings, it may set us up for attacks.
* According to the system, `script i don't have...` looks pretty close to `<script>console.log('inject something');...`

Let me know how you would like to me to move forward.

corsacca:
Unit tests:
Currently this works when displaying a comment:
- `[link name](http://dt.local/contacts/14493/)` makdown format for hyperlink
-  `@[bob](46)` this is a mention notification
- `http://dt.local/contacts/14493/` displays a link as clickable
- `Possible <strong>duplicates</strong> on Phone` shows up as: Possible **duplicates** on Phone. Note: if the users creates a comment with <strong>duplicates</strong> it will get escaped.

But this should not work:
- `<script>console.log('inject something');...`
- @todo look for a list of common xss attacks

@blachawk what special characters would be removed ?

blachawk:
lets start out with removing single or double quotation marks.

corsacca:
@blachawk Would this change this comment
[...]

PR #837, opened 2019-11-25
Bug/comments

blachawk:
This is relating to #815, and it may be a **very simple fix** for converting special characters (i.e. quotations) correctly.  We are keeping the entire Lodash process the same,  but simply tweaking the output process of jQuery to use text() instead of html().  We should be using text() over html() when dealing with inputs, wherever possible.

Please feel free to test this out and let me know.


corsacca:
@blachawk i like this approach
I get an error if the comment is `something <script>console.log("test")`
this seems to fix it:
`comment.comment_content = $("<div>").text(comment.comment_content).text()`

When editing a comment `don't` shows up as `don&#39;t` 
`.replace(/&#039;/g, "'");` needs to be `.replace(/&#39;/g, "'");` ?

blachawk:
@corsacca  let me know if these new adjustments make any difference on your end.

corsacca:
Using `comment.comment_content = $("<div>").text(comment.comment_content).text()` instead of 
`comment.comment_content = $("<div>").html(comment.comment_content).text()` at line 393 fixes the issue.
If it is sufficient i'd prefer that than removing all `>` and `<` signs

corsacca:
Unfortunately my suggestion made the initial issue a problem again.
`My script: i don't have "42"`
becomes: `My script: i don&#039;t have &quot;42&quot;` 
I had tested for other cases but forgot this one.

blachawk:
@corsacca in what case was the following causing an issue for you:

```js
comment.comment_content = $("<div>").html(comment.comment_content).text()
```

With this process enabled, and submitting `something <script>console.log("test")` into the comment field, I get the following result listed... `something &lt;script&gt;console.log(&quot;test&quot;)`

That should send a message to the viewer that we are not allowing custom scripts to be processed here.  If that is not what you want, let me know how the end result "should" look for this case.

Many thanks!

corsacca:
Sorry for the back and forth.
I'm not getting any issues any more with: 
`comment.comment_content = $("<div>").html(comment.comment_content).text()`
I must have had something else throwing that.
@blachawk thanks for finding a great solution!

Issue #732, opened 2019-07-01
Lists: Ability to see the last modified date for each record.


corsacca:

blachawk:
@corsacca, which wp-admin page are you referring to, in which you want a column for last modified date applied to?

corsacca:
It would be adding it to the contacts list. Probably next to the group column. It think the data is already available in the js.
![image](https://user-images.githubusercontent.com/24901539/69437123-f6edbe80-0d42-11ea-8830-31dfcbc80827.png)

blachawk:
if you think the data is already available from the database, and all it needs is to wired up via JavaScript, feel free to assign this to me and I'll give it a look locally and let you know.

blachawk:
requesting to close this conversation.  the branch is already merged.

PR #838, opened 2019-11-27
Enhancement/lists

blachawk:
This enhancement is a response to #732 
It will show the last modified date for contacts on the contact page.  The column is sortable.  And each row value will change with the given "contact details" are modified.


blachawk:
I want to cancel this request at the moment because it appears that **contact-enpoints.php**, is one file, which needs to be re-factored to include a last_modified value for "each" contact.  Without doing that, the last_modified value shows the same for every single contact.  I will need assistance in setting that up.

corsacca:
@blachawk the last_modified field is returned in the results:
image
And the list is built in the js. I don't think we need to refactor the php.

blachawk:
@corsacca I get the samee results but when you convert the value to php date() format, it's identical dates for every row.

corsacca:
try moment.unix( contact.last_modified ).format("YYYY-MM-DD")

blachawk:
Still stuck.  `contact.last_modified` appears unavailable in contacts-endpoint.php

blachawk:
Please test my efforts from pull request #842 Closing #838

PR #839, opened 2019-11-27
Giving each page a div.content class=...

mikeallbutt:
Setting up each page to have a class to use for scss defining.


single-contacts.php
```diff
@@ -211,7 +211,7 @@
 
 <!--    <div id="errors"></div>-->
 <!-- <div><a class="button small" id="backlink" href="../" style="margin:.9rem 0 0 1.3em"><?php esc_html_e( 'Back to Contacts List', 'disciple_tools' )?></a></div> -->
-    <div id="content">
+    <div id="content" class="single-contacts">
```
corsacca: This one and single-groups already have a class definition in the .scss files. Can you check that there are no conflict?

mikeallbutt: > This one and single-groups already have a class definition in the .scss files. Can you check that there are no conflict?

I had a look (using search in my app) but nothing was found using those terms, apart from the scss files. It would seem then that what is in the scss code is not being used.

corsacca: It does look like that scss code is not used.
To avoid complication down the road please remove the those 2 classes in the scss.

Issue #806, opened 2019-10-28
favicon design update


mikeallbutt:
The logo of DT has changed so the favicon should also.

Existing favicons
![image](https://user-images.githubusercontent.com/43966676/67710879-77104680-f986-11e9-8b0d-991d22b9be0d.png)

mikeallbutt:
Please assign this issue to me @mikeallbutt

mikeallbutt:
I found this online code that utilises gulp to make the favicons.
[https://realfavicongenerator.net/favicon/gulp](https://realfavicongenerator.net/favicon/gulp)

mikeallbutt:
@corsacca can you please have a look at this FAQ. [https://realfavicongenerator.net/faq#.Xd7WpJNKjwc](https://realfavicongenerator.net/faq#.Xd7WpJNKjwc)

I think we should follow the advice and move the favicons to root.

I would also like to remove the tags in the header.php file - doing so would make it easier to maintain too.

corsacca:
@mikeallbutt  why do we need a whole gulp process instead of just replacing the current icons with new ones?

mikeallbutt:
> @mikeallbutt why do we need a whole gulp process instead of just replacing the current icons with new ones?

We don't need all those images but the script makes them all - and it is a one off script that sped up the file creation process.

corsacca:
Ok, what are your steps forward?

mikeallbutt:
> Ok, what are your steps forward?
I will put the new files where you want them but my preference is in root.

I can also remove ones that don't get used, but if we leave them there it doesn't really affect anything.

> You asked about removing the favicon meta in the header. Is it not needed?

If you place the favicons in the root, the browsers will use the correct one. So defining in the meta tag in the header what favicon  to use is not needed.
That is explained well in this FAQ page
https://realfavicongenerator.net/faq#.Xd7WpJNKjwc

corsacca:
@mikeallbutt what do you mean by putting it the root folder? Is that the root folder of the D.T theme? Currently the favicons are in the dt-assets folder and your PR places theme in a different folder in dt-assets.

mikeallbutt:
> @mikeallbutt what do you mean by putting it the root folder? Is that the root folder of the D.T theme? Currently the favicons are in the dt-assets folder and your PR places theme in a different folder in dt-assets.

[...]

PR #841, opened 2019-11-27
favicons for issue #806

mikeallbutt:
Here are the new favicons using the new DT logo.


corsacca:
Thanks Mike!
image
It is hard to make out.
What would it look like with the green carret bigger? Also maybe without the blue background?
Like we have in Kingdom.Training:
image

This update would be at the same time as the logo update, which we don't want to do until we come out of beta.

corsacca:
There were more favicons previously than here, can you explain why they are not needed?

mikeallbutt:
> There were more favicons previously than here, can you explain why they are not needed?

As thouroughtly explained on this page https://realfavicongenerator.net/faq , all that is needed by the far majority of browsers etc. are what I have included. More icons and the header code could be added, but others have tested these browsers and found that they actually don't need all those other specific sizes.

package.json
```diff
@@ -25,6 +25,7 @@
     "gulp-newer": "^1.3.0",
     "gulp-plumber": "^1.2.1",
     "gulp-postcss": "^8.0.0",
+    "gulp-real-favicon": "^0.3.2",
```
corsacca: @mikeallbutt is this needed?

mikeallbutt: It can be used to generate the favicons so that one doesn't have to keep configuring all the settings, but instead use the now configure settings. The script is only run if desired (when a new favicon is needed).

Issue #749, opened 2019-08-20
Contacts: Ability to collapse/hide tiles on records.


corsacca:
This is already implemented for custom tiles created from the the wp-admin. It needs to be applied to the default tiles.
Also Save the closed/open preference for each user like the comment filter preferences.

Example:
Collapsed
![image](https://user-images.githubusercontent.com/24901539/63334787-06203300-c334-11e9-9c77-43c436143104.png)
Expanded
![image](https://user-images.githubusercontent.com/24901539/63334852-31a31d80-c334-11e9-9b7c-5edd9533ba16.png)

blachawk:
@corsacca is this requested suited for me to look into? or is this more PHP Class and Database related?

corsacca:
The html and css classes are already there on the custom tiles. They need to be copied from there to the default tiles.
Which tiles are open/close could be saved from js using a cookie. 
@blachawk I think you could give this one a go.

blachawk:
Ok feel free to assign to me and I'll look into it.

Thanks!

blachawk:
please feel free to close this conversation as the progress and communciation has advanced beyond this point.

corsacca:
Hey, we'll keep this open until the feature is complete and merged in.
I wasn't clear with the initial description, the intended tiles to be collapsible are the ones below the  details tile.
We also still need which tiles are closed/expanded to be a saved preference (cookie) so the layout is the same when opening a new contact/group.

PR #851, opened 2019-12-03
Collapsable tiles for issue #749

mikeallbutt:
All tiles have a chevron icon on the top right side which when clicked, opens or closes the tile.

### DIFF
--- a/dt-assets/js/details.js
+++ b/dt-assets/js/details.js
@@ -160,8 +160,8 @@ jQuery(document).ready(function($) {
   })
 
   // expand and collapse tiles
-  $(".section-header").on("click", function () {
-    $(this).parent().toggleClass("collapsed")
+  $(".section-header .section-chevron").on("click", function () {
+    $(this).closest('.bordered-box').toggleClass("collapsed")
     $('.grid').masonry('layout')
   })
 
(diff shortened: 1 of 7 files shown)


corsacca:
Hey, thanks for the initial work on this. Please limit the collapsible functionally to the dynamic tiles:
connections, progress and other. (similar for the group).
I'd like to have more discussion before hiding the contact details section (and the name).

mikeallbutt:
> Hey, thanks for the initial work on this. Please limit the collapsible functionally to the dynamic tiles:
> connections, progress and other. (similar for the group).
> I'd like to have more discussion before hiding the contact details section (and the name).

I have done as requested @corsacca but I think the details tile is very large, especially in mobile view) so having the option for a user to have it collapse is a good thing. If they don't want to collapse it then they won't. But if they wanted to and the option isn't there, there they are stuck.

corsacca:
@thanks @mikeallbutt 
We should come back to being able to collapse the details tile. But maybe with a UI upgrade for name to name not to be hidden.

corsacca:
Clicking the help icon still collapses/expands the tile.
There should be a way to exclude the help button from the collapsed js event.  @blachawk would you be up for joining @mikeallbutt in this one?

corsacca:
Implemented in: https://github.com/DiscipleTools/disciple-tools-theme/pull/877

PR #854, opened 2019-12-05
changed the Phone and Email Labels for internationalization

micahmills:
changed from esc_html to esc_html_e for the Email and Phone label

### DIFF
--- a/dt-assets/parts/contact-details.php
+++ b/dt-assets/parts/contact-details.php
@@ -2,11 +2,11 @@
 
 ( function () {
     $contact = Disciple_Tools_Contacts::get_contact( get_the_ID(), true, true );
-    $channel_list = Disciple_Tools_Contacts::get_channel_list();
+    $channel_list = Disciple_Tools_Contact_Post_Type::instance()->get_channels_list();
     $current_user = wp_get_current_user();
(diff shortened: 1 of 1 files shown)


corsacca:
thanks @micahmills !
`esc_html_e` can only be used with strings, so a variable won't work..
 `$channel_list['email']['label']` should already be translated, but like the definition `$channel_list` is being set before the user's locale is.
It looks like it might work by setting `$channel_list` with `Disciple_Tools_Contact_Post_Type::instance()->get_channels_list();` instead.
Can you give that a test?

micahmills:
I will take a look at the setting $channel_list differently but esc_html_e definitely works with a variable. I changed this in my local install and then checked it in several languages, including a new translation I am working on.

corsacca:
Looks good @micahmills 
You are right that the translation worked with the esc_html_e because the translation already exists in our .po files. If the variable was a new string ( and not translated elsewhere ) it would not be picked up when we build the translation files, so we try to avoid that use.

mikeallbutt:
> @mikeallbutt see the PR conversation #854 and the next commit

Thanks @corsacca , I read the conversation and saw the commit. I am now more knowledgeable. Gracias!

Issue #643, opened 2018-10-17
Ability to clear date pickers.


corsacca:

chancecox:
Which datepickers?  the baptized one worked fine for me. can you screenshot?

corsacca:
@chancecox once you add a baptism date, you can't then remove it. You can select the text, and deleted it but it does not trigger an update. 
This is the same for the 3 dates boxes on the group. And any custom date field added from the wp-admin.

mikeallbutt:
This issue has been around for a longtime. I wonder if @blachawk could help get it sorted?

corsacca:
Example: 
![image](https://user-images.githubusercontent.com/24901539/70112481-cb79a680-1655-11ea-97a9-225685773e77.png)

Date pickers that need to be clearable:

- [ ] Baptism Date on the contact record. Clicking clear should send an API call to delete the baptism date 
- [ ] In custom date fields added from the D.T settings. See render_field_for_display() in dt-contacts/contacts-template.php
- [ ] Start Date, Church Start Date and End Date in the groups edit modal

mikeallbutt:
Thanks @corsacca , so is there a place where the right type of code is already in place that can be used/copied to make this work?

corsacca:
No, that is why it is a feature to implement.

blachawk:
It appears that changing the date value works.  Removing it completely does not.  I feel confident I can solve this issue. If that sounds right, please feel free to assign this feature to me.

mikeallbutt:
What would be good and look nice would be to have the date display in a box like the what is in the who 'Baptised By' field, with a X to the right of the date, that when pressed, it deletes the date.

![Screenshot 2019-12-04 12 33 13](https://user-images.githubusercontent.com/43966676/70170574-8f186b80-1692-11ea-8d6a-d86cf11603b8.png)

Another option is to have the word 'Clear' but that doesn't looks as good IMO.

blachawk:
@mikeallbutt  what you are suggesting is right.  [But its not offered OOTB with ui-datepicker at the moment](https://github.com/OfficeDev/office-ui-fabric-react/issues/2725). 

We are going to have to a custom javascript hack for now. But the good news is that I found where the code needs to be modified in the javascript.  

Whats going on is that we are sending a `rest_api.update_post()` only when a date is selected. **We also** want to send a null value when the field is clear via keypress, [like this](https://stackoverflow.com/a/12649952/957186).

Working on it...

[...]

PR #858, opened 2019-12-07
Mix/datefields

blachawk:
This branch request is in response from conversations from issue #643 
Modifications here are bug and feature related.  

- Updated Contacts Detail page for setting and clearing baptism date
- Provided a button for members to click on, to clear the date
- Updated Groups Details area for date fields relating to start date, church start date and end date

Please test and let me know how it goes on your end!


corsacca:
Thanks @blachawk 
What is the issue with using `null` (or `undefined`) to clear the date in the API call?
It is now being set to "Mon Jan 01 1900" instead, which is would not be correct. From a stats and metrics perspective it will look like we had baptisms in 1900 which would be confusing.

Here is the current progress:
image
For visuals how about like this:
image
(taken from the edit modal)

blachawk:
Good question of concern. I did some research on this one...and here's my second thought.

While recently working on this issue, I discovered, that If we want to follow good practices with date/time operations in JavaScript, a non-null value should be implemented  when we are submitting and receiving date/time values.  This makes the process healthy for our code and the end-user gets what they want (a blank field).

The ui-datepicker naturally needs to set and get numerical date values in order to work effectively.   For blank entries from end-users, we can easily set the date value `1900 01-01`.  We know that it's an unreal situation for anyone to put in 1900 for a baptismal date.  With this condition, it makes it easy for us to keep the input field blank, until the end-user decides to put in the respected date.

That is my reasoning.  Let me know what you want to do.

corsacca:
@blachawk i think we'll use the null strategy of #864.
Thanks. I'll close this PR as it looks like everything is in the other branch as well.

Issue #783, opened 2019-10-08
Time Conversion to "how long ago"


ahillbilly:
As a user, I find it easier to read that an action took place 6 days ago or 6 months ago rather than seeing the date of the action and calculating it myself.

----- 

Normally The date/time is displayed with a standard timezone (show what one is used) for all guest users, and logged in users can usually change their timezone in their settings
At least, that’s how it traditionally works

-----
I just had a look at reddit to see how they do it.
They show how long its been, in the most appropriate single number digit, then hovering on that, shows the full time, with timezone.
so < 1 hour  = show minutes
< 1 day = show hours
< 1 month = show days
< 1 year = show months
1 year = show years
which I really like, it works well, means more at a glance, and hovering gives full detailed time

blachawk:
According to what I see in my Facebook timeline, Facebook resorts to a date, after 2 days. Within 2 days, they follow the methodology you are referring to @ahillbilly.  Do you see that as well?

PR #847, opened 2019-12-02
Features/contacts

blachawk:
This pull request connects to issue #749 
Attempting to apply collapse/hide functionality on contact default tiles. This is a test to confirm if the the `Progress` and `Other` tiles on the contact page are responding correctly. If so, then we can apply this logic to the rest of the sections on **single-contacts.php**.

Please let me know if the process looks good so far or not.  Thanks

PR #860, opened 2019-12-08
Enhance/timeconverson/howlongago

blachawk:
This pull request is in response to issue #783 
* Notifications should now be user-friendly with stating how long ago a comment was posted.


corsacca:
@blachawk It looks like the first if statement will match true for most dates:
`if ( $timestamp > $one_hour_ago ) {`
I'd expect the first entry to have the long ago format
image

corsacca:
Hey, you might still be working on this.
This branch shows
image
Master show:
image

blachawk:
@corsacca line #847 from the Travis CI log is throwing an error. Please advise on how I should correct this on my end.  Thanks!

corsacca:
@blachawk It works! :)
image

corsacca:
I like this improvement.
We need to make the strings translatable so it will work is more languages.

It would also be helpful to hover over the sentence to get the exact date.  I'm proposing sending both strings back, the formatted date and the how long ago string so that both can be used as needed.

Issue #710, opened 2019-05-24
Add tag functionality to group interface.


ahillbilly:
From user: 
We want to test tools and we want to track how many people/groups are trained with a tool to measure the tool’s effectiveness. Being able to add a “3/3” tag, or a “3 Circles Training” tag would be great. A while ago we were all trained on a new tool. We went out and tested it. Probably got 20+ attempts between our entire group but there is no way in Disciple.Tools to test how effective it was because there is no way to enter that in and track it across groups and individuals. Unless someone intentionally tallied up the results and is going to keep pushing it forward, the next person who comes along and wants to implement the same tool will have to start from square one. 20 attempts is not enough data points to prove effectiveness, but being able to track continued testing could provide better insights. 

This functionality is currently available for contacts.

mikeallbutt:
It would be good to use the same tags in groups that are available for individual contacts. One way I see this being used is that at a group, if they train 411, then all those who attended, would get a tag "411" to their contact tag system, as well as that group training/theme getting tagged.

mikeallbutt:
@corsacca how difficult would this be to implement? 
I suggest it uses the the same tags that are already available to contacts.
Is there an 'Other' tile for a group already for it to go in?

ahillbilly:
@mikeallbutt agreed on copying the functionality from contacts.

There is no tile for it, you will have to add/create it.

mikeallbutt:
I have begun a [branch](https://github.com/mikeallbutt/disciple-tools-theme/tree/group-tags) that I would like help with.

<img width="719" alt="Screenshot 2019-12-14 07 22 22" src="https://user-images.githubusercontent.com/43966676/70849348-bdd8d380-1e42-11ea-99f5-c916606ee550.png">

corsacca:
Closed by: https://github.com/DiscipleTools/disciple-tools-theme/pull/1020

PR #865, opened 2019-12-09
Add tag functionality to group interface. #710

mikeallbutt:


corsacca:
@mikeallbutt reference your branch from #710 and invite ppl to work on it from there.

PR #874, opened 2019-12-10
Translation branch renfinements

mikeallbutt:
While doing the Spanish translation, a few issues were noted and fixed that will make it easier in the future for the translators.


single-contacts.php
```diff
@@ -214,7 +214,7 @@
     <div id="content" class="single-contacts">
         <span id="contact-id" style="display: none"><?php echo get_the_ID()?></span>
         <span id="post-id" style="display: none"><?php echo get_the_ID()?></span>
-        <span id="post-type" style="display: none">contact</span>
+        <span id="post-type" style="display: none"><?php esc_html_e( 'contact', 'disciple_tools' )?></span>
```
corsacca: This needs to not be translated.

single-contacts.php
```diff
@@ -668,14 +669,14 @@ class="dt_multi_select <?php echo esc_html( $class ) ?> select-button button ">
...
-        <p class="lead"><?php esc_html_e( 'Create User', 'disciple_tools' )?></p>
+        <h1><?php echo esc_html_x( 'Make User From Contact', 'Make user modal', 'disciple_tools' )?></h1>
 
         <?php if ( isset( $contact['corresponds_to_user'] ) ) : ?>
-            <p><?php esc_html_e( "This contact already represents a user", 'disciple_tools' ) ?></p>
+            <p><strong><?php echo esc_html_x( "This contact is already connected to a user.", 'Make user modal', 'disciple_tools' ) ?></strong></p>
         <?php else : ?>
 
-        <p><?php esc_html_e( "This will invite this contact to Disciple.Tools as a multiplier", 'disciple_tools' ) ?></p>
+        <p><?php echo esc_html_x( "This will invite this contact to become as a of this system. By default, they will be given the role of a 'multiplier'.", 'Make user modal', 'disciple_tools' ) ?></p>
```
corsacca: This will invite this contact to become as a of this system.
Missing a word?

single-contacts.php
```diff
@@ -668,14 +669,14 @@ class="dt_multi_select <?php echo esc_html( $class ) ?> select-button button ">
...
+        <p><?php echo esc_html_x( "In the fields below, enter their email address and a 'Display Name' (which be changed by them later).", 'Make user modal', 'disciple_tools' ) ?></p>
```
corsacca: Better wording needed

single-contacts.php
mikeallbutt: This will invite this contact to become a user of this system

single-contacts.php
mikeallbutt: This is the phrase complete: (What do you suggest re-wording?)

Make User From Contact

Invite this contact to become a user of this system. By default, they will be given the role of a 'multiplier'.

In the fields below, enter their email address and a 'Display Name' (which be changed by them later).

corsacca: which be changed by them later

single-contacts.php
mikeallbutt: Ok. I have reverted it back.

single-contacts.php
mikeallbutt: Ok done.

corsacca: There is still something missing.
Which **can** be changed by them later?
Or maybe better: which they can change later.

single-contacts.php
corsacca: @mikeallbutt did you change this?

corsacca: yes you did.

dt-assets/scss/_main.scss
```diff
@@ -899,3 +899,14 @@ label.dt-switch:active:after {
     list-style-type: none;
     margin: 0;
 }
+
+
+/**
+ * Reveal
+ */
+ div.reveal {
```
corsacca: This means that any modal can't use the h1, h2 and h3 class differently anywhere on D.T
It would be better to create a class modal-header (like `.section-header`) and apply it where needed.
```
div.reveal {
  .modal-header { ... }
}
```

template-contacts-new.php
```diff
@@ -24,25 +24,25 @@
                     <button class="help-button" data-section="contact-name-help-text">
                         <img class="help-icon" src="<?php echo esc_html( get_template_directory_uri() . '/dt-assets/images/help.svg' ) ?>"/>
                     </button>
-                    <input name="title" type="text" placeholder="<?php esc_html_e( "Name", "disciple_tools" ); ?>" required dir="auto" aria-describedby="name-help-text">
+                    <input name="title" type="text" placeholder="<?php echo esc_html_x( "Name", 'input field placeholder', 'disciple_tools' ); ?>" required dir="auto" aria-describedby="name-help-text">
                 </label>
-                <p class="help-text" id="name-help-text"><?php esc_html_e( "This is required", "disciple_tools" ); ?></p>
+                <p class="help-text" id="name-help-text"><?php echo esc_html_x( "This is required", "disciple_tools" ); ?></p>
```
corsacca: missing context

dt-assets/scss/_main.scss
mikeallbutt: My way means that all of those H1, H2, H3 aree styled the same so that when coding you don't have to add classes, but if you want the headings to be different (ie. not use the styling) then add an over-ride styling.

template-contacts-new.php
mikeallbutt: `<?php esc_html_e( "This is required", "disciple_tools" ); ?>`
Interesting. My code has this and not what you quoted. Not sure why. Anyway, I think without context is better as it isn't too hard to translate.

PR #877, opened 2019-12-12
Assist/collapsabletiles

blachawk:
This PR is for assisting @mikeallbutt, at #851, allowing users to expand and collapse tiles, only when a section chevron is clicked for that given tile.


corsacca:
Perfect @blachawk.

corsacca:
@blachawk @mikeallbutt thank you both for your efforts on this feature.

Issue #856, opened 2019-12-06
People Group/Location/Source Name Translations


micahmills:
So the internationalization works really well and most of the data seems to be translatable. I have noticed that when creating or updating a contact's details the Location, Source, and People Group dropdown are populated with English data. I am sure that this is because the data sources for these fields is in English, and doesn't provided the localized names.

For my region and users I need particularly to have the people groups and city names written in the local script and language, or else they will get ignored and we will miss key data. 

I am happy to work on this, but since I am new to this project I wanted to get some input on the best path forward. My initial thought is to add custom fields to the Post Type, but then we would have to sort out getting those to populate in the dropdown based on the users locale. I could also add them to the translation files, but there might be some downsides there I haven't seen yet.

Any insight or ideas on this?

corsacca:
Hey, 
Great questions.

You can modify and add custom entries to fields like **sources** in the wp-admin under Settings (DT) -> Custom Fields.
The sources field was treated differently when we started (and still is in some ways) and it looks like we didn't come back and make those default options translatable. Issue created: https://github.com/DiscipleTools/disciple-tools-theme/issues/857
You can rename or delete and add new ones as you wish. The only downside with custom values is that all users will see the same translation for that option (see https://github.com/DiscipleTools/disciple-tools-theme/issues/809).

**People Groups** you can easily rename from the People Groups tab in the wp-admin.

**Locations** you can as well from wp-admin > Mapping > Locations List.
Chose a location in "Select a location to Edit" and update the name.
If in your context you have a considerable number of locations this could be quite a chore. We'd love help in automating this process.

mikeallbutt:
In my context, I just need 2 languages - Spanish and English - so getting this issue fixed will be one more step closer to having users use DT.

micahmills:
I think that @mikeallbutt and I are having the same issue and it is similar to #809. 

For instance, I need to have the list of People Groups and cities  to be in Latin script when the locale is English, and Persian/Arabic script when the local is Persian. If I change the name of the people group in the People Groups Tab, or the Locations List and make them all in Persian/Arabic Script then someone viewing it in English will only see the Persian script names.

I could create a second list of People Groups and Locations, one in English and one in Persian, but that will make reporting more difficult. 

The ideal situation is something similar to what happens when we select a gender for a contact. What shows up on the front end is pulled from the .po file based on their locale, and what gets saved to the DB from the input value can be in English.

mikeallbutt:
[...]

PR #881, opened 2019-12-17
Feature people groups translations

micahmills:
Set up Custom Meta for the People Group post type so there is now a field to enter a localized name for each supported language.

If there is a translation for the users locale we use that in the Typeahead, if not we use the standard post_title.

partially solves #856.


dt-assets/js/shared-functions.js
```diff
@@ -129,8 +129,7 @@ window.TYPEAHEADS = {
     typeaheadSource : function (field, url) {
         return {
             contacts: {
-                display: [ "name", "ID"],
-                template: "<span>{{name}}</span>",
+                template: "<span>{{label}}</span>",
```
corsacca: I think this is the change that leads to this for the groups typeahead on the contacts:
image

corsacca:
Nice work @micahmills 
Tested it out and seemed to work with added a french translation for a ppl group
I'm seeing this error on the people groups record edit in the wp-admin:
Disciple_Tools_Core_Endpoints::get_settings() should not be called statically in disciple-tools-theme/dt-people-groups/people-groups-post-type.php on line 370
I think i want to create a dt_get_available_languages (maybe in global-functions.php) instead of in the endpoint function.

dt-assets/js/shared-functions.js
micahmills: I can't seem to find a way to do conditional formatting for the template in typeahead. what I want to do is something like `if(label) { "<span>{{label}}</span>" } else { "<span>{{name}}</span>"` but that doesn't seem to work. 

The other option is to just add a label key to the groups endpoint like I did for the people groups. The conditional would live there then and if there is no translation label just use the post_title.

corsacca: We can create a "peopleGroupsTypeaheadSource" and use it for all the people group typeaheads.

corsacca: alternatively this might work:
```
`<span>{{ ${ label ? 'label' : 'name' } }}</span>`
```
But i prefer separating out the functionally since it is different enough

micahmills: should we make the TypeaheadSource more generic since we will probably use it for things like Locations, and Sources in the future? maybe something like "translatableTypeaheadSource"?

corsacca:
copied code to group details.

PR #885, opened 2019-12-22
change CRLF to LF endings for gulp sourcemaps using gulp-frep

wzcblair:
On Windows, the CSS and JS files that gulp generates has `\r\n` at line endings instead of `\n`. This change replaces all `\r\n` instances with `\n` to facilitate development on Windows.


corsacca:
@wzcblair thanks for helping us extend to windows dev too!

Issue #752, opened 2019-08-23
Round modal boarder corners


ahillbilly:
Round all modal boarder corners to match record detail tiles rounded corners.  This is a simple UI upgrade.
![Screen_Shot_2019-08-23_at_3_23_21_PM](https://user-images.githubusercontent.com/25323324/63599784-311bb880-c5ba-11e9-8821-f44f61753515.jpg)

mikeallbutt:
That might be something I can fix.

groovyjosh:
What kind of browser support are we targeting? That could affect the CSS rules we use.

corsacca:
Hi @groovyjosh!
Desktop and mobile. Some users have cheap older phones, so wider support is better.

corsacca:
Close by PR #886

PR #886, opened 2019-12-23
add border radius to help modals

wzcblair:
Resolves issue #752


corsacca:
@wzcblair thanks! This looks good for the help modal. We'd like to see the same for the other modals too.

wzcblair:
Those are all the modals that I can find, please let me know if I missed anything.

corsacca:
Thanks @wzcblair!
Would there be any down side to using the `.reveal` class instead?
The reveal class is a requirement of the foundations modal we are using (called reveal): https://get.foundation/sites/docs/reveal.html and then the change would also apply to the modals existing plugins are adding (demo content plugin for example).
Thanks for working on Disciple.Tools with us :)

wzcblair:
Ah, I wasn't sure whether `.reveal` has more general purposes. I've kept some of the other naming changes not related to this issue -- let me know if you'd like them the way they used to be.

corsacca:
Looks great.
The `-modal` adds clarity too. 
Thanks @wzcblair

PR #904, opened 2020-02-19
Made changes to dt_custom_dir_att() to accommodate multilingual DT us…

micahmills:
…e cases where the users locale might be different from the Default site's locale. Also added a translators note for the 'ltr' string for clarification.


micahmills:
I am moving this PR to the master branch instead of the strings branch since it is not specifically related to the strings in DT.

PR #905, opened 2020-02-19
Format numbers and dates

micahmills:
fixed the language attribute in Multilingual situations.

Fixes the time stamps and date strings to use localized string, but still uses Gregorian calendar.

Fixes the number formatting in the comments tab.

### DIFF
--- a/dt-assets/js/comments.js
+++ b/dt-assets/js/comments.js
@@ -10,6 +10,8 @@ jQuery(document).ready(function($) {
 
   let comments = []
   let activity = [] // not guaranteed to be in any particular order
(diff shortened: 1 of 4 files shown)


corsacca:
Should get_user_locale() be used all throughout the theme instead of $user->locale?

micahmills:
likely yes. In my use case $user->locale was empty and that was leaving the lang attribute empty on the HTML tag. get_user_local() has a check to fallback to get_locale() if $user->locale is empty.

micahmills:
So I had a chance to look at this. $user->locale will be empty if the user has their locale set to "site default." 

I will see if I can track down all the $user->locale instances and replace them.

corsacca:
I've created https://github.com/DiscipleTools/disciple-tools-theme/issues/906 to keep track of looking for the other place in the code that could use an update.
Thanks for the commits!

PR #907, opened 2020-02-20
Get user locale

micahmills:
I think I have found all the instances of get_user_locale that needed to be changed.


corsacca:
Thanks @micahmills!

Issue #751, opened 2019-08-20
Navigation menu hiding tool bar


corsacca:
When the screen is in between mobile and desktop size the to bar uses 2 rows and covers the toolbar below it.
![image](https://user-images.githubusercontent.com/24901539/63339538-053fcf00-c33d-11e9-9573-3eaaffb357e1.png)

corsacca:
This issue was fixed.

PR #909, opened 2020-02-21
changed the medium breakpoint to fix issue #751

micahmills:


corsacca:
Some of our extensions add more tabs which makes them overflow.
image
We might need a more responsive detection or a combination of tabs and the hamburger menu.

corsacca:
Thanks @micahmills!

Issue #897, opened 2020-01-20
Auto focus first input when opening a modal.


corsacca:

corsacca:
Here are the places i would expect the first input to be focused
- [ ] Create new contact page
- [ ] Create new group page
- [ ] Create new custom post type page ( template-new-post.php )
-- in modals:
- [ ] Create contact modal from the contact details page. The modal opens from the typeahead button
- [ ] Create group modal from the contact details page. 
- [ ] Create contact modal from the group details page. 
- [ ] Create group modal from the group details page. 
- [ ] Create tag modal from the contact details page.

PR #911, opened 2020-02-29
Add focus on first modal input

aclowwwn:
refs #897


corsacca:
This is great. Thanks @aclowwwn!
I've added to the issue other places we would like to see the same functionality.

aclowwwn:
Cool, thank you @corsacca . On it 😬

PR #914, opened 2020-03-09
fixes #879

### LINKED ISSUE #879: Language change doesn't reflect changes
My instance of DT uses `language1` as the default language but in the `my.discipletools.in/settings/` when I go to edit the language, and set it to `language2` for a user, the change isn't recognised if on the user profile page in /wp-admin/profile.php is set to use `site default`.

micahmills:


corsacca:
@micahmills this works great!

PR #916, opened 2020-03-10
Fixes a bug where the Notification badge is not properly aligned in R…

micahmills:
The badge icon on the notifications menu was not properly aligned if you have a RTL orientation.


corsacca:
Yeah. I'd noticed that too. Thanks for for fixing it!

Issue #880, opened 2019-12-14
Group details tile: Date fields using wrong language


mikeallbutt:
My site uses lanuage1/spanish as the primary language but I choose to override this and use language2/english. When I view the interface it correctly shows all the labels in language2/english - except fields where dates are displayed in the group details tile in lanuage1/spanish (the site default language).

![Screenshot 2019-12-14 07 06 57](https://user-images.githubusercontent.com/43966676/70849160-57eb4c80-1e40-11ea-8b5a-bb618eaf1dc9.png)

micahmills:
So I was seeing the same thing and I dug into this a bit. The group dates are set in line 131 of dt-core/global-functions.php. It is getting that data format from the General Settings of Wordpress. We can consider doing changing the way that function works, but the easy solution is to change the  Date Format in the WP admin Settings->General. I chose d/m/Y and that solved my problem for me.

corsacca:
Todo: get formatted dates in user's local

PR #925, opened 2020-04-01
Group date formatting

micahmills:
Fixes Issue #880

### DIFF
--- a/dt-assets/js/comments.js
+++ b/dt-assets/js/comments.js
@@ -51,7 +51,7 @@ jQuery(document).ready(function($) {
     let createdDate = moment.utc(currentContact.post_date_gmt, "YYYY-MM-DD HH:mm:ss", true)
     const createdContactActivityItem = {
       hist_time: createdDate.unix(),
-      object_note: settings.txt_created.replace("{}", formatDate(createdDate.local(), langcode)),
+      object_note: settings.txt_created.replace("{}", window.SHAREDFUNCTIONS.formatDate(createdDate.unix())),
       name: settings.contact_author_name,
       user_id: currentContact.post_author,
     }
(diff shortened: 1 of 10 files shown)


corsacca:
Can we still use the month name? I switch between English and French a lot so i can't tell if this is Jan or Feb.
image

corsacca:
It looks like the contact baptism date also has the some problem with using the site default language. I can open a new issue if you want to keep it separate.

micahmills:
I noticed this yesterday as well. I will take a look at it today and tomorrow.

micahmills:
A new issue isn’t necessary.

micahmills:
The TLDR version of this PR is that PHP/Wordpress doesn't have a easy way to format date strings based on the users locale instead of the servers locale. So I have switched most things to use the timestamp that we were already saving, and then converting that timestamp to a date string in JS, which has much nicer methods for handling these things. Now we don't have to worry about i18n date strings being saved into the DB.

corsacca:
Thanks @micahmills!
If we follow the {timestamp} route for the activity we'll need to get it working in the mobile app as well:
image
We can go that route. An alternative might be to add a meta_date timestamp field to the activity (dt-posts.php ~line 760) which can be used to display a better date.

micahmills:
I am fine with having a Time stamp from the metadata, but if we keep the string with the date in the mobile app or web, it will always appear in the wrong locale if the users locale is different from the DT instance. 

My opinion is we should always store dates as a unix time stamp and then calculate the date in the appropriate locale from the time stamp. I find the tools to calculate that more robust in JS than in PHP.

micahmills:
I created a PR for the mobile app to implement the activity dates as I have done here.

Issue #768, opened 2019-09-16
Comments: show comments and their translation using google translate or other service


corsacca:
Copy Facebook's comment/post auto translation where the comment is in one language, and shows the translation under it in the user's language. 
This will help speed up communication for multi lingual teams.
![image](https://user-images.githubusercontent.com/24901539/64955492-45349c00-d880-11e9-9da1-74b6b70b5286.png)

This will also help users understand the conversations happening on social media (ie Facebook messages) to determine the best next step.

mikeallbutt:
This would be a useful feature for the prayer team who might not be able to read the comments written by a multiplier of a different language.

micahmills:
I was looking into this today. Microsoft and Google offer Text Translation APIs but they require API keys and the free tiers stop at about 2 million characters of translation a month. 

One option would be to add a settings field in the admin with instructions on how to set up a Microsoft of Google account with and then enter the API key. Then if that API key exists we could show a button to translate the text similar to Facebook or Twitter. This would put the responsibility of setting up payment and managing the billing on the user, and it would leave those who don't need the feature unchanged.

Is this something that seems workable?

corsacca:
That is a good idea.
We use that strategy for some of the mapping features. In the mapping settings you can enter your mapbox key and that will enable extra features related to mapping and geocoding.

corsacca:
Thank you @micahmills! :)

PR #926, opened 2020-04-02
Comment auto translate

micahmills:
Adds the Auto translation of comments feature mentioned in Issue #768.

This uses Google Translates API, and requires you to sign up for and set up billing for an API key through he Goolge Cloud Console. I have included instructions for that in the Admin (wp-admin/admin.php?page=dt_options&tab=translation). Once a API key is added a Translation button will appear.


corsacca:
This is so beautiful!
You even detect the locale of the user and it translates into that language.

My only request is to change "See translation" into "Translate with Google" or something similar to add transparency for the users.

micahmills:
Yeah that makes sense to me and is easy to change.

corsacca:
Translate with Google and Hide translation will need be be translated too :)

PR #928, opened 2020-04-09
Fix "Waiting on Accept" Metric

cairocoder01:
On the Project Overview metrics dashboard, the Waiting on Accept is not showing the same number as when you filter the contacts list by the same status. It seems to be because the value in the database could be either blank or "no" and the metrics query was only checking for blank.

### DIFF
--- a/dt-metrics/metrics.php
+++ b/dt-metrics/metrics.php
@@ -859,7 +859,7 @@ abstract class Disciple_Tools_Metrics_Hooks_Base
                     JOIN $wpdb->postmeta as b
                     ON a.ID=b.post_id
                        AND b.meta_key = 'accepted'
-                       AND b.meta_value = ''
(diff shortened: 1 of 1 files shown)


corsacca:
@cairocoder01 it might be good to remove the meta_key = 'accepted' part of that query since the contact list page filter displays the number of those that have the 'Waiting to be accepted' status.

cairocoder01:
@corsacca That changes the logic of this existing metrics query. I didn't want to make a major change but wanted to fix the logic that was already in place. The current logic is those contacts that have been assigned but do not have a status of accepted.
I would agree that ideally the contacts list and metrics would query in the same way. If you want me to look at making that more significant change, let me know.

corsacca:
Thanks @cairocoder01

PR #935, opened 2020-04-14
Add `post_date` to `search_viewable_post` results

cairocoder01:
A contact's creation date is not currently available when using DT_Posts::list_posts, so this adds it in.


corsacca:
This looks good. I'm wondering if we change it to `created_date` to stay consistent with the get_post function. Tho i prefer the `post_date`. I'll get back to you.

corsacca:
Thanks @cairocoder01

PR #938, opened 2020-04-16
Custom comments and activities

micahmills:
Creates custom comment types for each of the items in the Contact Communication Channels.
Adds a comment type selector to the create and update comments.
Adds "comment_type" to the update_comment API endpoint.


micahmills:
The remaining question here is one of UI, What do you think of the UI of having a "Type of Comment" Select element below the Comment box?

micahmills:
Making a note here so I don't forget, when this branch gets merged we need to add 

- comment_type (string) optional. The comment type. Default: 'comment'

to the [Update A Comment](https://github.com/DiscipleTools/disciple-tools-theme/wiki/Post-Comments#update-a-comment) section of the Wiki

corsacca:
This looks cool! :)
For some reason The indent of the 2nd+ comments isn't aligned any more.
image
I'd remove "activity" from the comment type list. Activity messages are generated by the server.

corsacca:
Actually the alignment issue isn't related to this issue. So i created https://github.com/DiscipleTools/disciple-tools-theme/issues/942

micahmills:
I saw your new issue. Is the alignment issue related to something in this PR or was it there before?

corsacca:
It is not related to this PR. I haven't investigated but it looks like it might be something from the translation PR since it breaks alignment after "translate with google".

micahmills:
I may know what is causing it.

micahmills:
I’ll fix it in a separate PR

dt-posts/dt-posts.php
```diff
@@ -676,7 +676,7 @@ public static function add_post_comment( string $post_type, int $post_id, string
         return $created_comment_id;
     }
 
-    public static function update_post_comment( int $comment_id, string $comment_content, bool $check_permissions = true ){
+    public static function update_post_comment( int $comment_id, string $comment_content, string $comment_type = "comment", bool $check_permissions = true ){
```
corsacca, line 679: @micahmills this breaks anywhere that is using this function with the $check_permissions param.

dt-assets/js/comments.js
```diff
@@ -207,6 +209,8 @@ jQuery(document).ready(function($) {
 
   $(document).on("click", ".open-edit-comment", function () {
     let id = $(this).data("id")
+    let comment_type = $(this).data("type");
+    console.log(comment_type);
```
corsacca, line 213: Maybe remove the log.

corsacca:
We'll need to also remove the Activity option  from the edit modal.

dt-posts/dt-posts.php
micahmills: do you have a suggestion for how to fix this?

corsacca: In php add the args at the end of the list, so it keeps the same order for all functions.
here i'd suggest we future proof it and format it like this:
```
    public static function update_post_comment( int $comment_id, string $comment_content, bool $check_permissions = true, $args = [] ){
        $comment = get_comment( $comment_id );
        if ( $check_permissions && ( ( isset( $comment->user_id ) && $comment->user_id != get_current_user_id() ) || !self::can_update( get_post_type( $comment->comment_post_ID ), $comment->comment_post_ID ?? 0 ) ) ) {
            return new WP_Error( __FUNCTION__, "You don't have permission to edit this comment", [ 'status' => 403 ] );
        }
        if ( !$comment ){
            return new WP_Error( __FUNCTION__, "No comment found with id: " . $comment_id, [ 'status' => 403 ] );
        }
        $comment = [
            "comment_content" => $comment_content,
            "comment_ID" => $comment_id,
        ];
        if ( isset( $args["comment_type"] ) ) {
            $comment["comment_type"] = $args["comment_type"];
        }
        return wp_update_comment( $comment );
    }
```

micahmills: Yeah that makes sense, here I thought it was something difficult and was thinking too hard about it.

corsacca:
@micahmills please test :)

corsacca:
Thank you @micahmills !

Issue #931, opened 2020-04-13
If you create a a contact and delete the contacts name you cannot then open the user


micahmills:
When you create a contact if forces you to enter a name, but if you then open that new contact and edit the information you can clear out the contacts name. It will then show up in the list, but the link to open the contact is not there because there is not name to put the link on.

corsacca:
I think this is an issue on the list page and the notification page. 
It would be great to make the cell (or the whole row?) clickable instead of just the name.
Plus we might want to protect from the user erasing the name in the UI and in the create/update code

PR #940, opened 2020-04-16
Adds CSS to make the Name area clickable when thename has been deleted

micahmills:
This makes the area clickable even if there isn't a name. It solves the immediate problem of not being able to access a contacts record if the name has been deleted. fixing #931.

We should probably consider some sort of form validation that keeps someone from deleting a name int he first place.


corsacca:
NIce!

Issue #770, opened 2019-09-17
Comment: Paragraph line separations are deleted when submitting a comment


corsacca:
If i write a comment like this with a line in between 2 paragraphs:
![image](https://user-images.githubusercontent.com/24901539/65036103-973df600-d942-11e9-950d-08318ce47a6f.png)
The comment in the comments section does not show the extra line:
![image](https://user-images.githubusercontent.com/24901539/65036162-b6d51e80-d942-11e9-9fcb-fae58e0a279f.png)

corsacca:
Fixed by @micahmills!

PR #941, opened 2020-04-16
fixes issue #770

micahmills:


corsacca:
Brilliant. Merging. How does adding `min-height: 1em;` make all the problems go away?

micahmills:
We are replacing each line break with a div which meant when there was an empty line there was an empty div. because it was empty it had no height. By giving it a min height it forces a lines space between. 

It probably isn’t the most semantic solution but it works.

PR #948, opened 2020-04-19
Farsi translation update

micahmills:
The Persian (Farsi) translation is now 100%


corsacca:
@micahmills have you also updated poeditor?

micahmills:
Yes. Poeditor is updated

PR #968, opened 2020-04-28
Farsi translation update

micahmills:
changed word for people group.


micahmills:
It looks like Travis ran twice for some reason here. It passed the first time and has just been stuck for 2 days. All that changed in this PO were the Farsi .po and .mo files so there should be no problems.

corsacca:
@micahmills the simplest for us would be for the changes to be uploaded directly to poeditor.com
With each release we go in and take all the changes from there and update the theme.

micahmills:
I have added these changes in poeditor.com as well.

PR #984, opened 2020-05-06
combines the Translate with Google for all comments

micahmills:
moved the translation-bubble below all grouped comments to make it more legible


corsacca:
The translate option does not seem to appear if the section starts with an activity but has comments farther down.

image

micahmills:
I believe I have fixed this so the Translate buttons appear if there is a mixture of activity and comments.

corsacca:
Thanks @micahmills !

PR #993, opened 2020-06-11
Access by source for Digital Responder And Partner Role.

corsacca:
Added the partner Role.
Added ability to configure access for the Digital Responder and the Partner roles to have access to all the contacts of a specific sources.


corsacca:
@cairocoder01 Would you be up for giving this a test?

corsacca:
Thanks!

PR #998, opened 2020-07-02
Comment Bug Fix

micahmills:
removes escaping from the Javascript so that the PHP gets the unescaped HTML and then applies wp_kses_post() to it to strip out malicious attributes and tags.


corsacca:
Pasting this as a comment is still an issue:
`&lt;img src=. onerror=alert('imagealert2');&gt;`

micahmills:
It looks like this is fixed. But it might be worth having some automated tests to make sure we don't break this in the future. We could use a list like https://portswigger.net/web-security/cross-site-scripting/cheat-sheet as a list of XSS attempts to check.

PR #1000, opened 2020-07-03
Fixes #999

### LINKED ISSUE #999: Special Characters on Contact Lists not displaying properly
It looks like the contact list page does not handle HTML special chars properly. It is showing them in encoded format. This shows up when looking at the Contact List with contacts "Waiting to be accepted" and the language set to French.

image

micahmills:
updated the pot and all po files that had the apostrophe ' (&#39;) with the apostrophe ’ (&#8217;). I recompiled the .mo file for french but not the other translations.

Fixes #999.


corsacca:
@micahmills the .pot is generated from the source code. I'd expect we'd need to change the single quotes into apostrophes in the source strings instead, so that the .pot is not overwritten next time.

micahmills:
@corsacca yeah I was a little lazy and did a find and replace on all the apostrophes and replaced them, that was a bit to blunt of a hammer for that nail. I have reversed all the changes and only changed the msgstr lines that have an apostrophe. 

French is a particularly heavy user of the apostrophe compared to other languages.

corsacca:
Fixed the issue by not double escaping the values to be displayed: https://github.com/DiscipleTools/disciple-tools-theme/commit/8a05863b28bc3426940bf56d897db244efa87fe2

PR #1001, opened 2020-07-08
User Visibility by Role Preferences

micahmills:
Adds a Preference pane in Settings (DT) -> General Settings. This lists all the DT Roles and allows you to check the box if you want to allow that role to have access to the DT users on the front end. This will allow things like Assign, Sub-assign, and @ mentions to list all DT Users.

### DIFF
--- a/dt-users/users.php
+++ b/dt-users/users.php
@@ -66,7 +66,7 @@ class Disciple_Tools_Users
         $user_id = get_current_user_id();
         $users = [];
         $update_needed = [];
-        if ( !user_can( get_current_user_id(), 'view_any_contacts' ) ){
+        if ( !user_can( get_current_user_id(), 'view_any_contacts' ) && !user_can( get_current_user_id(), 'list_users' ) ){
             // users that are shared posts that are shared with me
             $users_ids = $wpdb->get_results( $wpdb->prepare("
                 SELECT user_id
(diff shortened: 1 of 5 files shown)


micahmills:
I am not 100% happy with the UI for the preference pane. It might need to be a little more clear, but this is functional.

corsacca:
Users were showing up in the groups typeahead on the contact record.

I'm not comfortable with multipliers with the `list_users` permission having access to the `/user-management/users/` page.

micahmills:
I didn't realize that they had that access to the /user-management/users I think that would probably be solved by adding a custom capability instead of the list_users.

dt-users/users.php
```diff
@@ -66,7 +66,7 @@ public static function get_assignable_users_compact( string $search_string = nul
         $user_id = get_current_user_id();
         $users = [];
         $update_needed = [];
-        if ( !user_can( get_current_user_id(), 'view_any_contacts' ) ){
+        if ( !user_can( get_current_user_id(), 'view_any_contacts' ) && !user_can( get_current_user_id(), 'list_users' ) ){
```
corsacca, line 69: dt_list_users

corsacca:
For clarity: can we `disable` the rows that have `view_any_contacts` permission. I'm not sure we wan't to disable "list users" for those roles, and it does not look like unchecking them removes the capacity to list users..
image

micahmills:
Yes that makes sense

micahmills:
Having the Admin and other roles with view_any_contacts capabilities listed but disabled felt awkward so I just removed them from the list. Now the list only have User roles that don't inherently have the ability to view all contacts.

corsacca:
Thanks @micahmills !

PR #1021, opened 2020-07-15
Perform site link check from backend

cairocoder01:
This will more closely simulate how contact transfer works so that those using IP whitelist can get properly setup


corsacca:
This looks great! 

For this scenario with the remove not having ssl set up.
image
I suggest ~line 1239: `$https_failed = strpos( $error_message, 'SSL' ) > -1 || strpos( $error_message, 'HTTPS' ) > -1 || strpos( $error_message, 'certificate verification failed' ) > -1;`

cairocoder01:
@corsacca Of course. Added that additional check

PR #1028, opened 2020-07-27
Add button to fetch server IP address while creating/editing site link

cairocoder01:
**Use case:**
Our hosting implements IP whitelisting at the firewall to prevent unwanted traffic. In order to authorize site link traffic via the API, we need to know the server IP address that requests will be coming from. This adds a button to make a request to ifconfig.co/ip to get the server's IP address since it cannot be consistently retrieved from PHP server variables because of varying hosting methods (i.e. Apache vs. nginx, basic single hosted site vs. load balancing, etc.)


corsacca:
Seems useful and could be used in the "Approved IP Address" field in the section below as well.

PR #1029, opened 2020-07-27
Add filter for assignable users

cairocoder01:
Allows the insertion of custom tagging options in comments by plugins


corsacca:
This will be useful!

PR #1031, opened 2020-07-30
Escapeing bug

### LINKED ISSUE #929: &amp; in title of contact or group
If the symbol of "&" is used in the name of a contact or group, it displays as "&amp;" in many places like members list. eg. Terry (&amp; Sandy) Kernel 

For a group the same issue happens eg. Surf &amp; Skate group

micahmills:
fixes #929.

The problem is occurring when someone with a role that does not have `unfiltered_html` capabilities (everything except Administrator) created a contact or group with an & in the title. This almost fixes the issue.

Right now this works for everyone, but if an Administrator (not a dt_admin) or Super_admin in Multisite creates a contact or group with an XSS attack it doesn't get filtered because they have unfiltered_html capabilities and Wordpress doesn't fun it through kses functions. This is Wordpress's default settings and any user with Admin or Super_admin role can put malicious XSS code into a post, but honestly someone with those roles could do this through any number of means with out needing a XSS vulnerability.

### DIFF
--- a/dt-assets/js/contact-details.js
+++ b/dt-assets/js/contact-details.js
@@ -965,7 +965,7 @@ jQuery(document).ready(function($) {
   let urlRegex = /[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/gi
   let protocolRegex = /^(?:https?:\/\/)?(?:www.)?/gi
   let resetDetailsFields = (contact=>{
-    $('.title').html(_.escape(contact.title))
+    $('.title').html(contact.title)
     let contact_methods = ["contact_email", "contact_phone", "contact_address"]
     contact_methods.forEach(contact_method=>{
       let fieldDesignator = contact_method.replace('contact_', '')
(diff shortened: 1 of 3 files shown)


corsacca:
Would it be a security issue to replace `&amp;` with & in the API for post titles instead on everywhere it is displayed?
starting with `get_post()` and `list_post()` (and the connections)

corsacca:
Possible solution: https://github.com/DiscipleTools/disciple-tools-theme/pull/1032

micahmills:
It might work but we would want to run all the XSS attacks we compiled through it to see what happens.

corsacca:
It works in the places that use the api to get the records.
I went to my "dashboard" tab from the dashboard plugin and see the issue there:
image

micahmills:
So the problem comes that sometimes we are escaping in PHP and sometimes not, and then we are again escaping in JS. 

Is it possible to force everyone's posts to go through the kses php functions, then everything is consistent coming from the backend.

corsacca:
The dashboard has custom queries so it gets its results directly from the db. Other places do too.
Because the title is store in the db as `blue &amp; red` it will be displayed as so unless each place also does the replace

mikeallbutt:
I guess it is best to have the data stored as the HTML characters and not just the & symbol.

micahmills:
I took another stab at this with #1092.

corsacca:
I'm still getting the same issue on the dashboard.

PR #1032, opened 2020-07-31
Replace "&amp;" with & in post_titles.

corsacca:


micahmills:
If #1092 works, we can close this.

PR #1038, opened 2020-08-11
filtering out duplicates from the people group compact end point.

micahmills:
This occurs when you have a translation setup for a people group, the search is looking for matches in both the title and the translation labels. If your search matches both (ie Title - American and Spanish Label - Americano and you search for American) it will return the people group twice. This compares the post ID to create a unique array that is returned.


corsacca:
Nice one @micahmills

micahmills:
Don't be to impressed it was mostly knowing what to search/copy and paste from stackoverflow 😁

Issue #1064, opened 2020-09-14
Date fields wrong timezone


corsacca:
The date field stores the time as the start of the day utc. This gets converted to the previous day in other time zones: seehttps://github.com/DiscipleTools/disciple-tools-mobile-app/issues/408

micahmills:
I looked into this a bit last night. The dates are not actually being saved at the start of the date UTC they are being saved at the start of the day for the timezone the user is in when they save the date. 

So if you save the date of October 1st in Istanbul, it saves the time stamp of 1601499600 if you save that date in San Francisco it saves the time stamp of 1601535600. 

We can fix this going forward by adding some parameters to the formatDate function and the moment.js calls (I will send in a PR shortly). But we will need to discuss what to do with past dates that are not saved at the beginning of the day UTC.

I see 2 possible solutions:
1. Leave them as they are and we will have a few cases where multi time zone teams will see things a day off.
2. On the next update run some sort of batch update that will update all the timestamps to utc. This is obviously the most complete solution, but it will take some work and I am not sure if it is worth it right now.

As a side note for anyone who might read this later. To make testing this easier you can change your timezone in Chrome without having to change your system time. See https://stackoverflow.com/a/60008052 for details.

corsacca:
Closed by #1071

PR #1071, opened 2020-10-19
Addresses #1064 going forward

micahmills:
by making all timestamps based on the start of the day utc not the local time

### DIFF
--- a/dt-assets/js/details.js
+++ b/dt-assets/js/details.js
@@ -80,7 +80,7 @@ jQuery(document).ready(function($) {
           date = " ";//null;
         }
         let id = $(this).attr('id')
-        rest_api.update_post( post_type, post_id, { [id]: moment(date).unix() }).then((resp)=>{
+        rest_api.update_post( post_type, post_id, { [id]: moment.utc(date).unix() }).then((resp)=>{
           if (this.value) {
             this.value = window.SHAREDFUNCTIONS.formatDate(resp[id]["timestamp"]);
           }
--- a/dt-assets/js/group-details.js
+++ b/dt-assets/js/group-details.js
@@ -624,7 +624,7 @@ jQuery(document).ready(function($) {
           date = $(val).val();
         }
         let id = $(val).attr('id')
-        API.update_post( 'groups', groupId, { [id]: moment(date).unix() }).then((resp)=>{
+        API.update_post( 'groups', groupId, { [id]: moment.utc(date).unix() }).then((resp)=>{
           if (val.value) {
               $(val).val(moment.unix(resp[id]["timestamp"]).format("YYYY-MM-DD"));
           }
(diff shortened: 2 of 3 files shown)


corsacca:
Hey @micahmills 
(https://disciple.tools/dev-docs/rest-api/post-types-fields-format/)
The API currently expects dates to be in 'yyyy-mm-dd' format. It then uses strtotime to convert the date into the unix time stamp which would result in a timestamp for the start of the day UTC.
Passing in the timestamp directly works, but requires the client to figure out the the right value like you've done here.
Thoughts on using yyyy-mm-dd instead?

micahmills:
We could use the yyyy-mm-dd format but we lose the internationalization and formatting that we have with the formatDate functions (https://github.com/DiscipleTools/disciple-tools-theme/blob/d915605c33f48b4c87c7fe23b4f3ed675c92d759/dt-assets/js/shared-functions.js#L379).

corsacca:
Yes, for displaying the date we should continue using the formatDate() function. Makes the user's life much nicer.
When sending the updated value to the server we need to use the unix timestamp format or the yyyy-mm-dd format. 
The api does the converting already when using receiving a date in the yyyy-mm-dd format, which is why i lean towards using it.

micahmills:
Just so I am clear we would save the date in the YYYY-MM-DD format in the database then in the formatDate function we would convert it to a time stamp for the international formatting?

corsacca:
> Just so I am clear we would save the date in the YYYY-MM-DD format in the database then in the formatDate function we would convert it to a time stamp for the international formatting?

No, the API transforms it to a unix timestamp before it is saved in the DB.

micahmills:
I believe that is what this PR does. The issue is that previously the time code were getting was midnight of the day selected in the users timezone not UTC. This should make sure that the saved time codes are for the start of day UTC. 

For previously saved dates we will need to put together a plan for either updating all the old timecode or just know that they might be a day off for teams that are working across multiple timezones and have saved dates previous to this update.

corsacca:
`moment(date).unix()` isn't correct because the time zone is off.
`moment.utc(date).unix()` is correct. 
`moment.utc(date).format('YYYY-MM-DD')` would also be correct (and what the documentation says to send to the API). The server would convert to utc and save the timestamp to the DB.

I'm fine with `moment.utc(date).unix()` and added it to the groups page too.

micahmills:
Should we do something with previously saved dates that might have the time stamp that is not utc? We could come up with a script to run through all the old dates and fix them but there will be some edge cases to plan for. Or we could just leave it as it is and know that previously saved dates might be off by a day for teams working in multiple time zones.

corsacca:
For fixing historical data:
The theme and the app have updated dates at the same time, so it will be generally hard to know which ones are correct and which ones are not.
I think we can leave it and new dates going forward will be correct.

micahmills:
Sounds good to me

PR #1076, opened 2020-11-18
The hook was moved from place to function correctly

ferleysilva:
Hi Corsac, I had to move the hook you created from line 264 to line 266 for it to work correctly, do not modify any validation of yours

### DIFF
--- a/single-groups.php
+++ b/single-groups.php
@@ -261,9 +261,9 @@ if ( ! current_user_can( 'access_groups' ) ) {
 
                                     </div>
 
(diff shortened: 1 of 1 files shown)


corsacca:
@ferleysilva I realize that i named the hook wrong. It should be `health-metrics` and it should stay in the health metrics tile.
I've added hooks for you to create non-standard tiles:
`dt_record_bottom_after_tiles` and `dt_record_bottom_below_tiles`

corsacca:
@ferleysilva  `dt_record_bottom_after_tiles`  or  `dt_record_bottom_below_tiles` fit the workflow you are going for?

PR #1077, opened 2020-11-18
Feature bulk edit

micahmills:
I have created a bulk edit feature for the Contacts and Groups page. Right now you can only Assign, Sub-assign, Share, Request and update and Follow contact or groups. 

In the sub-assign and share fields you can have multiple users, but there is not currently a way to remove a sub-assigned or shared user in bulk because currently we don't have a UI for that in this context. So if you add a user in those fields it will add it to the list, but there isn't a way to bulk remove that user at this time.

If you guys can give this your eye balls and see what you think, I also haven't tried this with hundreds of updates to see the performance yet. 

Here are a list of things I would like to add to it, let me know if there are other things that need to be added.

1. Add a Status update field for bulk updating the status of contacts and groups.
2. Add all the other fields custom and otherwise so they an be updated in bulk.
3. Think through a UI for removing sub-assigned and shared users in bulk.


corsacca:
I'm excited to review this and test it out. Thanks Micah! I'll get back to you when we have.

micahmills:
To make your review a little easier I can explain what is happening to make this work.

I largely took the JS for the typeaheads and other fields from the details.js files. The only thing that was changed other than names and such, was that instead of firing the update API when the field is completed, it instead saves it to a data attribute on the input labeled 'bulk_key_' and then the name of the field the field. 

When you click the update button it finds the ID of all the checked contacts/groups and gets all the bulk_key_ fields from the various bulk inputs. From there is compiles the update payload and fires that off to the API. The share function is a bit different so it has to do its own API call, but does basically the same thing.

I did update the render_field_for_display php function to add some optional params so it would output the bulk fields as needed.

corsacca:
Hey,
I like the UI for on the list with the checkboxes!

Help me understand the need to update the keys in  `render_field_for_display`.
Is there a conflict with the field inputs in the filter pannel?
I don't think the update the fields would call the update API functions because those listeners are only loader on the record details page in the details.js.

The page is refreshing for me before the updates are done, so it only gets about ~16 of the 100

micahmills:
the render_field_for_display changes are because those fields are used in the filters modal. I was having issues with the IDs duplicating and the JS for the filters being called instead.

I will look into he the issue of refreshing before all 100 are done. It is probably a promise problem.

micahmills:
I can't seem to recreate the problem with the page refreshing before all the items are updated. It is possible that it is a promise issue, but I can't seem to track it down. Lets chat tomorrow and see if we can figure out what is causing it.

corsacca:
It does look like the promine().then() is being called before the all the api calls are done.

micahmills:
I have refactored to use Promise all and that seems to take care of the problem. 

I have noticed that somewhere around 400 API calls they start to fail. I suspect this is a limitation of the server, I will look to see if there is a way to delay these so that they don't DDOS the server.

corsacca:
Can we do a queue of 10?
https://github.com/DiscipleTools/disciple-tools-import-export/blob/81139ebe0b9be8ff7fff76504b8b3a8b0d56ff9b/includes/admin/admin-menu-and-tabs.php#L1066
the import plugin has an example that we could harvest.

micahmills:
That is brilliant. I think I have it working. I just ran it through 1000 contacts changing the Assigned_to Subassigned_to share update_required, and follow values. It took 3 1/2 minutes but it got through them all. 

I haven't tracked it down fully but it appears that if you share a contact with someone that it is already shared with it throws an error and it shows up in the console. I will look to see if we can keep that from happening.

dt-core/global-functions.php
```diff
@@ -335,20 +336,28 @@ function render_field_for_display( $field_key, $fields, $post, $show_extra_contr
                 return;
             }
             if ( isset( $post['type']["key"], $fields[$field_key]["only_for_types"] ) ) {
-                if ( !in_array( $post['type']["key"], $fields[$field_key]["only_for_types"] ) ) {
-                    return;
+                if ( is_null( $post ) && $is_bulk_edit) {
```
corsacca, line 339: I don't this is_null() will ever be true based on the isset() above it.

corsacca:
I like how this is looking. The queue worked well in the 100 contacts i tested.
I created a PR for a suggestion on the render_field_for_display().
I wonder if the added prefix should only be added to the top level id="" on the field and not the lower level classes and data- attributes.

micahmills:
> I wonder if the added prefix should only be added to the top level id="" on the field and not the lower level classes and data- attributes.

I think that is fine for the simple inputs, but the typeaheads are fairly complex and there is a lot of markup and JS that I am not 100% sure what it does or how it does it. For the typeaheads I found that I needed to change more than just the top lever id to make things work. 

As for the data-attributes that I am adding to the fields, because I wasn't sure exactly what all is/might be done with those fields from jquery plugins or other JS in the project, I wanted to make sure to name space the attributes to make them unique and not overlap with anything else. 

That being said, If you see a way to make it more simple I am all for that.

micahmills:
For everyone that has followed this so far. I have added (almost) all the other fields to the bulk edit section. Tags, and Location are a bit more complicated and I will try to get those working. If you are so inclined please have a look at this and see if you find anything that doesn't work properly or you don't like the design of.

corsacca:
Sight bug: 
I have more than 1 selected
image

corsacca:
When the button has a number the loading spinner does not appear.

corsacca:
The overall status field does not apply to groups:
I think we should add a check to see if the field is available/enabled before display it.
image

corsacca:
Getting: `Undefined variable: contact in disciple-tools-theme/archive-template.php on line 234`

micahmills:
> When the button has a number the loading spinner does not appear.

Which loading spinner?

corsacca:
When the button has a number in it this loading spinner does not show:
image

micahmills:
@corsacca I believe that should take care of the bugs you found

corsacca:
Thank you @micahmills !!

PR #1081, opened 2020-12-03
Reporting upgrade

ChrisChasm:


dt-mapping/mapping-queries.php
```diff
@@ -1488,5 +1488,475 @@ public static function query_under_location_grid_meta_id( $post_type, $grid_id,
         return $results;
     }
 
+    /**
+     * Performance query for contacts location_grid without dependency on location_grid_meta
+     * @param null $status
+     * @return array
+     */
+    public static function query_contacts_location_grid_totals( $status = null ) {
```
corsacca, line 1496: @ChrisChasm  Why add these back in?

dt-core/migrations/0038-add-report-columns.php
```diff
@@ -0,0 +1,42 @@
...
+        $wpdb->query( "ALTER TABLE $wpdb->dt_reports ADD `parent_id` BIGINT(22) NULL DEFAULT NULL AFTER `id`" );
+
+        // add_posttype
+        $wpdb->query( "ALTER TABLE $wpdb->dt_reports ADD `post_type` VARCHAR(20) NULL DEFAULT NULL AFTER `post_id`" );
+
+        // move hash
+        $wpdb->query( "ALTER TABLE $wpdb->dt_reports MODIFY COLUMN `hash` VARCHAR(65) NULL DEFAULT NULL AFTER `timestamp`" );
+
+        // add parent_id index
+        $wpdb->query( "ALTER TABLE $wpdb->dt_reports ADD INDEX `parent_id` ( `parent_id` )" );
```
corsacca, line 21: @ChrisChasm we need to check if the index does not already exists.

PR #1084, opened 2020-12-07
add api endpoint to log activity from front-end

cairocoder01:
In order to facilitate greater use of the activity logging, especially for actions that only happen in the client, this exposes an endpoint to save activity logs.

### DIFF
--- a/dt-core/core-endpoints.php
+++ b/dt-core/core-endpoints.php
@@ -57,12 +57,39 @@ class Disciple_Tools_Core_Endpoints {
 
 
     /**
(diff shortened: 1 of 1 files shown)


corsacca:
@cairocoder01 I like the idea. 
Can we add some logic to make sure:
- the user can't create activity as other users (the user id must be the current authenticated user). 
- the user can't create activity on contacts they do not have access to

cairocoder01:
@corsacca Good thinking. Added those checks now.

corsacca:
Beautiful!

PR #1086, opened 2020-12-08
Site Links - Site URL incorrect for subdirectory multi-sites

### LINKED ISSUE #1085: Site Links - Site URL incorrect for subdirectory multi-sites
When running a multi-site instance that uses a subdirectory structure instead of subdomain, using the "add this site" button in the new site link page adds the root site instead of the current site. 

For example, the multisite is hosted on mydomain.com with sub-sites on mydomain.com/site1 and mydomain.com/site2, the "add this site" button will always insert "mydomain.com" instead of the appropriate site URL.

This also affects the ajax request to confirm the connection status. It always uses the root domain and thus does not give an accurate status message. It always returns "A valid URL was not provided" when on a sub-site.

Since I updated that ajax request a while back, I will attempt to take a look at this issue soon.

cairocoder01:
resolves #1085


corsacca:
Thanks @cairocoder01 !

PR #1092, opened 2020-12-14
HTML Entities in Titles

micahmills:
uses wp_specialchars_decode() to fix the HTML encoded characters in contact and group Titles.

wp_specialchars_decode() will handle &, <, >, ", and ‘.

The other PRs for this were a start, but were flawed. So I just started over. I believe I hit all the places that post titles are shown. I checked the Dashboard plugin, lists pages, and individual pages, and various metrics pages.

I also checked it against XSS injections and it seems to hold up, but that is always worth double checking.


ChrisChasm:
Thanks Micah, checking this out.

ChrisChasm (approved):

PR #1105, opened 2021-01-14
fixed bug that rounds time upwards and doesn't take time gaps

prykon:
Before, everything was getting rounded upwards.
eg. 3.2 weeks was considered = 4 weeks.

Additionally, the checks where being made on < 4 weeks ago and > 4 weeks ago but if something was exactly 4 weeks ago it wouldn't appear. This is a problem specially on notifications that fall between the cracks in terms of being equal to X amount of years.

### DIFF
--- a/dt-notifications/notifications.php
+++ b/dt-notifications/notifications.php
@@ -70,7 +70,8 @@ class Disciple_Tools_Notifications
         }
 
         return self::$_instance;
-    } // End instance()
+    }
+    /** End instance() */
 
     /**
      * Constructor function.
@@ -118,7 +119,7 @@ class Disciple_Tools_Notifications
             [ '%d', '%d', '%d', '%d', '%s', '%s', '%s', '%s', '%d', '%s', '%s' ]
         );
 
-        // Fire action after insert.
+        /** Fire action after insert */
         do_action( 'dt_insert_notification', $args );
     }
 
@@ -156,7 +157,7 @@ class Disciple_Tools_Notifications
             ]
         );
 
-        // Final action on insert.
+        /** Final action on insert */
         do_action( 'dt_delete_notification', $args );
     }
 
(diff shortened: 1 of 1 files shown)


dt-notifications/notifications.php
```diff
@@ -783,7 +815,7 @@ public static function get_notification_message_html( $notification, $html = tru
             $notification_note = sprintf( esc_html_x( '%1$s subassigned %2$s to you.', 'User1 subassigned contact1 to you.', 'disciple_tools' ), $display_name, $link );
         } elseif ( $notification["notification_name"] ==="milestone" ){
             $meta_value = $notification["field_value"] ?? '';
-            $contact_fields = DT_Posts::get_post_field_settings( "contacts", false ); //no cache to get the labels in the correct language
+            $contact_fields = Disciple_Tools_Contact_Post_Type::instance()->get_custom_fields_settings( true, null, false, false ); //no cache to get the labels in the correct language
```
corsacca, line 818: @prykon is no longer valid in v1 :
Disciple_Tools_Contact_Post_Type::instance()->get_custom_fields_settings(

corsacca: @prykon will you be able to update this?

prykon: @corsacca What is the v1 equivalent of this function? Thanks!

corsacca: What you had there is actually correct.

prykon: Reverted.

dt-notifications/notifications.php
```diff
@@ -432,10 +466,8 @@ public static function send_notification_on_channels( $user_id, $notification, $
         }
         if ( !in_array( 'email', $already_sent ) && dt_user_notification_is_enabled( $notification_type, 'email', $user_meta, $user_id ) ) {
             $user = get_userdata( $user_id );
-            if ( $user ){
```
corsacca, line 435: Why get rid of the user check. If a user is not found, the $user->user_email will throw an error.
@prykon

PR #1110, opened 2021-01-19
Fix padding and width on contact creation type boxes

micahmills:
per @mikeallbutt's comment in slack

https://disciple-tools.slack.com/archives/CDXNVC892/p1611026755024400


corsacca:
Thanks @micahmills

Issue #1123, opened 2021-01-24
Fields selector on mobile devices is difficult visually 


mikeallbutt:
Because there are so many options available when adjusting what fields to show on the record list page, it is difficult to work out which box is for which option. Is the item listed referring to the box above or below? 

![Screenshot_20210124-053406.png](https://user-images.githubusercontent.com/43966676/105629103-a3737680-5e06-11eb-8a9a-1b07a7794cfb.png)

PR #1124, opened 2021-01-24
fixes issues #1123

micahmills:
puts the checklist in one column if below 640px.

Also moved style from inline HTML styling to CSS


corsacca:
Thanks @micahmills !

PR #1129, opened 2021-01-24
Mobile view styling

micahmills:
This contains several minor changes that @mikeallbutt found during mobile testing.

It also contains a change to the top and second bar. First, I fixed the sticky headers so they work properly on all sizes. To do I used the "mobile" top blue bar on medium size screens because it was causing weird wrapping on the smallest medium screens.

I also added the record title into the second bar to make it easier to see what record you are working on if you are scrolled down the page. 

In adding the title I realized that I could also add all the data that was in the contact-type tile and animate the unneeded information away when the users scrolls and the menu is stuck to the top. Because I moved the contact-type tile info to the second bar, I deleted the contact-type tile. 

@corsacca you might have a different opinion on this, but I like having the data in the top bar instead of another tile. That being said, we could also keep the tile and only display it on medium to large screens and save the space on mobile. Or we could just put the record name in the bar and leave the tile in place.


corsacca:
Before Mobile:
image
After Mobile:
image
image

corsacca:
Before Desktop:
image


After Desktop:
image

corsacca:
If you refresh on mobile when while already scrolled down, and then scroll back up:
image
The top blue bar is missing.

corsacca:
The name tile and person icon change color depending on the contact type. I'd like to keep something similar:
image
image
image

After stays the same:
image

corsacca:
@micahmills Great questions and suggestions.
I like where you are going with this. 
I really like how the sticky bar on mobile and the extra fields disappearing.
I've added screen shots above to visualize what is happening.
We could also play with putting the "[Type] record, created on ..." in the details tile.

corsacca:
image
Maybe beyond the scope of this pr: I we might need to put "share, following, and tasks" in a dropdown menu to clean up the ui a bit on mobile.

corsacca:
Note, in the tile the record name is editable.

micahmills:
I have added the icon color as well as added the icon to the mobile view. I also reimplemented the title editable from the secondary bar. I also fixed the bug that caused the menu not to appear if it loaded while scrolled down the page.

Something to not, the sticky bars can possible get messed up  if you resize the window but usually a scroll to the top or refresh fixes that. I don't really consider that a bug, because most users don't resize windows on a regular basis.

micahmills:
I like the idea of consolidating the action buttons on the mobile view, but I also like the comments button that makes it quick to scroll to the comment section. I think we should probably make a separate issue/PR for that and see if we can work something out there.

corsacca (approved):

single-template.php
```diff
@@ -46,45 +46,19 @@
...
                     <section class="cell small-12 update-needed-notification"
                              style="display: <?php echo esc_html( ( isset( $dt_post['requires_update'] ) && $dt_post['requires_update'] === true ) ? "block" : "none" ) ?> ">
-                        <div class="bordered-box detail-notification-box" style="background-color:#F43636">
-                            <h4>
-                                <img src="<?php echo esc_html( get_template_directory_uri() . '/dt-assets/images/alert-circle-exc.svg' ) ?>"/>
-                                <?php echo esc_html( sprintf( __( 'This %s needs an update.', 'disciple_tools' ), strtolower( $post_settings["label_singular"] ) ) ) ?>
-                            </h4>
-                            <p><?php esc_html_e( 'Please provide an update by posting a comment.', 'disciple_tools' )?></p>
-                        </div>
+                        <a href="#comment-activity-section">
```
corsacca, line 49: @micahmills Brilliant!
this one is a bit weird on desktop, is just scrolls the page down some:

corsacca: image

micahmills: Yeah that would be a bit weird because presumably the comments section is on the screen in the second column on the desktop view. Let me think about this and I am sure we can come up with a way to handle it.

single-template.php
```diff
@@ -55,6 +55,13 @@
                                 <p><?php esc_html_e( 'Please provide an update by posting a comment.', 'disciple_tools' )?></p>
                             </div>
                         </a>
+                        <div class="bordered-box detail-notification-box show-for-large" style="background-color:#F43636">
+                                <h4>
+                                    <img src="<?php echo esc_html( get_template_directory_uri() . '/dt-assets/images/alert-circle-exc.svg' ) ?>"/>
+                                    <?php echo esc_html( sprintf( __( 'This %s needs an update.', 'disciple_tools' ), strtolower( $post_settings["label_singular"] ) ) ) ?>
+                                </h4>
+                                <p><?php esc_html_e( 'Please provide an update by posting a comment.', 'disciple_tools' )?></p>
+                            </div>
```
micahmills, line 64: Not the most elegant solution, but it works. If we can think of a way to do this without having to duplicate the notification box HTML that would be ideal, but we are doing the same things with menus and other places so it isn't the end of the world.

corsacca:
Noticed it here, but not a result of this PR. The Person icon is also used on  groups and  trainings. Before we had different icons.

single-template.php
corsacca: Thanks! This works great

Issue #1179, opened 2021-02-03
Have a `Save tile and field order` button at the top of the `Sort Tiles and Fields` tile


mikeallbutt:
Because the `Sort Tiles and Fields` on the /wp-admin/admin.php?page=dt_options&tab=custom-tiles page, the save button at the bottom of the tile is not visible many items. Have a button at the top and bottom will help the users.

mikeallbutt:
The same idea would be good for the `New record page` which has a `Save` button at the bottom of each tile.

PR #1182, opened 2021-02-04
add save button at top for custom tiles

mikeallbutt:
fix for #1179


corsacca:
@mikeallbutt can you paste a screenshot here of what this does?
A second button needs to have a different id, or the id needs to be turned into a class.

mikeallbutt:
Most times there are many columns so the save button is way down below and out of the users view.

Screen Shot 2021-02-04 at 02 07 42

corsacca:
@mikeallbutt thanks for the screenshot. Before i can merged this PR I expect you to fix the button id conflict.

mikeallbutt:
When I tested the buttons they worked as expected - even without the id's being set. I would add the ids but I don't know what to do (how to code it) so I would appreciate some help from someone who can do it for me.

micahmills:
Hey @mikeallbutt having elements with the same ID on a page will work on some browsers, but it will break things in weird unexplainable ways on other browsers. 

The best thing to do would be to change the id of both buttons `id="save-drag-changes"` to `class="save-drag-changes"` You will then also need to change the corresponding javascript to look for the class instead of the ID. That can be found on line 32 of `/dt-core/admin/js/dt-options.js` https://github.com/DiscipleTools/disciple-tools-theme/blob/77ce110e025a24fbd3fe2608c37c451ea6cd71a8/dt-core/admin/js/dt-options.js#L31

You will just need to change `#save-drag-changes` to `.save-drag-changes`

corsacca:
Closed by an intern at #1204

PR #1188, opened 2021-02-04
Generic Note Icons

mikeallbutt:
These icons can be used to display next to fields where no icon has been defined.


corsacca:
@mikeallbutt this PR contains commits that have been squashed and already merged in.
The correct workflow is to delete a branch after it has been merged (see https://github.com/DiscipleTools/Documentation/blob/master/Code-Contribution/From%20Fork%20to%20Pull%20Request.md) 

With your new PR, please provide a screen shot of what the icon looks like. Also what were exactly is this icon to be used?

PR #1206, opened 2021-02-10
Whatsapp Icon

mikeallbutt:
Added a whatsapp icon into the assets folder so it can be used in the details tile if needed.


corsacca:
Hey man, thanks for the icon. I suggest dropping icons here to be available when they are needed:
https://github.com/DiscipleTools/disciple-tools-theme/discussions/1208
We can then easily grab them and add them to the theme when needed.

Also:
We can not merge this PR because it contains commits that have been squashed and already merged in.
The correct workflow is to delete a branch after it has been merged (see https://github.com/DiscipleTools/Documentation/blob/master/Code-Contribution/From%20Fork%20to%20Pull%20Request.md)

PR #1212, opened 2021-02-16
Advanced Search

micahmills:
This gives an option for advanced searching of specific text fields in Contacts and Groups, and lays the groundwork for a new global search endpoint for the future.


archive-template.php
```diff
@@ -28,14 +28,67 @@ class="second-bar list-actions-bar">
...
+                ?>
+                <ul class="ul-no-bullets" style="">
+
+                <li style="" class="">
+                    <label style="margin-right:15px; cursor:pointer">
+                        <input type="checkbox" value="all"
+                                <?php echo esc_html( in_array( 'all', $fields_to_search ) ? "checked" : '' ); ?>
+
+                                style="margin:0">
+                        <b>Search All Fields</b>
```
corsacca, line 69: Translate

archive-template.php
```diff
@@ -50,10 +103,63 @@ class="second-bar show-for-small-only center list-actions-bar"><!--  /* MOBILE V
...
```
corsacca, line 140: translate

corsacca:
image
I would expect select one of the options to limit the search to that field. So if comments is selected, it would only search for record with "bob" in the comments.

archive-template.php
micahmills: I have fixed this

archive-template.php
micahmills: I have fixed this

micahmills:
> image
> I would expect select one of the options to limit the search to that field. So if comments is selected, it would only search for record with "bob" in the comments.

I think I have fixed this issues.

I also added a badge icon and number when an advanced fields search is selected so you can tell if they are toggled without having to open it.

corsacca:
Thanks @micahmills !

PR #1214, opened 2021-02-17
People Group Typeahead fix

### LINKED ISSUE #1157: Clickable People Group item resolves to 404 page
In the `people group` field the items are clickable but take you to a page that doesn't exist so a 404 is generated.

In the attached image, this is where it looks when the word `Mexican` is clicked. 

https://oikos.disciple.tools/peoplegroups/159

micahmills:
don't include links for peoplegroups typeaheads. fixes #1157

### DIFF
--- a/dt-assets/js/details.js
+++ b/dt-assets/js/details.js
@@ -271,14 +271,21 @@ jQuery(document).ready(function($) {
         return parseInt(item.ID) !== parseInt(post_id)
       },
       source: window.TYPEAHEADS.typeaheadPostsSource(listing_post_type, {field_key:field_id}),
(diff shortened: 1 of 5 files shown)


corsacca:
image
Other fields now look like this

corsacca:
I think we also have a permission error. I tried to add a translation on a people group, but got an insufficient permissions error.

micahmills:
> image
> Other fields now look like this

I think the latest commit fixes this.

corsacca (commented):
@micahmills did you see these two?

corsacca:
Thanks @micahmills

Issue #1066, opened 2020-09-21
Notifications timestamp not being calculated with system timezone


mikeallbutt:
The date when the notification was logged is not taking into account the timezone that the DT instance is set to use.
eg. this image is from an instance with -5 GMT and the notification should be just 10 minutes early - and not 5 hours.

<img width="256" alt="Screen Shot 2020-09-21 at 12 32 06 pm" src="https://user-images.githubusercontent.com/43966676/93800915-e1694f80-fc06-11ea-9f51-34a022f1f451.png">

corsacca:
Notifications should be based on the user's time. What do you mean by "10 minutes early"?

mikeallbutt:
> Notifications should be based on the user's time. What do you mean by "10 minutes early"?
> How is what is show different than what you expect to see?

It should say: less than 1 hour ago (or 10 minutes ago) and not 5 hours ago.

PR #1215, opened 2021-02-17
Notification Timezone bug fix

micahmills:
Use WP time settings instead of server settings.
fixes issues #1066


corsacca:
Thanks @micahmills!

PR #1216, opened 2021-02-17
autofocus

disciple-tools:


corsacca:
@disciple-tools thanks for the Pull request!
I've updated the js to remove the console logs. If a modal does not contain an input (like the help modals) it is ok that nothing happens.

PR #1217, opened 2021-02-18
added whatsapp and telegram svgs to the lists

micahmills:
Add Telegram and WhatsApp logos


corsacca:
I wonder if we could point to a CND where users could copy the link to different icons.
This would add flexibility for any icon to be used without D.T having to host them.

micahmills:
I am going to close this because I think we probably need to have a more robust icon system to resolve this problem.

PR #1219, opened 2021-02-23
RTL phone number fix

micahmills:
When the users language is RTL and they enter a phone number with a international prefix with the + the order was not correct. This adds the dir:auto to all inputs and aligns it right for rtl languages. This should allow users to type in RTL or Latin numbers and get things in the proper order and alignment.


dt-assets/js/details.js
```diff
@@ -145,9 +145,9 @@ jQuery(document).ready(function($) {
     const $list = $(`#edit-${field}`)
 
     $list.append(`<div class="input-group">
-            <input type="text" data-field="${_.escape( field )}" class="dt-communication-channel input-group-field" />
+            <input type="text" data-field="${_.escape( field )}" class="dt-communication-channel input-group-field" dir="auto" />
             <div class="input-group-button">
-            <button class="button alert input-height delete-button-style channel-delete-button delete-button new-${_.escape( field )}" data-key="new" data-field="${_.escape( field )}">&times;</button>
+            <button class="test button alert input-height delete-button-style channel-delete-button delete-button new-${_.escape( field )}" data-key="new" data-field="${_.escape( field )}">&times;</button>
```
corsacca, line 150: @micahmills what is the `test` class for?

micahmills: To prove that I am fallible 😁. I was trying to figure out if it was where the input was created and added the class as a test and forgot to remove it.

corsacca: :)

PR #1224, opened 2021-02-24
replace all instances of `_.` with `window.lodash.` to address bug #688.

### LINKED ISSUE #688: Lodash is overwritten when Jetpack plugin is installed.
This keeps the contacts list form loading and also keeps other parts of the theme from working.

micahmills:
see https://make.wordpress.org/core/2018/12/06/javascript-packages-and-interoperability-in-5-0-and-beyond/ for more details

fixes #688


ChrisChasm:
Holy smokes. :) I'm signing off for the night but nice work and thanks. I think we work through this tomorrow.

micahmills:
Yeah we will really want to test this out to make sure we don't break anything with a giant find and replace...But it appears to work. 

This will have implications for all the plugins as well since many of them use lodash as well.

PR #1226, opened 2021-03-02
Lodash preprocess. fixes #688

### LINKED ISSUE #688: Lodash is overwritten when Jetpack plugin is installed.

micahmills:
To fix #688 we need to use `window.lodash.` instead of` _.` but that makes the code ugly and harder to work with in dev.

This uses gulp to run a find and replace for `_.` and replace with `window.lodash.` and also minifies them and moves them to the build/ folder and changes the wordpress enqueues to use the minified scripts.


corsacca:
image
On the lists page.

micahmills:
> image
> On the lists page.

This was because uglify doesn't like es6 notation. So I changed to using terser instead.

PR #1230, opened 2021-03-13
Fixes HTML entities in typeaheads. Fixes #1176

### LINKED ISSUE #1176: The HTML & amp; symbol is showing in lists
There are places where the & amp; symbol is showing as HTML and it should not.

- the name of a member in the group members list

- drop down lists

micahmills:
I think this will finally put #1176 to bed.


corsacca:
Also the xss test should not fail on comments either now.

corsacca:
Does this look good to you @micahmills ?

micahmills:
Just got a chance to test this and it seems to work perfectly

PR #1232, opened 2021-03-13
Fixes Help Modal Text not appearing on settings page

### LINKED ISSUE #1173: Help icon opens empty modal on the Your Profile settings page
Clicking the ? icon in the top right corner of the Your Profile tile of the users settings page opens a modal but it is completely empty.

Screen Shot 2021-02-02 at 17 19 36

The same happens in the notifications tile down below.

Screen Shot 2021-02-02 at 17 22 27

micahmills:


micahmills:
This is probably not the most elegant solution for this problem, but the post type check at the beginning of the file was returning empty on the settings page. We could possibly add a check to see if we are on the settings page, but this seemed the easier and will work if we have other pages that are not post types in the future.

Issue #1049, opened 2020-09-08
WP-admin: Custom Fields: New field type: textarea


mikeallbutt:
Having a `textarea` field will allow the users to add more areas to write other types of data like comments that wrap onto new/more lines.

![Screen Shot 2020-09-07 at 7 27 47 am](https://user-images.githubusercontent.com/43966676/92425387-b5e95e00-f14c-11ea-8f33-fec4e6ea95c7.png)

corsacca:
Implemented by #1245

PR #1245, opened 2021-03-24
implemented a textarea 

micahmills:
implemented the JS and PHP to create the input and save the text.

This addresses #1049.


micahmills:
This seemed to easy so we need to make sure we double and triple check that this doesn't break something.

corsacca:
Thanks for taking this one on!

I created a textarea in the wp admin.

Here are areas we'll need to look at:
- The value from the create new page was not saved.
- Updating the value on the record should create an activity message.
- When adding the field to the list page the cell remains empty.

micahmills:
@corsacca I fixed the places you mentioned. I have looked for other places that might be a problem don't see any.

corsacca (approved):

corsacca:
Thanks @micahmills !

PR #1247, opened 2021-03-25
fix: add display: none; to hide #hide-hidden-fields button initially

squigglybob:
fixes the 'hide fields' button showing initially on the add group page.


corsacca:
Thanks @squigglybob !

PR #1254, opened 2021-03-26
1145 apply tag directly from suggest list

### LINKED ISSUE #1145: Create and apply a tag directly from the tag auto-suggest list
If you wanted to apply a tag to a record but as you typed you realised that the tag didn't yet exist, then it would be nice to be offered an option to create and apply that tag directly from the autosuggest list. The current icon to create new tags could stay but probably won't be needed too much.

squigglybob:
resolves #1145

### DIFF
--- a/dt-assets/js/details.js
+++ b/dt-assets/js/details.js
@@ -751,14 +751,13 @@ jQuery(document).ready(function($) {
         if (this.comparedItems.includes(query)) {
           return tagExistsText.replace('%s', query)
         }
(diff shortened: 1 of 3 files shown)


squigglybob:
as long as you're happy with how I've used the `emptyTemplate` functionality of the typeahead and the visual look of the button, it should be a goer @corsacca

corsacca:
Thanks @squigglybob !

PR #1255, opened 2021-03-26
537 react to comments

### LINKED ISSUE #537: Comments: like, favorite or react to a comment.
**Purpose**: Add clarity and usability to the comments area.

**Benefits**:
- Reactions will visually show that a user has seen the comment without having to write a whole reply.
- Reactions will allow the user to quickly scan the comments to find the most importand ones.
- Reactions will improve communication by allowing users to easily respond to any comment.

**Requirements**:
The user can react to any comment on a record.
The user can choose between 6 options of reaction.
If a comment has reactions they will be displayed under the comment. Each reaction icon with a number associated to it.
If the user hovers over (or clicks on for mobile) the reaction icon, it will show a list of users who reacted.
Default  reaction options (same as Facebook):  thumbs up, heart, laugh, wow, sad face, angry
Customization: add filters to allow for a plugin to add or remove from the list.

**Code considerations**:
Start with `dt-assests/js/comments.js`, `dt-assets/parts/loop-activity-comment.php`, `dt-posts/dt-posts.php` and `dt-posts/dt-posts-endpoins.php`
Reactions will likely be stored in the comment meta. This feature will need to extend the existing comment functions to include getting meta data.

squigglybob:
resolves #537


squigglybob:
Hi @corsacca @ahillbilly I think I'm done with this one if you could check it out.

Let me know if any of the styling needs adjusting.

Also @corsacca I wasn't sure what the best permissions to check for on the API of this, so have just used the `self::can_update(post, id)` function for now let me know if it needs anything else

dt-assets/functions/enqueue-scripts.php
```diff
@@ -197,9 +197,15 @@ function dt_site_scripts() {
                         "delete" => strtolower( __( "Delete", "disciple_tools" ) ),
                         "translate" => __( "Translate with Google Translate", "disciple_tools" ),
                         "hide_translation" => __( "Hide Translation", "disciple_tools" ),
+                        "reaction_title_1" => _x( '{{user}} reacted with {{emojiName}} emoji', 'Bob reacted with heart emoji (dont translate words in brackets)', 'disciple_tools' ),
+                        "reaction_title_many" => _x( '{{users}} and {{user}} reacted with {{emojiName}} emoji', 'Bob, Bill and Ben reacted with heart emoji (dont translate words in brackets)', 'disciple_tools' ),
                     ],
                     'current_user_id' => get_current_user_id(),
                     'additional_sections' => apply_filters( 'dt_comments_additional_sections', [], $post_type ),
+                    /**
+                     * reaction aliases must be lowercase with no spaces
```
micahmills, line 206: Just out of curiosity why do the aliases need to be lowercased?

dt-assets/js/comments.js
```diff
@@ -359,6 +387,62 @@ jQuery(document).ready(function($) {
...
+    })
+    document.querySelectorAll('.comment-reaction').forEach((element) => {
+      element.addEventListener('click', (e) => {
+        const commentId = e.target.dataset.commentId
+        const userId = commentsSettings.current_user_id
+        const reaction = e.target.dataset.reactionValue
+        rest_api.toggle_comment_reaction(postType, postId, commentId, userId, reaction)
+      })
+    })
+  }
```
micahmills, line 429: This uses vanilla javascript (which I prefer), but we are using jquery throughout the project. I personally don't think it is worth changing and the vanilla JS approach is better moving forward, but we should have @corsacca weigh in on if we should use jquery functions here for consistency with the rest of the project.

micahmills (commented):
Hey @squigglybob this is great! It works exactly as I would expect and even works in a right to left language!

One question I have is what is the advantage of using the images of emojis instead of the HTML entities directly? I would think we could save the HTTP requests for the images and the user would get the platform specific emoji designs if we used the HTML entities, but I might be missing something that requires the images.

I did notice one edge case styling thing. If you have a lot of emojis selected and the screen size is small enough the comment controls flow to multiple lines making the emoji pill buttons expend to multiple lines.

Screen Shot 2021-04-14 at 9 11 02 PM
Screen Shot 2021-04-14 at 9 39 26 PM

squigglybob:
I'll have a look at the small screen size issue, and see if I can get the edit/delete controls to go to the next line on smaller screens.

I'll also investigate just using the emoji utf8 charachters instead of paths, I guess one benifit is that any emoji image can be used, and it's not limited to what you can find in the utf8 emoji list (although it is pretty extensive :)

squigglybob:
image
Looks better on an iPhone 5 size now (320px!!!)

dt-assets/functions/enqueue-scripts.php
corsacca: The dev @corsacca decided that all keys in D.T should be snake_case for consistency.

dt-assets/js/comments.js
corsacca: plain JS is totally cool :)

corsacca:
Thanks for this great work @squigglybob and the review @micahmills 
I'm getting this error when adding a comment.
The submit comment button stays disabled, keeping me from adding another comment.
image

micahmills:
> Thanks for this great work @squigglybob and the review @micahmills
> I'm getting this error when adding a comment.
> The submit comment button stays disabled, keeping me from adding another comment.
> image

I am seeing the same thing with an console error of 
`TypeError: Cannot convert undefined or null to object
    at Function.entries (<anonymous>)
    at eval (lodash.templateSources[0]:39)
    at r (lodash.min.js?ver=4.17.19:9)
    at Function.hf (lodash.min.js?ver=4.17.19:83)
    at eval (lodash.templateSources[0]:18)
    at comments.js?ver=1618482543:371
    at Array.forEach (<anonymous>)
    at display_activity_comment (comments.js?ver=1618482543:345)
    at Object.<anonymous> (comments.js?ver=1618482543:30)
    at e (jquery.min.js?ver=3.5.1:2)`

squigglybob:
Ok, I've added in a default object for the comment reactions so hopefully that fixes the error you both found in the previous two comments. Let me know if that does it...

It's strange that I didn't get the same error on my machine Ubuntu 20.04, Chrome 88.0.4324.150

micahmills:
This seems to have fixed the issue, and the addition of the UTF-8 emojis is great. 

This looks good to go. Great work!

dt-assets/functions/enqueue-scripts.php
```diff
@@ -197,9 +197,18 @@ function dt_site_scripts() {
                         "delete" => strtolower( __( "Delete", "disciple_tools" ) ),
                         "translate" => __( "Translate with Google Translate", "disciple_tools" ),
                         "hide_translation" => __( "Hide Translation", "disciple_tools" ),
+                        "reaction_title_1" => _x( '{{user}} reacted with {{emojiName}} emoji', 'Bob reacted with heart emoji (dont translate words in brackets)', 'disciple_tools' ),
```
corsacca, line 200: @squigglybob can you change the {{key}} to use the %1$s,  %2$s etc
this will be more standard with the rest of the translation strings.
example: `'showing_x_items_matching' => _x( 'Showing %1$s items matching %2$s', 'Showing 30 items matching bob', 'disciple_tools' ),`

corsacca (changes requested):

corsacca:
Hey. Thanks bearing with me on one more request.
Can you make sure all the strings are being escaped?
One example is 
`${reaction.name}`
`reaction.name` can be a translated value, so it could contain js.
As much as possible we try to run all variables through `window.lodash.escape()` to deal with any xxs attacks, even ones that likely aren't (like reaction.path).

( Using `<%-` in the commentTemplate i think does escaping )

squigglybob:
@corsacca I've had a look through the `comments.js` file and think that the only one that needed escaping was the one mentioned in your change request.

The only other ones are in the lodash template and being escaped by the <%- tags

corsacca (approved):

corsacca:
Thank you @squigglybob !
I'm looking forward to using this one :)

PR #1256, opened 2021-03-26
Issue 726

cwuensche:
I am still testing on custom posts but this works on Contacts and Groups.

### DIFF
--- a/dt-assets/functions/details-bar.php
+++ b/dt-assets/functions/details-bar.php
@@ -20,13 +20,13 @@ function dt_print_details_bar(
 ) {
     $dt_post_type     = get_post_type();
     $post_id          = get_the_ID();
-    $post_settings    = DT_Posts::get_post_settings($dt_post_type);
-    $dt_post          = DT_Posts::get_post($dt_post_type, $post_id);
-    $shared_with      = DT_Posts::get_shared_with($dt_post['post_type'], $post_id);
+    $post_settings    = DT_Posts::get_post_settings( $dt_post_type );
+    $dt_post          = DT_Posts::get_post( $dt_post_type, $post_id );
+    $shared_with      = DT_Posts::get_shared_with( $dt_post['post_type'], $post_id );
     $shared_with_text = '';
 
     foreach ( $shared_with as $shared ) {
-        $shared_with_text .= sprintf( ', %s', $shared[ 'display_name' ] );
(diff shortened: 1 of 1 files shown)


dt-assets/functions/details-bar.php
```diff
@@ -112,8 +118,8 @@ class="second-bar" id="second-bar-large">
                         <?php if ( $share_button ): ?>
                         <div class="cell shrink center-items ">
                             <button class="center-items open-share">
-                                <img class="dt-blue-icon" src="<?php echo esc_url( get_template_directory_uri() . "/dt-assets/images/share.svg" ) ?>">
-                                <span style="margin:0 10px 2px 10px"><?php esc_html_e( "Share", "disciple_tools" ); ?></span>
+                                <img class="dt-blue-icon" src="<?php echo esc_url( get_template_directory_uri() . "/dt-assets/images/share.svg" ); ?>">
+                                <span data-tooltip title="<?php echo esc_html( sprintf( '%s', ltrim( $shared_with_text, ',' ) ) ); ?>" style="margin:0 10px 2px 10px"><?php echo sprintf( 'Share (%s)', esc_html( count( $shared_with ) ) ); ?></span>
```
corsacca, line 122: Hey @cwuensche, looks like "share" might no longer be translated

I'd suggest
`<?php echo sprintf( _x( 'Share (%s)', 'Share (23)', 'disciple_tools' )  , esc_html( count( $shared_with ) ) ); ?>`
or 
`<?php esc_html_e( "Share", "disciple_tools" ); ?> (<?php echo esc_html( count($shared_with) ); ?>)`
I think i might prefer the second one just because it is already translated and it is easier to translate.

corsacca (changes requested):

dt-assets/functions/details-bar.php
cwuensche: > Hey @cwuensche, looks like "share" might no longer be translated
> 
> I'd suggest
> `<?php echo sprintf( _x( 'Share (%s)', 'Share (23)', 'disciple_tools' ) , esc_html( count( $shared_with ) ) ); ?>`
> or
> `<?php esc_html_e( "Share", "disciple_tools" ); ?> (<?php echo esc_html( count($shared_with) ); ?>)`
> I think i might prefer the second one just because it is already translated and it is easier to translate.

Went ahead with the second solution.

corsacca (approved):

PR #1266, opened 2021-03-30
Issue 708

cwuensche:


corsacca:
@cwuensche 
The cookie didn't save for me when i had more than 30 records.
We can put the code in the details.js instead of the contacts.js and the groups.js, this way it will work for any post type.

corsacca:
If we just save an array of the ids, we can generated the links to the prev and next by looking at the position in the array.

cwuensche:
I have almost finished this. I changed the code so IDs are saved instead of entire posts. I'm working on implementing code that will fetch posts based on the ID so I can populate the next/prev links.

corsacca:
@cwuensche awesome. This seems to work well.
Would you be able to but the prev and next links as buttons in the top right?
Something like:
image

cwuensche:
@corsacca I added a navigation arrow on the left and right side of the bar. Let me know what you think about the way I did it. I used the original image as an example, but we could change it to display like how you have it here, except on the right side of the navigation bar.

corsacca:
This is looking great @cwuensche!
We are liking the icon placement.
A couple changes:
- set the height to 20px 
- have a hover text with "previous record" and "next record"
- have the icons show up on mobile view too on each side on the other icons:
image

cwuensche:
@corsacca I haven't pushed the changes yet, but it seems like it would respond better if I had the arrows on both mobile and desktop view right next to each other like you have in the image on this page. While I was resizing the browser on desktop, the arrows started going to the next line. It would probably work better having the arrows side by side like you have on both desktop and mobile.

corsacca:
Hey @cwuensche ,
I'm still seeing the large icons:
image
Is that the same for you?

corsacca:
Thanks @cwuensche !

PR #1278, opened 2021-04-07
1137 - API: Option to get api errors sent by email.

kodinkat:
Main requirement goals:

- Know when/why a form is not working
- Know why someone can't sign in on the mobile app
- A user is having trouble doing something in the theme or in the app. Having an admin visible error log would help a lot in debugging.
- Maybe catch errors caused by a version conflict between a server config, plugins and the theme that we didn't test.


corsacca:
This is brilliant. Thank you @kodinkat !

PR #1279, opened 2021-04-09
Reorder record top bar to look better on small-ish screens < 1200px.

corsacca:
Move duplicates detected message.
Fix update needed toggle not being displayed.

Before:
Screenshot from 2021-04-09 11-53-43
After:
image


micahmills (changes requested):
There are a couple of RTL bugs that can be fixed pretty easily. I also noticed a RTL bug with the chevrons in RTL. line 454-455 of `_details.scss` should use the `margin-inline-start` and `margin-inline-end` properties instead of left and right.
Screen Shot 2021-04-12 at 8 57 13 AM

This update moves the Duplicate Detection warning to a lower priority especially on smaller screens.
Screen Shot 2021-04-12 at 8 29 02 AM
Screen Shot 2021-04-12 at 8 27 42 AM

This is probably ok since the duplicate detection can sometimes be a bit naggy and not always super accurate, but I want to make sure that we are conscience of the fact that we are potentially moving the duplicate warning "below the fold" for smaller screens.

corsacca:
Closing in favor of: https://github.com/DiscipleTools/disciple-tools-theme/pull/1288

PR #1288, opened 2021-04-14
Second bar fixes

micahmills:
Moves the Duplicate Detected notification to the details pane (from #1279).

rearranged the sub menu bar to make it more clear and responsively sized.


corsacca:
@micahmills thank you!

PR #1297, opened 2021-04-15
feat: click tag to create filter #863

### LINKED ISSUE #863: Tags: clickable tag to create a list/filter
Have a tag (that is in the tags field of the other tile of a contact) be clickable so that when clicked, it creates a list/filter of contacts or groups that have that tag.

squigglybob:
resolves #863


corsacca:
This looks brilliant.

squigglybob:
Hi @corsacca This PR is ready to go for review I think :)

corsacca:
Thank you @squigglybob 
To simplify I think we could remove the name part of the filter, and have it default to the 'Custom Filter' as you do here: https://github.com/DiscipleTools/disciple-tools-theme/pull/1297/files#diff-f4c9631e9bb5cbd0faec26098367fc3c1eb03a42b1df5e238c0fe29bd9f8a5a7R130
The label is what is displayed.

Will this setup handle a multiple field query?
ie: overall_status: 'active', seeker_path: 'met'
or tags: ['tag_1', 'tag_2' ]

squigglybob:
No worries @corsacca. Yes it will handle multiple field query, each filterLabel query param defines a new field query, and then they all get built together into a custom query.

Also just to clarify, do you want me to remove the optional queryparam `filterName` ?

corsacca:
@squigglybob Yeah. Please remove it, i'm looking at the list code and the name field does not look like it is being used.

corsacca:
`fieldQuery` is maybe better too.

squigglybob:
ok @corsacca those last query param name changes are done also

corsacca:
Thanks @squigglybob looks good.
I created a tag `this&that` and the list doesn't seem to filter to it. Is that the same for you?

squigglybob:
Yes, i'm getting the same thing.

Hmmm. It looks like the ampersand & is  [not being encoded by encodeUri](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI) it doesn't encode 

Not Escaped:

    A-Z a-z 0-9 ; , / ? : @ & = + $ - _ . ! ~ * ' ( ) #

It looks like `encodeUriComponent` should do the trick though...

PR #1301, opened 2021-04-20
#1063 missing clear search button firefox

### LINKED ISSUE #1063: Missing the clear search button in Firefox browser
The icon and functionality to clear the search box on the list pages of groups and contacts is available in Safari and Chrome but missing from Firefox.

Safari
Screen Shot 2020-09-12 at 2 49 11 pm

Chrome
Screen Shot 2020-09-12 at 2 48 53 pm

Firefox
Screen Shot 2020-09-12 at 2 49 27 pm

squigglybob:
resolves #1063


squigglybob:
hi @corsacca. This one is ready for review, I've checked it on Chrome and Firefox and is working well on those two.

corsacca:
Tested on mobile and desktop.
Looks good to me!
Thank you also for moving the inline css to the scss file.

jollypirate268:
@corsacca  I still have this issue.  FF 94.0.1 MacOS.

corsacca:
@jollypirate268 which D.T version are you using?
Can you post a screenshot here of what the field looks like for you?

jollypirate268:
@corsacca 

As I've looked at this again, it looks like this is not a Firefox issue -- it's either a different problem or maybe it's just me misunderstanding how it's supposed to work.

If you search by a term, you *do* get an x beside the search term.

However if you then view a contact, and then go back to the contacts page, the filter is still active, but there is no clear button (the x beside the form field) is no longer there.  The only indication that you have a filter active is now on the bottom left side.  And there no obvious (to me) way to clear it -- I just end up hitting the search button with an empty text field.  

I feel like the end user I am working with tripped over this as well.  What is the intended way for this to work?  I get that you might want to preserve the status of the filter -- but how are you *supposed* to clear the filter?  Now I realize that even if you restored the search term to the box -- it doesn't re-run the query if you clear it out.  

So, this is not a regression and the X appears to work the same in Chrome and firefox now (sorry for mis-attributing it to this issue) - but what is the expected way to clear out a filter?  Is it to make sure the search box is empty and hit search?

Here is my version info, since you asked, though now I think it doesn't matter.

WordPress version: 5.8.2 \| PHP version: 7.3.31-1~deb10u1 |  
-- | --
Server: nginx/1.14.2
D.T Theme Version: 1.17.0 |

corsacca:
@jollypirate268 I think i see what you mean.
When you have searched for a term, and come back to the contacts list, the search list is still filtered to the search.
To change the list you can click on any of the filters. 

An option might be to add an x to the boxes showing what the list is filtered to:
image

jollypirate268:
@c

> @jollypirate268 I think i see what you mean. When you have searched for a term, and come back to the contacts list, the search list is still filtered to the search. To change the list you can click on any of the filters.
> 
> An option might be to add an x to the boxes showing what the list is filtered to: image

Yes, that seems like a sensible solution.  Or a clear filters button to the right of all of these.

I had thought about restoring the search term to the box, but really that doesn't solve the problem as that does not help you clear it, it just provides another indicator that a filter is active.  

It could just be me, but It was not obvious to me that a filter was active (though now I see in your screenshot that it lists it there.  I am not sure what the best way to make this clear.  

Maybe even adding more text to the label, similar to the sort button there:
[**Active Filter:** "Bob Fred"]

corsacca:
Created this to capture the clear filter idea: https://github.com/DiscipleTools/disciple-tools-theme/issues/1527

PR #1302, opened 2021-04-21
Private fields

micahmills:
Ability to make a field private so that only the current user can see the inputted data.

Works for text, textarea, date, boolean, key_select, and multi_select fields.


micahmills:
I think everything is functioning now. I added a Privacy check box on the field creation screen in wp-admin. We have the ability to toggle a field between private and public and the data switches appropriately. 

I am working on getting phpunit running on my machine to get test written for this

dt-posts/dt-posts.php
```diff
@@ -179,6 +179,10 @@ public static function create_post( string $post_type, array $fields, bool $sile
                 $post_user_meta[$field_key] = $field_value;
                 unset( $fields[ $field_key ] );
             }
+            if ( $field_type === 'number' && $is_private ) {
+                $post_user_meta[$field_key] = $field_value;
+                unset( $fields[ $field_key ] );
+            }
```
corsacca, line 185: @micahmills would it be safe right here (and on contact update?) to add a catch all that would keep fields marked as private from being saved in the postmeta table?

dt-posts/dt-posts.php
```diff
@@ -445,15 +477,41 @@ public static function get_post( string $post_type, int $post_id, bool $use_cach
         $author = get_user_by( "ID", $wp_post->post_author );
         $fields["post_author_display_name"] = $author ? $author->display_name : "";
 
+        // phpcs:disable
```
corsacca, line 480: Why do we need to disable phpcs here?

corsacca: (and a couple lines below)

micahmills: I copied that from other places that use the `$wpdb->get_results()` or `$wpdb->prepare()` SQL statements. I don't think phpcs likes the SQL syntax. That being said the one on line 488 should be phpcs:enable not disable.

corsacca: In some areas we need to pass in an array (like array or IDs), the ->prepare does not deal well with those so we have to disable it's check.
Here we are using ->prepare and the values in the intended way, so we don't need the disable.

corsacca (approved):

PR #1303, opened 2021-04-21
feat: return extra meta data for group members in detailsSettings

### LINKED ISSUE #1112: Feature request: more information in group members list
In the same way that you use the feet symbol to show who is the leader in a group, could you show a symbol (like maybe a vertical cross) next to each baptised contact names who are group members?  
In this way, we would quickly see who is a baptised and who is not. Right now, we have to open a tab for each group member in order to get the info. This feature is a must for a team manager who may not necessarily know each name of the people who got baptized.

squigglybob:
resolves #1112


squigglybob:
@corsacca this one should be ready to review...

squigglybob:
@corsacca ok, _now_ it should be ready for review :smile:

dt-groups/groups.js
```diff
@@ -75,11 +75,20 @@ jQuery(document).ready(function($) {
       if( member.leader ){
         leaderHTML = `<i class="fi-foot small leader"></i>`
       }
+      const contactStatusHTML = ( member.data && member.data.overall_status )
```
corsacca, line 78: This section needs a bit of love from window.lodash.escape()

squigglybob: doh :man_facepalming:

squigglybob: resolved at 1c0dc67b

corsacca:
:) looks good!

PR #1314, opened 2021-04-23
feat: flexible time chart by field. #1305

### LINKED ISSUE #1305: Metrics: flexible time chart by field.
The goal is to take any field and drill down into the stats of that field.
The post_date field will show when and how many contacts were created each month.
The metrics for the Status field could tell us how many contact became active each month, or how many were active on a particular month.

User flow:
Choose a post type (Contacts, Groups)
Choose one of any field (ie: post_date, overall_status, quick_action_call, milestones)
If the field is a key_select or a multi_select: choose one of the field values.

Get data and show 2 possible charts:
- Totals over time
First month 200 total contacts, second month total 500, third total 1000, etc

- Activity over time
First month with added 200 contacts, second month we added 300 contact, third month we added 500 contacts, etc.

image

By default we will have a column for each month, but have the option to show by year as well.

---
Dev notes:
Some logic can be taken from the critical path charts.
For users with the `view_project_metrics` permission
Put this chart menu under the "project" tab

squigglybob:
resolves #1305

### DIFF
--- a/dt-metrics/combined/critical-path.php
+++ b/dt-metrics/combined/critical-path.php
@@ -34,17 +34,6 @@ class DT_Metrics_Critical_Path_Chart extends DT_Metrics_Chart_Base
         add_action( 'rest_api_init', [ $this, 'add_api_routes' ] );
     }
 
-    public function has_permission(){
-        $permissions = $this->permissions;
-        $pass = count( $permissions ) === 0;
-        foreach ( $this->permissions as $permission ){
-            if ( current_user_can( $permission ) ){
-                $pass = true;
-            }
-        }
-        return $pass;
-    }
-
     public function add_url( $template_for_url ) {
         $template_for_url['metrics/critical-path'] = 'template-metrics.php';
         return $template_for_url;


dt-metrics/combined/time-charts.php
```diff
@@ -0,0 +1,113 @@
...
+
+        $url_path = dt_get_url_path();
+        if ( "metrics/$this->base_slug/$this->slug" === $url_path ) {
+
+            add_action( 'wp_enqueue_scripts', [ $this, 'scripts' ], 99 );
+
+        }
+
+        // $post_types = DT_Posts::get_post_types(); // This is only giving [ 'peoplegroups' ] at the moment
+        $post_types = [ 'contacts', 'groups', 'peoplegroups' ];
```
squigglybob, line 40: @corsacca this was the bit of code put in to enable me to continue, but the line above was not giving the full list of post types :(

corsacca: @squigglybob no problem, i'll tackle that one.

squigglybob:
@corsacca Currently the field types that are graphable are 'date', 'tags', 'multi_select', and 'key_select'.  Let me know how it goes with real data, and we can make any necessary adjustments.

squigglybob:
@corsacca This one now has all of the field types that have been requested. It just needs checking on some larger data.

One thing I wasn't sure about was whether there could be more than one row for a p2p_id in the wp_p2pmeta table. If it is possible then I would need to add in a DISTINCT function call before getting the count in the query (this is for the connection field type)

corsacca:
image

Since  `p2p_id` primary it will be unique, so we should be good.

dt-assets/js/shared-functions.js
```diff
@@ -186,6 +186,15 @@ window.API = {
...
 
+  getTimeMetricsByYear: (postType, field, year) =>
+    makeRequest('GET', `metrics/time_metrics_by_year/${postType}/${field}`),
+
+  getTimeMetricsByMonth: (postType, field, year) =>
+    makeRequest('GET', `metrics/time_metrics_by_month/${postType}/${field}/${year}`),
+
+  getFieldSettings: (postType) =>
+    makeRequest('GET', `metrics/field_settings/${postType}`),
+
```
corsacca, line 197: These are specific to the metrics section. So let's keep them in the time-charts.js

dt-metrics/combined/time-charts.js
```diff
@@ -0,0 +1,630 @@
+jQuery(function() {
```
corsacca, line 1: This file needs some window.lodash.escap()

dt-metrics/combined/critical-path.js
```diff
@@ -37,7 +37,7 @@ function project_critical_path() {
     <div id="ongoingChart" style="width:90%;"></div>
     <!--<div id="chartdiv" style="height: 800px; width:100%;"></div>-->
     <br>
-    <h4>${ translations.filter_critical_path }</h4>
+    <h4>${ window.lodash.escape( translations.filter_critical_path ) }</h4>
```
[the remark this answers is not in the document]
squigglybob: no problem :slightly_smiling_face:

dt-assets/js/shared-functions.js
squigglybob: resolved 7afd3930

dt-metrics/combined/time-charts.js
squigglybob: I think I got them all. I made sure to run them through my `escapeObject` function, which does all the escaping in one go e.g. lines 19-28

squigglybob:
@corsacca ok, i think that's all of those changes done

dt-assets/js/shared-functions.js
corsacca: what i meant for here is to not include the functions for metric enpoints in the shared-function.js since the endpoints might not be commonly available
We could either move these code declarations to time-charts.js, or call the makeRequest function from the time-charts.js

squigglybob: ah :slightly_smiling_face:  gotcha! sure thing

squigglybob: resolved 58b10bd5

squigglybob:
@corsacca I've just made a change to discount any timestamps with a leading 0, so that should get rid of the 0 timestamp that was giving you an earliest date of 1970... let me know if that does it.

corsacca (approved):

PR #1318, opened 2021-04-24
User apps project

ChrisChasm:
- Expands the ability to add a filter against the post_user_meta table
- Adds a section to hook in the settings that allows users to enable user specific apps to their environment.
- These edits are preparation for allowing user driven mico apps to the system.


ChrisChasm:
Okay, user_id isolation is added to this PR from conversation with @corsacca . The pull can wait until the work @micahmills is doing with the private key select. Once that core element is added, then I can evaluate and either implement with that method, or make a case for this other strategy as non-redundant.

corsacca (approved):

PR #1321, opened 2021-04-27
Tags field type

### LINKED ISSUE #1269: Field Type: Tags
Make tags it's own field type to go along side multi_select, key_select, etc
It currently rides off of the multi_select field type.
the multi_select is design to be a finite list of items. 
The tags field is meant to have many options and lets the user add options.

This will let a record have multiple different tag fields. 

Bonus points for also implementing this issue at the same time: #567 
Normal tags are stored in the `postmeta` table.
Private tags would be stored in the `dt_post_user_meta`

cairocoder01:
Resolves #1269 

Adds `tags` field type and converts the existing tags field to use it.


corsacca:
Great work @cairocoder01 !! :) This all looks good.
I pushed 2 small fixes as i tested it out.

One spot i forgot to mention to look for is: the create new record page. The tag fields now show up there but we have no js to run the search.

cairocoder01:
@corsacca Does your comment mean that you have fixed the new record page, or do I need to check and confirm that since you said there is no js to run the search?

corsacca:
> @corsacca Does your comment mean that you have fixed the new record page, or do I need to check and confirm that since you said there is no js to run the search?

Tags on the new record page still need to be implemented

cairocoder01:
@corsacca This should be ready now

dt-core/global-functions.php
```diff
@@ -422,11 +422,35 @@ function render_field_for_display( $field_key, $fields, $post, $show_extra_contr
                         </option>
                     <?php endforeach; ?>
                 </select>
+            <?php elseif ( $field_type === "tags" ) : ?>
+                <div class="tags">
+                    <var id="<?php echo esc_html( $display_field_id ); ?>-result-container" class="result-container"></var>
+                    <div id="tags_t" name="form-tags" class="scrollable-typeahead typeahead-margin-when-active">
```
corsacca, line 428: I think id needs to be the $display_field_id too, so we don't have multiple fields with the same id

corsacca: Ha. That was fast :)

PR #1325, opened 2021-04-28
1295 - Error Log Table New User Column

kodinkat:
## Implementation Summary

- New user column added to Utilities (DT) Error Log table.
- User's display name is shown if a valid user id is detected.
- Empty/Blanked string for all other states.


corsacca:
Thanks @kodinkat ! :)

PR #1333, opened 2021-04-30
set generic icon in mobile view

### LINKED ISSUE #1329: Icon for `Create new ...` button on mobile view to be generic
When in normal view on a browser the icon for the green button to `create new contact` or `create new group` or `create new training` is a generic plus icon. But in a reduced screen, the icon uses an image for add contact - and that is the same icon for a group and a training - which is confusing.

Screen Shot 2021-04-29 at 09 26 16
Screen Shot 2021-04-29 at 09 26 43

So, replace the add contact icon to simply use the generic plus icon for all instances - whether normal or mobile view.

mikeallbutt:
closes #1329 by adding the white generic icon

Screen Shot 2021-04-29 at 22 27 22


corsacca:
Thanks @mikeallbutt

PR #1337, opened 2021-04-30
Map module bug

prykon:
GUI gets updated upon editing a Geo

### DIFF
--- a/dt-mapping/drill-down.js
+++ b/dt-mapping/drill-down.js
@@ -24,7 +24,7 @@ let rebuild_drill_down = ( response, bindFunction, grid_id )=>{
 
             // create button
             html += `<li><button id="${window.lodash.escape( section.parent )}" type="button" ${disabled ? "disabled" : ""}
-                onclick="DRILLDOWN.get_drill_down( '${window.lodash.escape( bindFunction )}', '${window.lodash.escape( section.selected )}' )"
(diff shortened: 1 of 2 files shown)


dt-mapping/drill-down.js
```diff
@@ -24,7 +24,7 @@ let rebuild_drill_down = ( response, bindFunction, grid_id )=>{
 
             // create button
             html += `<li><button id="${window.lodash.escape( section.parent )}" type="button" ${disabled ? "disabled" : ""}
-                onclick="DRILLDOWN.get_drill_down( '${window.lodash.escape( bindFunction )}', '${window.lodash.escape( section.selected )}' )"
+                onclick="DRILLDOWN.get_drill_down( '${window.lodash.escape( bindFunction )}', '${window.lodash.escape( section.selected )}', false )"
```
corsacca, line 27: @prykon Is this one called from the admin?

prykon: Yes; it's the breadcrumb that shows the selected geo buttons.

corsacca: I would expect that enabling caching here would produce the opposite effect, show old names

prykon: I double checked it and the breadcrumb updates on edit if this line is set to true or false. Your call :)

corsacca: The metrics use this function too, so if the impact is the same i'd prefer not changing the functionality

corsacca:
Thanks Dario!

PR #1340, opened 2021-05-05
fix: location deletion not showing up in activity log

### LINKED ISSUE #1309: Deleting a location is not getting logged
If a location is set in the `Location or Address` field in the `Details tile` and the location is removed (by clicking on the red X on the right side of the field), 
Screen Shot 2021-04-22 at 22 34 11
then the information is correctly removed and the field becomes empty. When this type of thing is done with other fields (ie. data being removed or updated - like a phone number) then that activity is logged and appears on the right site in the `Comments and Activity` tile.

When a location is set, then it does get logged and appears in the `Comments and Activity` tile. The same should happen when a location is removed.

squigglybob:
resolves #1309

### DIFF
--- a/dt-core/logging/hooks/class-hook-posts.php
+++ b/dt-core/logging/hooks/class-hook-posts.php
@@ -206,7 +206,11 @@ class Disciple_Tools_Hook_Posts extends Disciple_Tools_Hook_Base {
 
         if ( ! empty( $fields ) && ! $object_note ) { // Build object note if contact, group, location, else ignore object note
             if ( $new ) {
-                $object_note = 'Added ' . $this->_key_name( $meta_key, $fields ) . ': ' . $this->_value_name( $meta_key, $meta_value, $fields );
+                if ( $meta_key === 'location_grid_meta' ) {
+                    $object_note = Disciple_Tools_Mapping_Queries::get_location_grid_meta_label( (int) $meta_value );
+                } else {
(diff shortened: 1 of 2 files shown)


squigglybob:
@corsacca I think that fixes it

corsacca:
In testing i see this fixes the issue for all locations added from now on.
In deleting the a location created before this PR, the delete message will look something like:
image

corsacca:
I don't know if there is anything we can do since the data was not there originally

corsacca:
Unless we write a special check that tells if it is the old format and just displays "location removed"

squigglybob:
that would be possible I'm sure

squigglybob:
@corsacca something like this? although it adds two new translation strings.

The 'location removed' string shouldn't be hit by anybody though, as prior to this PR there hasn't been an activity log for removal of `location_meta`

corsacca:
@squigglybob Thanks coding a solution.
If the new way adds extra translations and will rarely be hit, then lets just for for the solution you had initially

squigglybob:
@corsacca only one extra translation string "location added" will still be needed for any activity logs for the addition of a location where that location was subsequently removed.

Previously in that case it just showed the truncated string " added to locations" (from '%1$s added to locations" translation but the new translation string would change that to "location added".

We could just leave the older logs as they are as "added to locations" still says the same thing as "location added".

Let me know what you think.

PR #1343, opened 2021-05-05
Map module delete locations

prykon:
Custom locations and sublocations can now be deleted from the admin mapping location list.

- Only custom locations get 'delete' button (server side validation is also in place... of course)
- Location list gets updated on delete


corsacca:
@prykon can resolve the conflict?

prykon:
Hey Jon. I saw that jQuery call to the element with id 'delete-sub-location-button' yet it never gets called... ever... it the whole DT theme. So I removed it. If you want I can add it back in, but I think we should remove it. Please let me know how to proceed. Thanks!

corsacca:
@corsacca what are you referring to?

prykon:
@corsacca I'm referring to the code snippet that causes the conflict in the PR...

`jQuery('#delete-sub-location-button').on('click', function () {
                    jQuery('#delete-location-spinner').show()
                })`

corsacca:
@prykon this is code that you added in your previous PR. If you removed the target, then yes please remove the code.

prykon:
Done.

dt-mapping/drill-down.js
```diff
@@ -24,7 +24,7 @@ let rebuild_drill_down = ( response, bindFunction, grid_id )=>{
 
             // create button
             html += `<li><button id="${window.lodash.escape( section.parent )}" type="button" ${disabled ? "disabled" : ""}
-                onclick="DRILLDOWN.get_drill_down( '${window.lodash.escape( bindFunction )}', '${window.lodash.escape( section.selected )}' )"
+                onclick="DRILLDOWN.get_drill_down( '${window.lodash.escape( bindFunction )}', '${window.lodash.escape( section.selected )}', false )"
```
corsacca, line 27: @prykon this looks like the same area as our previous conversation.

dt-mapping/mapping.php
```diff
@@ -436,6 +442,31 @@ public function modify_location_endpoint( WP_REST_Request $request ) {
             return DT_Mapping_Module_Admin::instance()->process_rest_edits( $params );
         }
 
+        public function delete_sublocation( WP_REST_Request $request ) {
```
corsacca, line 445: Missing permission check

corsacca (changes requested):

dt-mapping/drill-down.js
prykon: @corsacca
We should leave the 'false' cache parameter... explanation below

**Behavior with the 'false' cache parameter:**
1. Enter a country
2. Change its name
3. Click the 'World' button in the breadcrumb
4. You're back in the main locations list with the updated country name

**Behavior without the 'false' cache parameter:**
1. Enter a country
2. Change its name
3. Click the 'World' button in the breadcrumb
4. You're back in the main locations list with not-yet updated country name

dt-mapping/mapping.php
corsacca: @prykon thanks for this change!
In general, let me mark the conversations as resolved

dt-mapping/drill-down.js
corsacca: Since this code is used in other places in the theme (metrics), how about invalidating the cache when a record is updated or deleted instead of completely disabling it?
could be as simple as doing: `window.drilldownModule.drilldown = null`

corsacca (approved):

PR #1344, opened 2021-05-06
Allow the user to change which record types fields show up on.

corsacca:
Move settings (DT) > New Record Page tab to the individual fields.
Make user_select field a core field


corsacca:
Current example with the age field:
image

PR #1347, opened 2021-05-07
fix: some date's not showing in user's selected language

### LINKED ISSUE #1252: Date using instance language and not user language
In the area where a contact record displays when the contact was created, the date information is displayed in the language of the instance (like Spanish) and not of the language preferred by the user (English)

squigglybob:
resolves #1252


squigglybob:
hey @micahmills yes that's what I'd worked out too after a lot of digging :)
I tried hooking into the `gettext` hook to change the translation to be done using the disciple_tools namespace, but the date didn't come through there which was surprising!

All of the dates that are translating correctly look like they are either using moment on the frontend or using `window.SHAREDFUNCTIONS.formatDate` neither of which are respecting the WP date format option.

I'll try and make the details bar date done by the frontend, which should be fairly straight forward, however the dates in the user list might be a bit trickier as the table is created purely in php :slightly_frowning_face: 

@corsacca might want to weigh in too I guess

micahmills:
It was awhile ago but I think we ended up using JS to format dates in the date fields as well because it allowed us to do it in the appropriate date format for the users locale.

corsacca:
Thanks @squigglybob !

PR #1348, opened 2021-05-07
1312 change and fix list filters date options

### LINKED ISSUE #1312: List Filters: written dates and old dates.
1. The date selector has a limit to how far back you can select a year (10 years ago).
image

We can increase this to 20 years.

2. When the user updates the text of the date in the input, this change is not saved and so does not reflect in the filtered records.

squigglybob:
resolves #1312

### DIFF
--- a/dt-assets/js/modular-list.js
+++ b/dt-assets/js/modular-list.js
@@ -985,6 +985,11 @@
   $("#filter-modal input.dt_date_picker").on('blur', function (e) {
     // delay the blur so that if the user has clicked we get the correct date from the input
     setTimeout(() => {
+      if (!e.target.value) {
+        const clearButton = $(this).prev('.clear-date-picker')
+        clearButton.click()
+        return
+      }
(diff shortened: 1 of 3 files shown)


squigglybob:
@corsacca I may have gone overboard with allowing dates to go back 20 years, as I didn't realise it was just for the filters :slightly_smiling_face:

squigglybob:
only to the ones that didn't already have a `dateRange` set yes

squigglybob:
@corsacca that should be done I think. Now if the user clicks/tabs away from the datepicker after manually entering in the date, then the onSelect will fire and update things as if they had selected the date using the dropdown

corsacca:
Thanks @squigglybob,
If the field is empty and you click into it and then out it automatically puts in a date.
I think we can keep this from happening by checking if the val input val is empty

corsacca:
Slick! thanks @squigglybob

PR #1353, opened 2021-05-10
Feature favorite posts

micahmills:
Adds the ability to favorite or star any post type


dt-assets/functions/details-bar.php
```diff
@@ -116,6 +116,11 @@ class="second-bar" id="second-bar-large">
                         </div>
                     </div>
                     <div class="cell small-5 large-4 align-right grid-x">
+                        <div class="cell shrink center-items">
```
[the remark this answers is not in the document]
corsacca: I get the ones in the modular-list.js being inline.

dt-contacts/base-setup.php
```diff
@@ -521,6 +529,18 @@ public static function dt_user_list_filters( $filters, $post_type ){
                     [ "id" => 'recent', 'name' => __( "Last 30 viewed", 'disciple_tools' ) ]
                 ]
             ];
+            $filters["filters"][] = [
```
corsacca, line 524: I think this filter should be available on all post types by default.
Maybe: `dt-posts/custom-post-type.php`

corsacca (changes requested):
This is looking good :)

corsacca:
@micahmills I move the field declaration to `dt-posts/custom-post-type.php` so it is available on all post types.

dt-contacts/base-setup.php
micahmills: I tried to add this in the `dt-posts/custom-post-type.php` but because the tabs have different names between post types it is hard to do it in a generic way. I have included it in groups as well as contacts, but right now we are doing filters for each post type separately.

dt-assets/js/modular-list.js
```diff
@@ -477,16 +477,21 @@
...
               })
             } else if ( field_settings.type === "boolean" ){
-              values = ['&check;']
+              if (field_key === "favorite") {
+                values = [`<svg class='icon-star selected' viewBox="0 0 32 32" data-id=${record.ID}><use xlink:href="${window.wpApiShare.template_dir}/dt-assets/images/star.svg#star"></use></svg>`]
+              } else {
+                values = ['&check;']
+              }
             }
+          } else if ( !field_value && field_settings.type === "boolean" && field_settings.name === "Favorite") {
```
corsacca, line 486: @micahmills  Here too ;)

dt-contacts/base-setup.php
corsacca: I'll take a look at this one. I like not having to copy paste the same code for each post type.
We need something similar for the "recently viewed" filter too.

corsacca (approved):

corsacca:
Thank you @micahmills  :)

PR #1355, opened 2021-05-10
fix: get correct sort label when doing a reverse sort #1349

### LINKED ISSUE #1349: `Sorting by` displays key and not label
When a user sorts the lists (on either a contacts or records or trainings etc.) then when the sort is in reserve order, the `key` name for the field is displayed and not the `key label`.

In the attached images, not the blue arrows and which way they are pointing.

**Wrong**
sort--sub

**Correct**
sort-sub

squigglybob:
resolves #1349

### DIFF
--- a/dt-assets/js/modular-list.js
+++ b/dt-assets/js/modular-list.js
@@ -335,7 +335,11 @@
       } else if (  sortLabel.includes('post_date') ) {
         sortLabel = list_settings.translations.creation_date
       } else  {
-        sortLabel = window.lodash.get( list_settings, `post_type_settings.fields[${filter.query.sort}].name`, sortLabel)
+
+        // remove leading dash from sort filter key when reverse sorting
+        const leadingDashSearch = new RegExp('^-')
+        const querySortKey = (sortLabel.search(leadingDashSearch) > -1) ? sortLabel.replace(leadingDashSearch, '') : sortLabel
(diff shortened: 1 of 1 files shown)


squigglybob:
@corsacca I think that's done

prykon:
Hi @squigglybob ! Found a minor thing... if you reverse sort, the sort label shows the sortLabel and not the pretty name.

How to reproduce:
1. Got to /contacts
2. Sort by contact name; label displays: `Sorting By: Name` (this is correct)
3. Click again on the `Name` column to reverse sort it; label now displays: `Sorting By: -name`

This happens in all the columns. For instance, if you reverse sort the `Contact Status` column, the label displays: `Sorting By: -overall_status`

I think having the label display `Sorting by: Contact Status (desc.)` could be a proper solution.

squigglybob:
Hi @prykon Thanks for your comment. What you have descibed is the issue that this PR fixes. We'll have to let @corsacca decide about the wording of the label.

Looking at how google does it in their tables they have the following:
image 
image

Which is not dissimilar to the one in D.T.
image
image

So I guess the question is whether the label should also show this as you have mentioned.

prykon:
Hi @squigglybob ! I removed my comment upon better understanding what this issue was about. Everything's looking good now! :)

PR #1356, opened 2021-05-10
Update UI to have the same logic as Custom Tiles tab in admin menu

prykon:


corsacca:
@prykon keep this section visible, so that after selecting a field the user can select another field without going "back":
image

PR #1359, opened 2021-05-11
feat: add ability to search for 'all' posts with a particular connection field

### LINKED ISSUE #1285: List API: Ability to search for 'any' on a connection field.
Search for any contact that has a baptism connections
Get a list for any contact that is coaching.
Get a list of any group that is a "parent group"

squigglybob:
resolves #1285

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,9 +2,9 @@
     "require-dev": {
         "dealerdirect/phpcodesniffer-composer-installer": "*",
         "wp-coding-standards/wpcs": "*",
-        "squizlabs/php_codesniffer": "*",
-        "phpunit/phpunit": "7.4.*"
+        "squizlabs/php_codesniffer": "*"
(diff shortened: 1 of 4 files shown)


squigglybob:
@corsacca I think it should be as simple as this... :crossed_fingers: 

This should allow the front end to use 'all' in the e.g. 'baptized_by' array to get all posts that have a connection in the 'baptized_by' field
image

I've given it a whirl with Postman and it seems to be working well :slightly_smiling_face:

squigglybob:
@corsacca i've added a unittest and also a dev depencdency of the correct version of phpunit if that's ok? It should mean that people have the right version of phpunit to run the tests (I had a different version installed globally on my machine) from the installation folder.

If that's not ok, I can easily revert those changes.

otherwise, I think that the API side is done.

PR #1360, opened 2021-05-11
feat: list filters - add in ability to search for all on connection fields

### LINKED ISSUE #1286: List Filters: ability to search for 'any' on connection field
Search for any contact that has a baptism connections
Get a list for any contact that is coaching.
Get a list of any group that is a "parent group"

We could add a check box/option under the typeahead in the filter list to simple search for any connection.

This feature depends on the list API supporting it: https://github.com/DiscipleTools/disciple-tools-theme/issues/1285

squigglybob:
resolves #1286


squigglybob:
@corsacca This just needs a check over and then I think should be ready to go.  The API side matches this side in the sense that they are both using '*' as the all key.

It'd probably be good to get the API one merged into here to test that the two work correctly together before finalising this PR though.

PR #1361, opened 2021-05-11
Custom fields UI update - navigation fix

prykon:


corsacca:
Thanks for the fix @prykon !

PR #1362, opened 2021-05-12
disables private checkbox in wp-admin for default fields.

micahmills:


corsacca:
Nice solution! Works perfectly, except for name field for some reason:
image

micahmills:
That is strange, for some reason `name` is not a key in the defaults array.

micahmills:
So now all fields that are created as a default or through a plugin will not be allowed to be made private.

corsacca:
that should fix it

PR #1364, opened 2021-05-12
Author update

prykon:
Unify all author names except for those in third party files

### DIFF
--- a/dt-core/admin/multi-role/class-user-edit.php
+++ b/dt-core/admin/multi-role/class-user-edit.php
@@ -4,7 +4,7 @@
  *
  * @package    Members
  * @subpackage Admin
- * @author     Justin Tadlock <justin@justintadlock.com>
+ * @author     Disciple.Tools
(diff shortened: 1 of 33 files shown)


dt-core/admin/multi-role/class-user-edit.php
```diff
@@ -4,7 +4,7 @@
  *
  * @package    Members
  * @subpackage Admin
- * @author     Justin Tadlock <justin@justintadlock.com>
+ * @author     Disciple.Tools
```
corsacca, line 7: This should stay as the original i think @prykon

dt-core/global-functions.php
```diff
@@ -67,7 +67,7 @@ function dt_write_log( $log ) {
          *          Also supports WP installations in subfolders
          *
          * @returns boolean
-         * @author matzeeable
+         * @author Disciple.Tools
```
corsacca, line 70: maybe just remove this line

prykon:
PR resent through different branch in order to remove unnecessary commits.

PR #1368, opened 2021-05-14
Cleans up Second bar

### LINKED ISSUE #1366: `created on` text is different size
As seen in the attached image, the text of the `created on "date"` is a different size other text in the phrase.

text-size-differences-created-on

Have the text be the same size as the other text in the phrase.

micahmills:
We have a lot of inline styles in the header, some makes sense others do not. This fixes #1366 and cleans up the styling and puts it into the CSS for easier maintenance later.


corsacca:
This looks good. Thanks @micahmills !

PR #1369, opened 2021-05-14
feat: basic feature tour using Shepherd for desktop #748

### LINKED ISSUE #748: Create a Feature Tour for when a user first logs in to D.T
Library ideas here: 
https://stackoverflow.com/questions/23363529/client-side-feature-tour-tutorial-instructional-overlay-system

squigglybob:
resolves #748 

First thought about using intro.js but this didn't work well with D.T. straight out of the box.

Second choice was [Shepherd.js](https://shepherdjs.dev/docs) which worked well out of the box.


squigglybob:
hi @corsacca this one should be ready for review, I guess particularly to check that it's the kind of thing you're after.

I've tried to make it so that you could create tours for other pages too. Currently untested, but should work in theory.

dt-assets/functions/enqueue-scripts.php
```diff
@@ -265,7 +274,7 @@ function dt_site_scripts() {
             $dependencies[] = 'mapping-drill-down';
         }
 
-        dt_theme_enqueue_script( 'dt-settings', 'dt-assets/js/settings.js', $dependencies, true );
+        dt_theme_enqueue_script( 'dt-settings', 'dt-assets/js/bb', $dependencies, true );
```
corsacca, line 277: this looks accidental

squigglybob: oops, yes, i must have hit my keyboard at the wrong moment and not noticed 🤦🏻‍♂️

squigglybob: resolved with 1a291dd

squigglybob:
@corsacca the feature tour has been pulled out of this PR into a plugin [disciple-tools-feature-tour](https://github.com/squigglybob/disciple-tools-feature-tour)

I've left in some refactoring of `escapeObject` into Shared functions, some handy id/classes on the elements that I was attaching the tour guides to (for simplified element finding), and also removed the `disable_feature_tour` REST api end point as this has moved to the feature tour plugin.

PR #1371, opened 2021-05-17
feat: email batch notifications #724

### LINKED ISSUE #724: Notifications: Batch send notification emails.
**Purpose**
Reduce email clutter and increase clarity on activity related to a record.

**use case**
- As an admin user, I get a lot of email notifications for updates on groups - often multiple notifications for the same group/contact per day. I don't want to turn them off and miss what's happening, but I would like to have them batched together to come less often than in real-time.

**Requirements**
- User preference that can specify the frequency with which someone would like to receive email notifications. Options: real-time, hourly, or daily. 
- Email queue for check for emails to send
- Group notifications for a user on a record together
- Cron (wp scheduler) to check for emails to send every hour.

**Code considerations**
Start with `dt-notifications/notifications.php` and `dt-notifications/notifications-email.php`. The email queue is a new system that will need to be implemented. 

Note: Making sure scheduled emails are sent on time will be greatly improved by implementing a real cron service: 
see https://github.com/DiscipleTools/disciple-tools-theme/issues/888 and https://developers.disciple.tools/hosting/cron

squigglybob:
resolves #724


squigglybob:
ok @corsacca I think this is one is pretty close (as long as the build checks fine).

hour:minute timestamps are in the email now, but I also realised that the time will be coming out in UTC, so put a `(UTC)` string in the subject to make it a bit clearer what the timezone of the times is.

I guess this should be fixable to display in the user's timezone but would be more complicated and wanted to get the PR reviewing and some feedback a-coming.

squigglybob:
@corsacca this fix should fix the multisite issue. I changed the update code in `users-endpoints.php` as I had previously just used the `update_user_option` function which was instance specific on multisites

squigglybob:
@corsacca ok. removed times from email subject and body. And also condensed the spacing for comments and mention notifications.

corsacca:
Thanks @squigglybob !

PR #1402, opened 2021-05-27
fix: iron out warnings/errors/deprecations from PHP 8.0 upgrade #1384

### LINKED ISSUE #1384: Make sure D.T runs with php8
With `WP_BEBUG` on, errors are being thrown on the list page and probably more places.

squigglybob:
resolves #1384

### DIFF
--- a/archive-template.php
+++ b/archive-template.php
@@ -459,7 +459,7 @@ dt_please_log_in();
                                             if ( isset( $b["show_in_table"] ) ){
                                                 $b_order = is_numeric( $b["show_in_table"] ) ? $b["show_in_table"] : 90;
                                             }
-                                            return $a_order > $b_order;
+                                            return $a_order <=> $b_order;
                                         });
                                         foreach ( $post_settings["fields"] as $field_key => $field_value ){
                                             if ( ( isset( $field_value["show_in_table"] ) && $field_value["show_in_table"] ) ){
--- a/dt-contacts/access-module.php
+++ b/dt-contacts/access-module.php
@@ -348,7 +348,7 @@ class DT_Contacts_Access extends DT_Module_Base {
 
             //order overall status options
             uksort( $declared_fields["overall_status"]["default"], function ( $a, $b ) use ( $fields ){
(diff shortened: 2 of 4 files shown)


squigglybob:
A couple of deprecation warnings for a vendor package of posts-to-posts are fixed in a PR on the (scribu repo)[https://github.com/scribu/wp-scb-framework/pull/60]

squigglybob:
Ok @corsacca , here's where I am currently. I have currently made code changes to remove WP_DEBUG warnings/errors.
I also wanted to make sure that the phpunit tests would run under php8.0

Here's what I've hit so far

## Environment:

Using docker-compose with `mysql:5.7` image, and `wordpress:php8.0` image in which i'm globally installing phpunit so as not to include any phpunit dependencies in our `composer.json`

Installing any version of phpunit below version 8.0 results in the error 

```
Package phpunit/phpunit at version ^6.0 has a PHP requirement incompatible with your PHP version, PHP extensions and Composer version:  
    - phpunit/phpunit 6.5.14 requires php ^7.0 which does not match your installed version 8.0.6.
```

However using `phpunit ^8.0` results in the wp error

```
root@76c68c1e686a:/var/www/html/wp-content/themes/disciple-tools-theme# ~/.composer/vendor/bin/phpunit 
Error: Looks like you're using PHPUnit 8.5.19. WordPress requires at least PHPUnit 5.4 and is currently only compatible with PHPUnit up to 7.x.
Please use the latest PHPUnit version from the 7.x branch.
```

I have also tried installing `phpunit ^6.0` outside the docker container (so in my own php 7.x local environment so there are no complaints) in the vendor folder within the theme and then running it within the docker container (php 8.0 env) and the phpunit test runs fine, but i'm not sure what downsides there would be as the lower versons of phpunit are not technically compatible with php8.0

Any thoughts would be greatly appreciated :)

corsacca:
@squigglybob 
Thank for this research!
Let's fix the errors in the theme display, and wait for phpunit upgrades when it is supported.

corsacca:
@squigglybob would this one be ready then?

squigglybob:
@corsacca yes i think so

corsacca:
thanks @squigglybob !

PR #1403, opened 2021-05-28
1374 - Test Geocoding Keys in Admin Area

kodinkat:
## Implementation Summary

- Mapbox:
  - Refactored `is_active_mapbox_key()` logic to also include/display additional error messages.
- Google:
  - Implemented is_active_google_key() logic to display additional error messages and check across the following api types:
    - Address based search.
    - Latitude and longitude coordinates based search.
    - Address based search with components. 

Screenshot 2021-05-28 at 14 14 15

Screenshot 2021-05-28 at 14 42 42

Screenshot 2021-05-28 at 14 46 09

### DIFF
--- a/dt-mapping/geocode-api/google-api.php
+++ b/dt-mapping/geocode-api/google-api.php
@@ -197,8 +197,15 @@ if ( ! class_exists( 'Disciple_Tools_Google_Geocode_API' ) ) {
                 $status_class = 'connected';
                 $message      = 'Successfully connected to Google Maps API.';
             } else {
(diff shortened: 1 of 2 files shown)


corsacca:
Thanks @kodinkat 
Let's not display an error if no google key is added, as that is an ok state.

kodinkat:
> Thanks @kodinkat
> Let's not display an error if no google key is added, as that is an ok state.

Screenshot 2021-06-02 at 10 29 34

corsacca:
Thank you @kodinkat !

Issue #1365, opened 2021-05-13
Delete user doesn't remove relevant user meta from database


prykon:
WP Admin screen user delete doesn't remove all relevant user meta

**To Reproduce**
1. Go to any contact that doesn't have a user created
2. Click on 'Make a user from this contact' from the 'Admin actions' dropdown
3. Complete the form and create the user
4. Go to the WP Admin and delete the user
5. Repeat step 2 for the same user.

**Expected behavior**
It should allow you to create a user from the contact. Instead, it says there's already a user created from the contact.

PR #1404, opened 2021-05-31
User delete workflow fixes for #1365

micahmills:
adds a function on the 'delete_user' hook to delete the 'correspons_t…o_user' meta key and value.


functions.php
```diff
@@ -485,3 +485,16 @@ function dt_theme_admin_notice_required_php_version() {
     </div>
     <?php
 }
+
+add_action( 'delete_user', 'dt_delete_user_contact_meta' );
+function dt_delete_user_contact_meta( $user_id ) {
+    global $wpdb;
+    // phpcs:disable
+    // WordPress.WP.PreparedSQL.NotPrepared
+    $wpdb->get_results( "DELETE FROM $wpdb->postmeta pm
```
corsacca, line 494: @micahmills this one should work fine using $wpdb->prepare() so with don't have to disable phpcs

corsacca:
Thanks @micahmills !
We can put this function in dt-users/users.php

corsacca:
If a user is deleted from the network side, this might need to run on each subsite.

micahmills:
According to the Wordpress codex (https://developer.wordpress.org/reference/hooks/delete_user/) doesn't run on all multisite situations when a user is deleted, but if we change to the 'deleted_user' hook it should run. I will play with it and see what it does.

micahmills:
Something to note here is that the record type continues to be a 'user.' We could revert it to another user type in this function, but there are possible security concerns with changing it to an access record when it should be a private record.

corsacca:
> According to the Wordpress codex (https://developer.wordpress.org/reference/hooks/delete_user/) doesn't run on all multisite situations when a user is deleted, but if we change to the 'deleted_user' hook it should run. I will play with it and see what it does.

`deleted_user` Fires immediately after a user is deleted from the database.

i wonder if the `remove_user_from_blog` hook would work better. It would be called for each site

corsacca:
> Something to note here is that the record type continues to be a 'user.' We could revert it to another user type in this function, but there are possible security concerns with changing it to an access record when it should be a private record.

Yeah. i don't know how best to solve this one. I think we keep it as a user record and change the status to archived.

prykon:
I'm able to delete a user and then re-create it for the same contact, but there's a new bug:
The display name gets ignored. I set a display name for the user but it still shows the contact name. This worked fine with the previous code.

micahmills:
Where is the display name not displaying?

prykon:
In the contacts page and in the contact details page.
It does, however, appear in the wp-admin/users.php and in user-management/users/

prykon:
I re cloned the branch and it works fine now. Must have been an old file that was still in there somewhere.

corsacca:
thanks @micahmills  and @prykon !

PR #1408, opened 2021-06-21
1407 - Metrics: System Activity By Day

kodinkat:
### Implementation Summary

- A new metrics section, under contacts, has now been implemented; which highlights the activity by day.
- AMCharts' curve charts has been used to represent a curved timeline:
  - Timeline can be filtered to represent data from 3 months prior.
  - Using embedded scale-bar; users can zoom in/out specific date ranges; as well as scrolling between boundaries.
- Animated tooltip; which cycles through each timeline bullet, presenting a few top counts.
- Double clicking timeline bullets will display a breakdown of counts for a given day, covering the following metrics:
  - New Contacts
  - Active Groups
  - Active Churches
  - Baptisms
  - Other Post Creations
  - Seeker Path Updates
  - First Meetings
  - Ongoing Meetings
  - Health (Practising):
    - Baptism
    - Bible Study
    - Communion
    - Fellowship
    - Giving
    - Prayer
    - Praise
    - Sharing the Gospel
    - Leaders
    - Church Commitment

Screenshot 2021-06-21 at 09 55 50
Screenshot 2021-06-21 at 09 56 47


corsacca:
Hey @kodinkat , Nice work! 
The main goal for this feature is to see at a glance the activity for the past x amount of time.
I like the concept of the timeline. Is there a way to always show the values so the user doesn't have to hover/wait to see them? 

Other comments
- Active church/group aren't activity. Contacts that become active would be.
- The timeline does not show me activity for today.
- When clicking on a date we see a pie chart. A bar chart might be more applicable as we are not really looking at percentages.
Though i think in this case we might keep to a list (with icons?) and the number.
- Other post creations can be confusing (posts have a wide range of users)
- Show all milestones/health/seeker path values.
- In the time range select we could do this month, last month, 2 months ago (something similar to the critical path range select)
- Idea: select a value (new contacts) and have the timeline's day icon colors change depending on the # of new contacts.

kodinkat:
> Hey @kodinkat , Nice work!
> The main goal for this feature is to see at a glance the activity for the past x amount of time.
> I like the concept of the timeline. Is there a way to always show the values so the user doesn't have to hover/wait to see them?
> 
> Other comments
> 
> * Active church/group aren't activity. Contacts that become active would be.
> * The timeline does not show me activity for today.
> * When clicking on a date we see a pie chart. A bar chart might be more applicable as we are not really looking at percentages.
>   Though i think in this case we might keep to a list (with icons?) and the number.
> * Other post creations can be confusing (posts have a wide range of users)
> * Show all milestones/health/seeker path values.
> * In the time range select we could do this month, last month, 2 months ago (something similar to the critical path range select)
> * Idea: select a value (new contacts) and have the timeline's day icon colors change depending on the # of new contacts. Then it can answer: when in the month did we have the most new contacts.

@corsacca comments noted! I'll jump on this first thing!

corsacca:
@kodinkat  I have a baptism date set for Jul 21, but no activity is showing for that day.
With the baptized milestone it might look like the baptisms (off of the baptism date field) is a duplicate, but that is ok.

dt-metrics/combined/daily-activity.php
```diff
@@ -250,7 +252,7 @@ private function health_metrics( $start, $end, $group_fields ): array {
                 JOIN $wpdb->postmeta as d
                   ON a.ID=d.post_id
                      AND d.meta_key = 'group_type'
-                     AND ( d.meta_value = 'group' OR d.meta_value = 'church' )
+                     AND ( d.meta_value LIKE %s )
               WHERE a.post_status = 'publish'
```
corsacca, line 256: @kodinkat this whole join statement can be removed

dt-metrics/combined/daily-activity.php
```diff
@@ -265,7 +267,7 @@ private function health_metrics( $start, $end, $group_fields ): array {
                 JOIN $wpdb->postmeta as e
                   ON a.ID=e.post_id
                      AND e.meta_key = 'group_type'
-                     AND ( e.meta_value = 'group' OR e.meta_value = 'church' )
+                     AND ( e.meta_value LIKE %s )
```
corsacca, line 270: @kodinkat this whole join statement can be removed

corsacca:
Thanks @kodinkat

PR #1409, opened 2021-07-09
Avoid double task creation

corsacca:
update_post_user_meta_fields() only needs to be called once to process all the private fields.


micahmills (approved):
This looks good to me. It appears to work as expected on all field types.

corsacca:
Thanks for confirming @micahmills !

Issue #1373, opened 2021-05-18
Navigation Bar: Too many tabs


corsacca:
Sometimes space is limited on the navigation bar and there are many items:
![image](https://user-images.githubusercontent.com/24901539/118647075-d76d1000-b7d8-11eb-9b44-75932d27f665.png)

Collapse the extra tabs under a "More" drowdown to keep the tabs from wrapping.

prykon:
Is there a specific way we want it to appear?

corsacca:
There is code already there for the dropdown, so here we can reuse it and nest the overflow in the dropdown.

PR #1410, opened 2021-07-11
Top Bar Overflow for #1373

micahmills:
Add a 'More' dropdown if the menu items are to long for the space available. The number of items adjusts based on the screen size. The More button is hidden if it is not needed.


corsacca:
Thanks @micahmills !

Issue #1062, opened 2020-09-09
Choosing the system language when creating a new user


mikeallbutt:
When a DT instance is used by people who speak different languages, the instance by default sends the signup process on the system language. If the contact doesn't speak that language but gets the email invitation in it, it is confusing for the new user.

Have a field on the `Make User From Contact` modal (and in WP admin > New User) that offers the selection of what language the email invitation will be sent out in, and that setting is used to assign the language that the new user can use (but they can control once their user account process is complete).

mikeallbutt:
On the following screen image, a language selection field could be displayed at the bottom to inform/choose what language the invitation will be sent in.

<img width="615" alt="Screen Shot 2021-05-04 at 18 54 17" src="https://user-images.githubusercontent.com/43966676/117142336-32cce600-ad75-11eb-870d-4f184b731593.png">

If the invited contact becomes a user then that language should be offered as the default DT system language for them - unless they later choose to change it.

prykon:
I'm on it.

PR #1412, opened 2021-07-14
User creation language selector

micahmills:
Adds a language picker on user creation. related to #1062 

If the language has the Wordpress translation installed it will use that language for the user email. If not it will use the site default. This could be resolved with including our own welcome emails in the .po files for DT, but this PR doesn't do that.


corsacca (changes requested):

corsacca:
Nice work at @micahmills !
Works as expected. The new user locale will be the selected language.
The email was in English for me, as stated. I suggest we save that to be considered in a new issue.

corsacca:
Thank you @micahmills !

corsacca:
THey email part can be accomplished via: https://github.com/DiscipleTools/disciple-tools-theme/issues/1413

Issue #1414, opened 2021-07-21
CSS Refactor


micahmills:
The CSS/SCSS of the project has become a bit of a patchwork, with things being added over time. There is a lot of code that isn't needed anymore, duplicate code that could be gotten rid of, and some code used for older browsers that we may not need anymore. We could also do a lot to standardize the colors that we are using across the app and setup a proper color system.

There are a lot of ways to go about this, I would like to see us move to a more modern system using CSS custom properties instead of SASS variable since that would make theming and color switching easier and doesn't require a build step. 

This isn't a high priority need because what we have is working for now, but it is worth consideration and discussion.

micahmills:
After looking at the CSS using cssstats.com Here are some things that I think we can look at in a CSS refactor.

- [ ] standardize color declarations (currently have hex, reba, hsla)
- [ ] reduce colors used (currently 43 unique background colors, and 31 unique text colors many of which are very similar)
- [ ] standardize size units (currently px, em, rem, %)
- [ ] standardize font-sizes (currently 44 unique font sizes)
- [ ] reduce font families used. (currently 6 unique font families)
- [ ] When possible move inline styles into css
- [ ] reduce overall number of rulesets and selectors
- [ ] reduce duplicate declarations

corsacca:
44 unique font sizes? Oh man :)
Foundation came with a bunch of the scss variables. How much of this come from foundation?

micahmills:
I haven't really gotten far enough to determine the cause. The font families, and high number of rulesets and selectors, are likely because of foundation. The font sizes are likely because different people over the years have declared font sizes in rem, px, em, % etc. and make sometimes weird custom 1 off sizes for a font like 1.17188rem. 

I plan to do a more comprehensive look at what all is going on and how it can be fixed in the near future.

PR #1417, opened 2021-07-26
CSS refactor

micahmills:
This is a work in progress


micahmills:
This is related to #1414

corsacca:
@micahmills Thanks for your work.
There is a lot to review here.
I see px to rem changes like `4px` to `0.26667rem`. Is there a better way to do this that will be more consistent in future dev? 0.26667 isn't a common number.

Also please help me understand the color change from ffffff to fefefe

micahmills:
I am going to close this for now, and we can revisit it later.

PR #1419, opened 2021-07-27
Wizard https notification

prykon:
Added an HTTPS check for the WordPress Address URL setting.


prykon:
Going to clean up committed files and re-send.

PR #1420, opened 2021-07-27
Added HTTPS Check in Setup Wizard

prykon:
Setup wizard checks if WordPress Address URL is HTTPS and fires notification if not.


prykon:
Going to add PR to setup_wizard branch.

PR #1425, opened 2021-08-03
1386 fix emails sent in correct locale

### LINKED ISSUE #1386: Make sure all emails are sent with the correct user locale
Make sure that emails are sent in the user's language and not the sender's language or the system's language.

In `get_notification_message_html()` `get_notification_message_html` line ~872 there is code for switching to the user's language so that notifications are translated correctly.

This code needs to be abstracted out and applied to all emails and to email subject lines.

squigglybob:
resolves #1386


squigglybob:
@corsacca I think this one is ready to go, at least for the translations of notification emails

squigglybob:
@corsacca  I think for the password reset emails, the WP system translation files would need to be downloaded, and the user language set to the correct language in the /admin section. I think?

corsacca:
@squigglybob  We might want to translate our own password reset email.
The main issue is that we support some languages that are not part of the WP system translations. WP has no Swahili translation.

squigglybob:
could you give me a hint on where to look for the D.T password reset email stuff, thanks

corsacca:
Thank you @squigglybob !

PR #1426, opened 2021-08-04
1389 - Workflow Builder UI - When-X-Then-Y

kodinkat:


dt-workflows/workflows-defaults.php
```diff
@@ -0,0 +1,138 @@
...
+    public const ACTION_UPDATE = [ 'id' => 'update', 'label' => 'Updated To' ];
+    public const ACTION_APPEND = [ 'id' => 'append', 'label' => 'Appended With' ];
+    public const ACTION_CONNECT = [ 'id' => 'connect', 'label' => 'Connect To' ];
+    public const ACTION_REMOVE = [ 'id' => 'remove', 'label' => 'Removal Of' ];
+
+    public function __construct() {
+        add_filter( 'dt_workflows_defaults', [ $this, 'fetch_default_workflows_filter' ], 10, 2 );
+    }
+
+    public function fetch_default_workflows_filter( $post_type, $workflows ) {
```
corsacca, line 33: @kodinkat a filter's first param needs to be the option that is returned

PR #1429, opened 2021-08-04
Redirect wp-admin new user page to the front end user management add user page. Upgrade Add User page

corsacca:
- redirect the wp-admin new-user.php to /user-management/add-user/
- update add-user with all the needed functionality

The wp admin new user page an a multisite is confusing,  and it sends a whole bunch of different kinds of emails.
The "Add user" page on the front end is less confusing and sends on one kind of email and has a better sign up flow for the user too.

This will give us more control over the emails out, giving us the opportunity to translate them into all our languages.


corsacca:
image

PR #1430, opened 2021-08-06
feat: add filter to translate add user email #1413

### LINKED ISSUE #1413: Make sure new user invite email goes out in user's language
As a continuation of this issue: https://github.com/DiscipleTools/disciple-tools-theme/issues/1062

### LINKED ISSUE #1427: User Management: adding a new user
When adding a new user on a multisite from the front end:
image
And email is sent to the user with the activation link:
image
The user sets their password and then clicks the login link:
image
image
The login then does not work (it just redirects to the login page) because the user is not an the correct subsite.

A fix to this problem would be to make sure the activation link in the email goes out with the subsite included in the link:
Instead of `http://multisite.local/wp-login.php?action=rp&key=iOKDiQEV1htRiShhrwm3&login=testuser1%40multisite.local`
it would be `http://multisite.local/24/wp-login.php?action=rp&key=iOKDiQEV1htRiShhrwm3&login=testuser1%40multisite.local`
or `http://24.multisite.local/wp-login.php?action=rp&key=iOKDiQEV1htRiShhrwm3&login=testuser1%40multisite.local`

### LINKED ISSUE #1428: User Management: Adding an existing user
Scenario: a user is part of a multisite on a different subsite.
When adding the existing user on a different subsite from the front end:
image

The user is added to the subsite, but they are not sent any email or notification that they have been added.

Todo: make sure they get sent a welcome email

squigglybob:
resolves #1413 
resolves #1427 
resolves #1428

### DIFF
--- a/dt-notifications/notifications.php
+++ b/dt-notifications/notifications.php
@@ -41,13 +41,18 @@ function dt_notification_delete_by_post( $args = [] ) {
     Disciple_Tools_Notifications::delete_by_post( $args );
 }
 
-function dt_switch_locale_for_notifications( $user_id ) {
+function dt_switch_locale_for_notifications( $user_id, $locale = '' ) {
     // load the local for the destination usr so emails are sent our correctly.
-    $destination_user = get_user_by( 'id', $user_id );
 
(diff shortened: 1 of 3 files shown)


squigglybob:
@corsacca I think that's it for that one

PR #1441, opened 2021-08-25
extracted the langcode variable code to its own function

micahmills:
The langcode variable was being used several times so in an effort to keep the code DRY I made it its own shared function.


corsacca:
Thanks @micahmills !

PR #1446, opened 2021-08-31
feat: add without e.g. 'baptized by' option to connection filters #695

### LINKED ISSUE #695: List Filters: Exclude/not filter
The one thing I would be very useful would be to have the ability to find people that aren’t listed. eg.  I want to see a list of contacts who are baptised but aren't baptising.

NathTech:
resolves #695


NathTech:
Hi @corsacca.

I think this one is ready for review.
image
It currently looks like this which could be changed if needed.

The one thing I would say is that it doesn't quite enable Michael to do his initial example

eg. I want to see a list of contacts who are baptised but aren't baptising.

As the baptising flag is not a connection type and there is currently no way of negating the multi/key-selects

But other than that it works for connections

corsacca:
Thanks @squigglybob! This looks good.
I think it does fit the example in https://github.com/DiscipleTools/disciple-tools-theme/issues/695
If the "baptized by" field is filled, then the contact is baptized. e
If the "baptized" field isn't filled, then they have not baptized any one else (they are not baptising).

Issue #1444, opened 2021-08-30
Bulk Edit Comments


micahmills:
Bulk Edit Idea from a User:
How hard would it be to add the comment/activity box to the bulk edit more options area so we could click emailed them this message:

- then type in or copy/paste in the message?
- then it automatically updates ALL of those clicked in the Bulk Edit? 
- including the message?

because many times, it’s very appropriate to just send the same PING email to all the ones that need updated. this would really empower our coaches and save them tons of time and help them have way more positive feelings about their “grunt” work in DT. they would be able to focus more on responding to those who respond investing in those who are receptive and eager for the communication and interaction

PR #1449, opened 2021-08-31
Bulk Comments

micahmills:
adds a bulk comment box with type picker on contacts and the appropriate JS and CSS to make it work.


micahmills:
Resolves #1444

corsacca:
Thanks @micahmills !

Issue #1448, opened 2021-08-31
'emoji' comment icons not clickable on iOS mobile web version (Chrome)


trip-nine:
**Describe the bug**
'emoji' comment icons not clickable on iOS mobile web version (using Chrome)

**To Reproduce**
1. Log into the latest version of DT with a mobile device using a web browser (not the app)
2. go to a contact
3. create a test comment
4. try to 'like' your comment with the new comment icons

**Expected behavior**
what should happen is - the icon should be press-able and save, but it does not on a mobile device - logged in with a browser..

**Screenshots**
![image](https://user-images.githubusercontent.com/37304634/131543827-72db379a-7711-4928-92ff-6fe33be86332.png)

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser: Chrome
 - Version: DT 1.11

**Smartphone (please complete the following information):**
 - Device: iPhone 12 Pro Max
 - OS: iOS 14.7
 - Browser: Chrome
 - Version: newest

**Additional context**
NA

PR #1450, opened 2021-08-31
Reaction Click Bug fix

micahmills:
Bug fix for #1448.

event.submitter only works in Chrome and Firefox not in Safari (desktop or mobile). So we have to use a different method for getting the value of the reaction submit button.

### DIFF
--- a/dt-assets/js/comments.js
+++ b/dt-assets/js/comments.js
@@ -405,11 +405,17 @@ jQuery(document).ready(function($) {
       `
       reactionForm.addEventListener('submit', (e) => {
         e.preventDefault();
-        const formDataEntries = new FormData(e.target).entries();
-        const entries = Array.from(formDataEntries);
-        const reaction = entries.map((arr) => `${arr[1]}`);
+        if (e.submitter) {
+          //This gets the reaction submit button data for Chrome and other Blink based browsers that have the event submitter data.
+          var reaction = e.submitter.value;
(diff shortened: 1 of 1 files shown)


corsacca:
Under brave on linux the react api call is failing because the reaction is not being sent with the payload
image

micahmills:
Were the reaction buttons working on brave before?

corsacca:
@micahmills, yup. They work on master

micahmills:
This is a weird browser compatibility bug, where the event submitter data isn't provided by Safari/webkit based browsers but it is on Chrome/Blink based browsers. In another weird turn for some reason the work around for Safari doesn't work on Chrome/Blink based browsers. So we now have an if statement that checks if the submitter data is available and uses that if not is uses the work around.

corsacca:
@micahmills, Thanks! This fix worked for me.

Would it be simple to not use a form and do something like?
```
$('.pick-reaction-form .add-reaction').on('click', function(){
let key = $(this).val() // or change to use data-value and  (this).data('value')
})
```

micahmills:
I considered that, and that is probably how I would have coded it had I wrote the original code. I didn't want to change to much here because it didn't know what the side effects might be. I think this is a good bug fix for now, but we might think about refactoring this in the future.

corsacca:
Thanks @micahmills!

PR #1452, opened 2021-09-01
Settings permission shift

ChrisChasm:
One critical bug fix to a redirect in the settings page.
Added 'access_settings' as a new capability for accessing the settings page instead of access_contacts. This allows you to give only the strategist role, without making them a multiplier.


corsacca:
See https://github.com/DiscipleTools/disciple-tools-theme/pull/1455

PR #1462, opened 2021-09-09
feat: strategy for letting multipliers make multipliers

### LINKED ISSUE #1385: Strategy for letting users (multipliers) invite other users to the system
Currently a user with admin permissions needs to be the one to invite new users into the system.
This is great for more secure environments and follow up systems that need to keep a tight control over who sees what.

In other environments, where users are free to share that they are using D.T and where the system uses more of an oikos outreach strategy, we want users to be able to invite other users to use D.T with them.

This feature would be off by default, but the administrator could enable it for their instance.
The invite UI could be on a contact record ("create user from this contact') and maybe from the profile page.
New users would be sent and invitation and added as a multiplier. So by default they would not have access to any contacts.

squigglybob:
resolves #1385

### DIFF
--- a/dt-contacts/user-module.php
+++ b/dt-contacts/user-module.php
@@ -78,14 +78,19 @@ class DT_Contacts_User {
         global $wp_roles;
         if ( $post_type === "contacts" ){
             $contact = DT_Posts::get_post( $post_type, $post_id );
-            if ( current_user_can( "create_users" ) ){
+            if ( current_user_can( "create_users" ) || DT_User_Management::non_admins_can_make_users() ){
                 ?>
                 <li><a data-open="make-user-from-contact-modal">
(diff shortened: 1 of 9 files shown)


squigglybob:
@corsacca Ok I've come to a stop on this one. All the main stuff is done, but could do with checking to make sure I haven't opened it up too much or anything else i've missed

squigglybob:
Disregard step 3, the record didn't have an email the 1st time i tried.
5. The multiplier should not be able to create a user with a dispatch role.
Test: write thin is the browser console. The new user should just have the multiplier role
window.API.create_user({"user-display": "m4","user-email": "m4@dt.local", "user-roles": ['dispatcher']})
image
image
image

corsacca:
Awesome. This is great @squigglybob!

Let's add a "invite users" section to bottom of the the /settings page that links to the /user-management/add-new page
image

squigglybob:
@corsacca ok, link is added on settings screen

corsacca:
Thank you @squigglybob !

PR #1465, opened 2021-09-13
feat: User Activity Feed #1019

### LINKED ISSUE #1019: Metrics: User activity feed
As a user, I would like a way to see what I have done. Potentially this data could be accessed via the User Profile page (accessed via the user name on the blue nav strip) by having a tile listing history.

squigglybob:
resolves #1019


squigglybob:
- [ ]  Create activity page
- [ ]  Refactor activity list creation
- [ ]  Refactor user-activity endpoint
- [ ]  Create new route for getting filtered list of current users activity
- [ ]  Add which field was updated to query
- [ ]  Re-style activity log

squigglybob:
@corsacca ok, Hopefully that has made the styling a bit better and got rid of the non record titles.

PR #1471, opened 2021-09-15
Fix new post sequence

corsacca:


ChrisChasm (approved):
Wow. Great. That was a deeper dive than I expected.

PR #1477, opened 2021-09-21
Expanded Group Metrics

incraigulous:
Screen Shot 2021-09-21 at 2 13 39 PM

This pull request adds the following fields to the group dashboard: 

- Believer Count
- Baptized Count
- Baptized in Group Count

Believer count and baptized counts are auto-updated with milestones are added/removed from contacts within a group. "Baptized in Group" is not auto-updated, because there is currently no relationship in place between baptisms and groups. This could be implemented in the future if such a relationship was in place. 

It also expands the dt-metrics multiplying_groups_only query to include the new metrics along with church health metrics. This expanded data will be used in a forth-coming pull request to `disciple-tools/disciple-tools-dashboard`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ package-lock.json
 */development-only/*
 
 /composer.lock
+/.phpunit.result.cache
--- a/dt-groups/base-setup.php
+++ b/dt-groups/base-setup.php
@@ -27,7 +27,6 @@ class DT_Groups_Base extends DT_Module_Base {
         add_filter( 'dt_details_additional_tiles', [ $this, 'dt_details_additional_tiles' ], 10, 2 );
         add_action( 'dt_details_additional_section', [ $this, 'dt_details_additional_section' ], 20, 2 );
         add_action( 'wp_enqueue_scripts', [ $this, 'scripts' ], 99 );
-
         add_filter( 'dt_record_icon', [ $this, 'dt_record_icon' ], 10, 3 );
 
         // hooks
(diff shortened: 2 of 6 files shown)


corsacca:
Hey @incraigulous!

Thanks for the awesome Pull Request and for going above and beyond in creating php unit test!

Can you give us a little more detail on what why these fields might be important to anyone/everyone using Disciple.Tools?

Would you be able help me understand the difference between Baptized Count and Baptized in Group Count?

We could explore implementing "Baptized in Group Count" in a plugin if it isn't yet as valuable for the majority of D.T users.
Thanks!

Email me at corsacca at gmail.com and i'll get you added to our D.T slack org for more conversation.

PS, is there overlap with the four fields concept? https://zume.training/four-fields-tool/

incraigulous:
Hi, @corsacca! Sure, I'll provide some context. We're working with East-West (eastwest.org) on this project. Our current goal is to increase the usefulness of the GenMaps plugin by adding church health metrics. 

Here is a preview of our upcoming PR to the genmaps plugin: 

Screen Shot 2021-09-22 at 1 05 43 PM

Notice that we've made use of the new fields. We're also displaying the existing church health data from the Church Health panel in the group dashboard. 

Here is a file pulled directly from the genmapper repo that shows that our implementation of the health data is on par with what others in the M2M movement are already doing. We didn't create this. It was already in the repo. EW makes use of these metrics and trains churches/missions groups/leaders to do the same.

Screen Shot 2021-09-22 at 1 19 27 PM

genmapper-node-example-church-circles

From what I understand, the baptisms in group metric is one that isn't in use by every org, but that many do make use of. It is a means of separating the number of contacts baptized by the group (or at least since the group became active) from contacts that were already baptized when they joined the group.

I hope I was able to answer your questions, but if you have more, feel free to ask!

PR #1483, opened 2021-10-04
added last_comment field to fields_to_return

viktorsheep:
Added last comment column to Contact or Groups if the user has a customer field named `last_comment`.


corsacca:
Thanks for the pull request @viktorsheep !
It looks like we are loading all comments for each record. So if the list contains 100 records we'll be doing 100 database calls, loading 100s or 1000s of comments and then only returning ~100 of them. This can potentially slow down the request more than we want.
I would like to propose that we create a single database query that gets the last comment for each record in one go.

PR #1488, opened 2021-10-08
Icons in help modal are now all the same size.

prykon:
Makes all icons in help modals the same size, irregardless of the file dimensions.


corsacca:
Thanks @prykon!

PR #1490, opened 2021-10-11
feat: add expandability to activity log metrics

### LINKED ISSUE #1486: Expanded details for activity metrics
As per conversation with @corsacca 

On the new activity log metric page:

As a user I want to be able to see what changes I actually made to a record on this page so that I don't have to flick between the record page and the activity log.

E.g.
Group 1
* updated x, y, z
  * Expanded view:
    * set x to "i love pie"
    * set y to "update needed"
    * assigned to bob
* made 2 comment(s)
  * Expanded view
    * "We met and talked about the price of tea in Mexico"
    * "Called, no answer"

Group 2
etc

squigglybob:
resolves #1486


corsacca:
Chevrons are good.
Looks great on the users_management page.
On the metrics page i'm worrier it will be overlooked because it is far off to the right:
image
Thoughts on how to bring the chevron in closer?

squigglybob:
@corsacca We could make the paper half the width of the screen on larger screens. The width seems too large for the content.

squigglybob:
also, and this may be connected to creation of translateable object_notes, but the date here is going to need `moment`ifying
image

corsacca:
That looks better!
window.SHAREDFUNCTIONS.formatDate() for the date fields
we don't want to display the array field type, or any non usual field type

PR #1492, opened 2021-10-12
Made non-practiced elements more opaque

prykon:


prykon:
Fetching upstream in order to avoid conflicts.

PR #1494, opened 2021-10-13
feat: add activity highlights section to personal metrics

### LINKED ISSUE #711: Activity Highlights Feature: List/Feed of user's activity and comments across all records
From user: 
This could simply be a feature to pull all the comments or milestone events from the past 30 days (or a selectable timeframe) for a Multiplier’s Contacts and/or Groups records. For example a user could select the Month of April and it would give a chronological “ministry log” of the month. It would pull the testimony typed up at the beginning of the month, the Baptism mid-month (milestone or status change), the Group becoming a church right after that, and the other comments made about various contacts as the user interacted with them throughout the month. 

It should leave out the automated “Quick Actions” (didn’t pick up phone, phone ringing, etc) and other system notifications for greater clarity and only include the user's activity and comments instead of all activity and comments for the same records.

If people took the time to type up full testimonies and good ministry updates then their newsletter would consist of hitting this tab under Notifications and copying and pasting into Mailchimp or Pages or whatever they use.

squigglybob:
resolves #711


NathTech:
@corsacca On point 7 is that for contacts I baptized as well as contacts others baptized?

NathTech:
@corsacca I've implemented it for the contacts I have baptized as well, and put it into a table so as to reduce translation strings and bring clarity.

One thing it has highlighted though, is that the baptized by others is picking up those that a contact of mine has baptized as if it were the other way around.

e.g. my contact 1, was baptized by X (shows correctly in table)
my contact 1, baptized Y (shows as Y baptized contact 1 in table)

I'm not going to be able to fix that bit before Christmas, but the rest of the points should hopefully be done.

NathTech:
@corsacca I think this is ready for review now :)

corsacca:
thanks @squigglybob!! :)

Issue #1400, opened 2021-05-27
Transferred contact: get details from remote contact.


corsacca:
Add a tile to a contact record that has been transferred that:
- shows a summary gotten from the remote transferred contact
- textarea and the ability to push a comment to the remote contact with an update.

The summary can be retrieved when the user visits the contact.

PR #1501, opened 2021-10-21
#1400 - Retrieve Transferred Contact Summary

kodinkat:
- Retrieval of transferred contact summary
- Support remote contact updates.


corsacca:
Nice work!

image
Let's transform this to be more read only so the user does not try to click the buttons. It can be just the text
[icon]Status: Active
Seeker Path: X
Milestones: [icon] One, [icon] Two, [icon] Three

PR #1510, opened 2021-11-01
Group health fix

prykon:
- Added contingencies for missing labels and icons (see image)
- Added an image for missing group health elements
- Updated minor UI bug


corsacca:
Thank you @prykon

PR #1523, opened 2021-11-18
Adding a new full-color leader count icon

incraigulous:
This adds full-color support for use in the Genmapper, and anywhere else we need full-color icons.

Screen Shot 2021-11-18 at 12 30 17 PM
Screen Shot 2021-11-18 at 12 29 50 PM
.


corsacca:
Hey @incraigulous , 
Would it be possible to explore using a servant leader icon like https://thenounproject.com/term/servant/230531/
Instead of the traditional top down leader?

We use a foot in the members list, which isn't the most intuitive but does the job.

We try to communicate that the leader isn't the one that has everything figured out and somehow on top of others, but is there to serve the others and point to Jesus

incraigulous:
@corsacca, I love this so much! YES! Stay tuned.

incraigulous:
leader

@corsacca Updated!

PR #1541, opened 2021-12-14
Bulk magic link distribution

ChrisChasm:
- Modifies Magic Classes to have shared Apps tile strategy
- Adds Magic endpoint for sending magic links by email
- Adds two main variables to the registered magic link array to govern bulk support and apps tile support
- Adds qrcode graphic
- Edits the archive-template.php to include additional bulk support.


corsacca:
@ChrisChasm 
I can't unselect "share app".
It would make more sense as checkboxes
image
I

PR #1544, opened 2021-12-16
Assigned-To: New Suggestions Section

kodinkat:


dt-contacts/access-module.php
```diff
@@ -1526,7 +1526,7 @@ public function get_dispatch_list( WP_REST_Request $request ) {
             }
         }
 
-        return $list;
+        return apply_filters( "dt_get_dispatch_list", $list, DT_Posts::get_post( $params['post_type'], $params['post_id'], true, false ) );
```
corsacca, line 1529: @kodinkat, just pass in the post type and id in the filter and the do the get_post() in when running the filter.
This way the function is only running when `dt_get_dispatch_list` is hooked in to.
I think we can have $check_permission be true here, no?

Issue #1537, opened 2021-12-06
User's role(s) in User management table


corsacca:
Add a column for the user's role(s) in the user management list (/user-management/users/)
Add a dropdown to filter users by role.

PR #1548, opened 2021-12-20
#1537 - Updated User Management Table

kodinkat:
- #1537


corsacca:
Nice!
Can we just have individual languages in the dropdown? I'm confused at the "English, Spanish"
image
Use Role Labels instead of keys:
image

PR #1553, opened 2022-01-03
Add a "Open With..." box when someone clicks a phone link

micahmills:
on the details pane creates a "Open with.." box and creates links for Whatsapp, Signal, iMessage (if on an apple device) and a normal phone link.

The strings are created in JS and are not included in translation strings yet. If this looks like something that is usable in D.T core we can add the translations.


dt-assets/js/details.js
```diff
@@ -1024,9 +1024,47 @@ jQuery(document).ready(function($) {
...
+        for (const service in messagingServices) {
+          let link = messagingServices[service].link.replace('PHONE_NUMBER_NO_PLUS', phoneNumber.replace(/^((\+)|(00))/,"")).replace('PHONE_NUMBER', phoneNumber);
+
+          messagingServicesLinks = messagingServicesLinks + `<li><a href="${link}" title="Open With ${messagingServices[service].name}" target="_blank" class="phone-open-with-link"><img src="${messagingServices[service].icon}"/>${messagingServices[service].name}</a></li>`
+        }
+
+        let openWithDiv = `<div class="phone-open-with-container __${phoneNumber.replace(/^((\+)|(00))/,"")}">
+        <strong>Open With...</strong>
+          <ul>
+            <li><a href="${PhoneLink}" title="Open With Telephone" target="_blank" class="phone-open-with-link"><img src="${window.lodash.escape( window.wpApiShare.template_dir )}/dt-assets/images/phone.svg"/> Phone</a></li>
```
corsacca, line 1053: Translate "Phone"

corsacca: And remove the space before it so it lines up with the other labels.

corsacca:
Thanks @micahmills!

Issue #1527, opened 2021-11-24
List: Ability to modify a search or filter query by removing a parameter


corsacca:
Clicking on the x would remove the parameter.
![image](https://user-images.githubusercontent.com/24901539/142852238-5c033a30-a963-4a7a-a5b5-2ee71a26cee2.png)

If no parameters are left, return the user to the default filter "All my contacts"

PR #1558, opened 2022-01-07
#1527 - Closable Filter Labels

kodinkat:
- #1527


corsacca:
Great work @kodinkat !
Can you have the style/ui for the x match that of the connection field?
image

The x with a border around it and the hover over it changing its color

corsacca:
Thanks @kodinkat !

Issue #1551, opened 2022-01-03
Ability to set which email notifications are sent from instead of the admin email


corsacca:
IN WP Admin > Setting (D.T) 
![image](https://user-images.githubusercontent.com/24901539/148560085-04734a8b-e705-4957-8e1e-c0bdd868011b.png)

PR #1559, opened 2022-01-07
#1551 - Ability To Specify Alternative From Email Addresses

kodinkat:
- #1551


corsacca:
@kodinkat Nce work

Let's throw this in a table to make it more organized.
Can we also specify want the default is?
maybe " Specify notification from email address. Leave blank to use default (wordpress@example.com)

image

corsacca:
Thanks @kodinkat!

Issue #1561, opened 2022-01-10
API, make sure comments are created with correct dates.


corsacca:
Return an error if a comment is created with an invalid date.
Add some phpunit tests to confirm that comments with valid dates are created correctly and comments with invalid dates return errors.
Comments created with no date specified should return a comment with the current date/time.
See https://developers.disciple.tools/theme-core/api-posts/post-comments for valid date format.
`7/6/2009 14:16` is not a valid date.

PR #1566, opened 2022-01-13
#1561 - API Comments Date Format Validation

kodinkat:
- #1561

### DIFF
--- a/tests/dt-posts/dt-posts/unit-test-post-comments.php
+++ b/tests/dt-posts/dt-posts/unit-test-post-comments.php
@@ -1,0 +1,99 @@
+<?php
+require_once( get_template_directory() . '/tests/dt-posts/tests-setup.php' );
+
+class DT_Posts_DT_Posts_Post_Comments extends WP_UnitTestCase {
+
+    public static $sample_contact = [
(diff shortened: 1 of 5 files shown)


tests/dt-posts/dt-posts/unit-test-post-comments.php
```diff
@@ -0,0 +1,99 @@
...
+
+        $this->assertNotWPError( $comment_id );
+        $this->assertSame( 1, (int) $comments['total'] );
+        $this->assertSame( $comment_html, $comments['comments'][0]['comment_content'] );
+    }
+
+    public function test_comment_creation_with_invalid_data() {
+
+        $comment_args  = [
+            'comment_date' => '0000-00-00 00:00:00'
```
corsacca, line 53: @kodinkat , let's test a couple more invalid? cases. I can think of
- '' (empty string)
- null,
- 7/6/2009 14:16

kodinkat: ....however, any string literal; i.e. `null`; which does not adhere with required date format, will trigger exception...

corsacca:
Thanks @kodinkat !

Issue #1569, opened 2022-01-14
Metrics Side menu, highlight current page


corsacca:
When on a metrics page, highlight on the left which page is being viewed.
Also apply to user management view.
![image](https://user-images.githubusercontent.com/24901539/149490179-b34e5fe7-c567-4287-9f73-24a14d0ad80f.png)

PR #1574, opened 2022-01-26
#1569 - Metrics Side Menu, Current Page Highlights

kodinkat:
- #1569 

Screenshot 2022-01-26 at 16 45 40

Screenshot 2022-01-26 at 16 45 05

Screenshot 2022-01-26 at 16 44 47


dt-users/user-management.php
```diff
@@ -15,7 +15,7 @@ public static function instance() {
     public function __construct() {
         $url_path = dt_get_url_path();
         if ( self::has_permission() || self::non_admins_can_make_users() ) {
-            if ( strpos( $url_path, 'user-management/user' ) !== false || ( strpos( $url_path, 'user-management/add-user' ) !== false && ( current_user_can( "create_users" ) || self::non_admins_can_make_users() ) ) ){
+            if ( strpos( $url_path, 'user-management/user' ) !== false || strpos( $url_path, 'user-management/hover-map' ) !== false || ( strpos( $url_path, 'user-management/add-user' ) !== false && ( current_user_can( "create_users" ) || self::non_admins_can_make_users() ) ) ){
```
corsacca, line 18: @kodinkat Lets find a better strategy for when we add more user management pages.

corsacca (changes requested):

dt-users/user-management.php
kodinkat: @corsacca As `Hover Map` template is shared across metrics and user-management views, decided to shift away from class name, but instead; based conditional logic on `url_path` structure...

dt-users/user-management.php
```diff
@@ -15,16 +15,16 @@ public static function instance() {
     public function __construct() {
         $url_path = dt_get_url_path();
         if ( self::has_permission() || self::non_admins_can_make_users() ) {
-            if ( strpos( $url_path, 'user-management/user' ) !== false || ( strpos( $url_path, 'user-management/add-user' ) !== false && ( current_user_can( "create_users" ) || self::non_admins_can_make_users() ) ) ){
+            if ( str_contains( $url_path, 'user-management/' ) && ( current_user_can( "create_users" ) || self::non_admins_can_make_users() ) ){
```
corsacca, line 18: @kodinkat is this part still needed?

kodinkat: @corsacca I believe so; so as to future-proof against addition of any new page sections..

corsacca: @kodinkat the js part that we would need is no longer there.

corsacca: Oh. I mean for line 128

corsacca: and 24 i think

corsacca (approved):

PR #1575, opened 2022-01-27
disable inputs if user can view a post but not update it.

micahmills:
This currently doesn't work if someone is sub assigned to a contact.

### DIFF
--- a/dt-contacts/access-module.php
+++ b/dt-contacts/access-module.php
@@ -414,7 +414,10 @@ class DT_Contacts_Access extends DT_Module_Base {
 
     public function dt_render_field_for_display_template( $post, $field_type, $field_key, $required_tag, $display_field_id ){
         $contact_fields = DT_Posts::get_post_field_settings( "contacts" );
-
+        $disabled = 'disabled';
+        if ( current_user_can( "assign_any_contacts" ) || isset( $post["assigned_to"]["id"] ) && $post["assigned_to"]["id"] == get_current_user_id() ) {
+            $disabled = '';
+        }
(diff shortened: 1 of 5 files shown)


dt-contacts/access-module.php
```diff
@@ -414,7 +414,10 @@ public function dt_details_additional_tiles( $sections, $post_type = "" ){
 
     public function dt_render_field_for_display_template( $post, $field_type, $field_key, $required_tag, $display_field_id ){
         $contact_fields = DT_Posts::get_post_field_settings( "contacts" );
-
+        $disabled = 'disabled';
+        if ( current_user_can( "assign_any_contacts" ) || isset( $post["assigned_to"]["id"] ) && $post["assigned_to"]["id"] == get_current_user_id() ) {
```
corsacca, line 418: i think we should use the current_user_can( "update_$post_type") to determine this one

micahmills: Ah yes I changed it in the render function but forgot it here

corsacca: Actually, DT_Posts::can_update( $post_type) is better

corsacca: because there are server caps that could give permission to update

micahmills:
This should disable all the inputs on a post that a user has access to view but not change. You can use the prayer request plugin as an example of these types of posts (https://github.com/micahmills/disciple-tools-prayer-request). 

On typeaheads, I disabled the search input with an HTML attribute, and the cancel buttons using CSS pointer-events: none. You can still click the links to go to the record that is connected in a multi-select typeahead.

corsacca:
Thanks @micahmills !

Issue #1044, opened 2020-08-28
Update needed triggers:  UI and/or filter to add more triggers.


corsacca:
Let the user or a plugin add more triggers to the update needed options.

The update needed triggers are defined in config-site-defaults.php and then stored within the wp_options table.

Here is a way to add to them for now from a plugin.

```
add_action( "init", [ $this, "setup" ] );
public function setup(){
        $site_options           = dt_get_option( "dt_site_options" );
        //@todo check if you've added the option already. You don't want to add it multiple times. Maybe: if count( $site_options["update_required"]["options"] ) < x
        if ( false ) {
            $site_options["update_required"]["options"][] = [
                "status"      => "active",
                "seeker_path" => "your_custom_path",
                "days"        => 3,
                "comment"     => __( "Custom comment", 'disciple_tools' )
            ];
            update_option( 'dt_site_options', $site_options, true );
        }
}
```

PR #1580, opened 2022-02-04
#1044 - Custom Seeker Path Option Triggers Support

kodinkat:
- #1044

### DIFF
--- a/dt-core/admin/menu/tabs/tab-custom-fields.php
+++ b/dt-core/admin/menu/tabs/tab-custom-fields.php
@@ -827,8 +827,8 @@ class Disciple_Tools_Tab_Custom_Fields extends Disciple_Tools_Abstract_Menu_Base
             }
 
             // key_select and multi_options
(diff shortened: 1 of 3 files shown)


corsacca:
@kodinkat 
image
-- 
Issue: i had an existing custom option and it did not show up.
Let's set the default to 30 days instead of 3 
---
I've added a custom option. After 3 days will it post a comment? What message will it post?
Maybe default to:
We haven't heard about this person in a while. Do you have an update for this contact?

kodinkat:
> ## @kodinkat
> image
> ## Issue: i had an existing custom option and it did not show up.
> Let's set the default to 30 days instead of 3
> I've added a custom option. After 3 days will it post a comment? What message will it post? Maybe default to: We haven't heard about this person in a while. Do you have an update for this contact?

Screenshot 2022-02-08 at 13 17 47

Screenshot 2022-02-08 at 13 18 10

PR #1610, opened 2022-03-29
Updated Contacts Filter List - Optional Comms Filter

kodinkat:


corsacca (approved):

dt-posts/posts.php
```diff
@@ -1066,26 +1066,28 @@ public static function search_viewable_post( string $post_type, array $query, bo
...
 
             if ( empty( $fields_to_search ) ) {
-                $post_query .= "AND ( ( p.post_title LIKE '%" . esc_sql( $search ) . "%' )
-                    OR p.ID IN ( SELECT post_id
-                                FROM $wpdb->postmeta
-                                WHERE meta_key LIKE 'contact_%'
-                                AND REPLACE( meta_value, ' ', '') LIKE '%" . esc_sql( str_replace( ' ', '', $search ) ) . "%'
-                    )
-                ";
+                $post_query .= "AND ( ( p.post_title LIKE '%" . esc_sql( $search ) . "%' )";
```
corsacca, line 1069: @kodinkat i think this part stays the same. 
On default it should search contacts_%

corsacca, line 1069: @kodinkat i think this part stays the same. 
On default it should search contacts_%

Issue #1604, opened 2022-03-11
Lists: Closing a text search


corsacca:
When closing a text search filter, the list should revert to "all contacts", or "all groups"
![closing_text_search](https://user-images.githubusercontent.com/24901539/157895019-9e966a2f-da89-4f4c-9be2-6003b4d98a34.gif)

PR #1611, opened 2022-03-31
#1604 - Lists: Closing A Text Search

kodinkat:
- #1604

### DIFF
--- a/dt-assets/js/modular-list.js
+++ b/dt-assets/js/modular-list.js
@@ -195,21 +195,9 @@
             // Remove text property, to force a return to all filtered view
             delete current_filter.query['text'];
 
-            // Determine id to be found
-            let id = '';
-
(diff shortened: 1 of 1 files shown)


dt-assets/js/modular-list.js
```diff
@@ -190,6 +190,29 @@
...
 
+          } else if (current_filter.query['text']) {
+
+            // Remove text property, to force a return to all filtered view
+            delete current_filter.query['text'];
+
+            // Determine id to be found
+            let id = '';
+
+            switch (list_settings.post_type) {
```
corsacca, line 201: @kodinkat what happens in we are in a different post type?

corsacca, line 201: @kodinkat what happens in we are in a different post type?

corsacca:
Thanks @kodinkat

PR #1612, opened 2022-04-01
Additional Support For - #19 - Magic Link Form Template Creator

kodinkat:


dt-reports/magic-url-class.php
```diff
@@ -547,7 +549,13 @@ public function api_create_key( WP_REST_Request $request ){
...
          * @return array
          */
-        public static function list_bulk_send() {
+        public static function list_bulk_send( $post_type = null ) {
+
+            /**
+             * Ensure to pre-load any available magic link templates prior to
+             * main list_bulk_send() execution.
+             */
+            do_action( 'dt_magic_link_load_templates', $post_type );
```
corsacca, line 558: what if we hooked in to `dt_magic_url_register_types` instead 
Since that is what is called in the next line?

kodinkat: Spawns a recursive loop... Approach to be discussed further..

PR #1618, opened 2022-04-07
Centralize magic links app tile.

corsacca:


dt-reports/magic-url-setup.php
```diff
@@ -0,0 +1,159 @@
+<?php
+if ( ! defined( 'ABSPATH' ) ) { exit; } // Exit if accessed directly
+
+class DT_Magic_URL_Setup {
+    private static $_instance = null;
```
ChrisChasm, line 5: You don't need the singleton if you are loading the class with 'new' at line 159

corsacca: Awesome!

Issue #1629, opened 2022-04-20
Magic Links > Apps Tile > Send Email with Link


corsacca:
Show the email found or let the user enter in an email.
![image](https://user-images.githubusercontent.com/24901539/164200736-32b3a913-f69a-48e9-828f-2be2d27871ff.png)

PR #1631, opened 2022-04-22
#1629 - Magic Links > Apps Tile > Send Email with Link

kodinkat:
- #1629

### DIFF
--- a/dt-reports/magic-url-endpoints.php
+++ b/dt-reports/magic-url-endpoints.php
@@ -121,8 +121,8 @@ class Disciple_Tools_Magic_Endpoints
             }
 
             // check if email exists to send to
-            if ( ! isset( $post_record['contact_email'][0] ) ) {
-                $errors[$post_id] = 'no email';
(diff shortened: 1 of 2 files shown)


corsacca:
Disable button unless the input contains a valid email
image

corsacca:
@kodinkat  Let's also add the valid email check on the server side.

dt-reports/magic-url-endpoints.php
```diff
@@ -143,6 +143,13 @@ public function email_magic( WP_REST_Request $request ) {
 
             // build email
             $email   = $params['email'] ?? $post_record['contact_email'][0]['value'];
+
+            // final email format sanity check
+            if ( ! filter_var( $email, FILTER_VALIDATE_EMAIL ) ) {
+                $errors[ $post_id ] = 'invalid email format';
+                continue;
```
corsacca, line 150: @kodinkat what does continue do here?

corsacca: never mind

PR #1632, opened 2022-04-25
feat: add a toggle into the list page #1578

### LINKED ISSUE #1578: Recreate toggle filter to hide/show archived records
Prior to Version 1 modularization, contact and group records had a toggle that could hide or display archived contacts.  Being able to filter out archived records at a glance is a super helpful feature.

squigglybob:
resolves #1578

### DIFF
--- a/dt-contacts/base-setup.php
+++ b/dt-contacts/base-setup.php
@@ -536,6 +536,9 @@ class DT_Contacts_Base {
                 ],
                 'query' => [
                     'sort' => '-post_date',
+                    'overall_status' => [
+                        '-closed'
+                    ]
                 ],
             ];
 
--- a/dt-contacts/base-setup.php
+++ b/dt-contacts/base-setup.php
@@ -61,8 +61,10 @@ class DT_Contacts_Base {
         if ( $post_type === $this->post_type ){
             $settings['label_singular'] = __( 'Contact', 'disciple_tools' );
             $settings['label_plural'] = __( 'Contacts', 'disciple_tools' );
-            $settings['status_field'] = "overall_status";
-            $settings['archived_status_key'] = "closed";
+            $settings['status_field'] = [
+                "status_key" => "overall_status",
+                "archived_key" => "closed",
+            ];
         }
         return $settings;
     }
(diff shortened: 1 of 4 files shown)


squigglybob:
- [ ] add toggle to the list page
- [ ] add key value pair for archived status to contacts and groups post type
- [ ] provide field key and archived value to list page
- [ ] add correct filter option to status to filter out archived items
- [ ] remove this filter when the toggle is on to include archived items
- [ ] if the status filter is being set make sure that this overrides filtering out archived items
- [ ] add key value pair to other post types

squigglybob:
-closed doesn't seem to be honoured for recently viewed records

dt-contacts/base-setup.php
```diff
@@ -61,6 +61,8 @@ public function dt_get_post_type_settings( $settings, $post_type ){
         if ( $post_type === $this->post_type ){
             $settings['label_singular'] = __( 'Contact', 'disciple_tools' );
             $settings['label_plural'] = __( 'Contacts', 'disciple_tools' );
+            $settings['status_field'] = "overall_status";
+            $settings['archived_status_key'] = "closed";
```
corsacca, line 65: @squigglybob 
can we do something like this so we can expand it more later if needed?
```
status_field => [
field_key=> "overall_status"
archived_key=> "closed"
]
```

corsacca: alternately we would add a flag to the existing status field

squigglybob: good idea. what sort of things do you think it might expand into? Would it always be to do with the status field? If not maybe we could give it a more generic name instead of status_field...

squigglybob:
@corsacca as long as these phpcs fixes do the job, I think this one is ready for review

squigglybob:
@corsacca ok, the toggle is global now, and all of the past code was used, so no time lost... phew

dt-contacts/base-setup.php
```diff
@@ -534,6 +538,9 @@ public static function dt_user_list_filters( $filters, $post_type ){
                 ],
                 'query' => [
                     'sort' => '-post_date',
+                    'overall_status' => [
```
corsacca, line 541: @squigglybob Does this one do anything now?

squigglybob: yes it does. Having the query stuff directly under the query object still works along side the nested fields object. The code for checking for -closed also still checks in both places too.

corsacca:
Thanks @squigglybob !

PR #1635, opened 2022-04-26
RELATED_TO: #21 - Use POST Request Methods To Support Large Payloads

kodinkat:


dt-core/global-functions.php
```diff
@@ -714,7 +714,11 @@ function render_field_for_display( $field_key, $fields, $post, $show_extra_contr
                     <div class="small button-group" style="display: inline-block">
                         <?php foreach ( $fields[$field_key]["default"] as $option_key => $option_value ): ?>
                             <?php
-                            $class = ( in_array( $option_key, $post[$field_key] ?? [] ) ) ?
+                            $haystack = $post[ $field_key ] ?? [];
```
corsacca, line 717: @kodinkat what is happening here?

kodinkat: Safeguarding against a bug I unearthed; which resulted in a string field value instead of the required array... It's most likely an edge-case, as the offending field, stems from a plugin... However, thought I'd work in some safeguards..

corsacca:
Thanks @kodinkat !

Issue #1398, opened 2021-05-26
Group leadership icon


mikeallbutt:
As seen in the attached image, the leadership icon (the black footprint) blends in too much in the members list of a group record. Also, having all the grey footprints on the right side is a bit confusing too.

![Screenshot_20210506-072023_Firefox](https://user-images.githubusercontent.com/43966676/119712144-eaf62780-be25-11eb-9913-db5dbcd296f4.png)

The suggestion is to move the grey footprints to the far left of the row - which leaves just the X icon on the far right - and remove the black footprint from the icons (representing bible and baptism) and if a member is a leader, then the footprint next to their name becomes black.

corsacca:
How about keeping the leader icon next to the x, but changing it to black if it is selected?

prykon:
> How about keeping the leader icon next to the x, but changing it to black if it is selected?

This, then?

PR #1636, opened 2022-04-27
Cleaned up leader footstep ui in 'Member List' tile

prykon:
Responds to: #1398 

**Current behavior:**
- Black Footstep icon removed for leaders
- Leader toggle footstep icon is now black if member is leader

### DIFF
--- a/dt-groups/groups.js
+++ b/dt-groups/groups.js
@@ -129,6 +129,7 @@ jQuery(document).ready(function($) {
   }
   /* End Health Metrics*/
 
+  let { template_dir } = window.wpApiShare
 
   /* Member List*/
   let memberList = $('.member-list')
@@ -151,6 +152,8 @@ jQuery(document).ready(function($) {
         leaderStatus = 'leader';
         leaderStyle = 'color:black;';
       }
+      
+
       const contactStatusHTML = ( member.data && member.data.overall_status )
         ? `<i class="fi-torso small" style="color: ${window.lodash.escape( member.data.overall_status.color )}" title="${window.lodash.escape( member.data.overall_status.label )}"></i>`
         : '<i class="fi-torso small"></i>'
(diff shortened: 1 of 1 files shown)


corsacca:
Awesome work @prykon!

leader
The in between state is a bit confusing. 
@prykon what do you think of this idea?
When the foot is clicked: transform the icon into the loading circle. When the API call comes back, show the correct foot.

prykon:
> Awesome work @prykon!
> 
> leader The in between state is a bit confusing. @prykon what do you think of this idea? When the foot is clicked: transform the icon into the loading circle. When the API call comes back, show the correct foot.

Sounds good! I'm on it...

prykon:
The spinner appears when you make a member a leader. When you remove leadership, the row disappears before the spinner can be shown.

(screenshot frame rate slower than reality)

dt-groups/groups.js
```diff
@@ -204,14 +207,15 @@ jQuery(document).ready(function($) {
...
   $(document).on("click", ".make-leader", function () {
+    $(this).children('i').attr('class', 'small')
+    let spinner = `<img src="/wp-content/themes/disciple-tools-theme/dt-core/admin/img/spinner.svg" width="15px">`
+    $(this).append(spinner)
     let id = $(this).data('id')
     let remove = false
     let existingLeaderIcon = $(`.member-row[data-id="${id}"] .leader`)
     if( window.lodash.find( post.leaders || [], {ID: id}) || existingLeaderIcon.length !== 0){
       remove = true
       existingLeaderIcon.remove()
```
corsacca, line 218: @prykon  Remove `existingLeaderIcon.remove()` and the spinner will work both ways:

dt-groups/groups.js
```diff
@@ -204,14 +207,15 @@ jQuery(document).ready(function($) {
     }
   })
   $(document).on("click", ".make-leader", function () {
+    $(this).children('i').attr('class', 'small')
+    let spinner = `<img src="/wp-content/themes/disciple-tools-theme/dt-core/admin/img/spinner.svg" width="15px">`
```
corsacca, line 211: use this format: ``

corsacca: spinner.svg is in the theme root.

corsacca (changes requested):

corsacca:
Very nice!

PR #1639, opened 2022-04-29
Suggestion for D.T font icons library.

corsacca:
Ability to let 'font-icon' instead of 'icon'
```
     'font-icon' => 'fi-dislike',
//   'icon' => get_template_directory_uri() . "/dt-assets/images/nametag.svg?v=2",
```


micahmills (approved):
The only thing I see that is slightly off is that it looks like the font icons are slightly smaller (14px instead of 15px). Other than that this looks good. 

Are we thinking that we will have a font-library sort of picker on the WP admin side so people can choose from there for custom icons?

corsacca:
@micahmills i set the font size to 15px. Nice catch.

> Are we thinking that we will have a font-library sort of picker on the WP admin side so people can choose from there for custom icons?
This would be nice i think when create new custom fields.
More discussion on the issue

corsacca:
example with mdi:
```
$fields['assigned_to'] = [
                'name'        => __( 'Assigned To', 'disciple_tools' ),
                'description' => __( "Select the main person who is responsible for reporting on this group.", 'disciple_tools' ),
                'type'        => 'user_select',
                'default'     => '',
                'tile' => 'status',
//                'icon' => get_template_directory_uri() . '/dt-assets/images/assigned-to.svg?v=2',
                'font-icon' => "mdi mdi-account",
                "show_in_table" => 16,
            ];
```

PR #1641, opened 2022-04-29
feat: list page filter from url query params

### LINKED ISSUE #1503: List page filters in url parameters
It would be really nice to be able to build an filter strings url to send to other users so they can see the same filter you do.
Examples:
- `example.com/contacts?tags={one,two,three}&overall_status=active` // get all active contact that have one of these tags
- `example.com/groups?assigned_to={12,43}` // groups assigned to user 12 or user 43

To work with the default field types: https://developers.disciple.tools/theme-core/customization/fields#field-types
Except location_meta and array

Also the ability to set what columns to show in this list table  
`?table={name,assigned_to,my_awesome_cool_field}`

squigglybob:
resolves #1503

### DIFF
--- a/dt-assets/js/details.js
+++ b/dt-assets/js/details.js
@@ -914,8 +914,8 @@ jQuery(document).ready(function($) {
         href: function (item) {
           const postType = window.wpApiShare.post_type
           const query =  window.SHAREDFUNCTIONS.createCustomFilter('tags', [item.name], `Tags: ${item.name}`)
-          const encodedFilterLabel = window.SHAREDFUNCTIONS.uriEncodeFilter(query)
-          return window.wpApiShare.site_url + `/${postType}?query=${encodedFilterLabel}`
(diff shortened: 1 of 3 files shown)


squigglybob:
@squigglybob double check that the columns to show variable gets used from the link

squigglybob:
Hopes and dreams:

* to be able to create a filter on any instance, and adjust for their instance and send them a URL

for that to work 

* to be able to pop to the list view in a certain filter is desirable (in a documented way)
* to go to the list page from anywhere in the site to a particular filter
* a list of suggested filters with urls, you click on one and save it to your filter list
* to be able to change the columns viewable per filter rather than be global.
* could have a split by accordian drop down when you click on a filter, and split by

squigglybob:
@corsacca this one should be ready for checking as long as there are no issues with the checks

corsacca:
Thanks @squigglybob !!!!

PR #1653, opened 2022-05-05
#1619 - Remember Post Type WPAdmin Selection

### LINKED ISSUE #1619: `Select Post type` in WPADMIN Field explorer tab doesn't remember selection
At present the `Contacts` post type is always selected. When you select a post type other than `Contacts` the page reloads whit the correct info but the field is always `Contacts`. It should be indicating which field type is being used.

Screen Shot 2022-03-23 at 13 45 39

Where it says `Fields on this Instance` it would be useful to add to the start of the phrase the type of post type. ie. `{People Groups/Contacts/Groups/Trainings} Fields on this Instance'

kodinkat:
fixes #1619


corsacca:
Thanks @kodinkat

PR #1655, opened 2022-05-05
feat: add custom number field to new field options in admin

### LINKED ISSUE #1587: Add support to the number field for min and max values
Example: Count fields would have a min of 0 and not allow for negative numbers

The min and max parameters would:
- update the theme UI to accept only numbers within the bounds
- add a check to the api to not accept numbers outside the bounds

### LINKED ISSUE #1623: Add Custom Number Field as an option when creating a new field.
Oh. Alas. Where is it?
Check php unit test are created too.
image

squigglybob:
resolves #1623
resolves #1587


squigglybob:
@corsacca I need to write a php_unit test for this and then it should be done

squigglybob:
@corsacca I can confirm that the unit tests exist for the number fields, as I broke them in the commit before last 😄

squigglybob:
@corsacca I can't check the button in your comment, but the test is now in there for update and creation

corsacca:
This is awesome. Thanks @squigglybob

PR #1666, opened 2022-05-06
feat: hide tiles set by group tile preferences

### LINKED ISSUE #1626: Group Tile Preferences not being respected in tiles
The option to not use and show the `group tile preferences` of `church metrics` and `four fields` are not being respected.

As seen in the image, the option is deselected but the fields and tiles still appear in the back and front end.
Screen Shot 2022-04-19 at 22 18 17
Screen Shot 2022-04-19 at 22 19 12
Screen Shot 2022-04-19 at 22 20 18

The only way to hide the tile is to do it as seen in the next image. But IMO these tiles should not be being offered if the `group tile preferences` were being respected.
Screen Shot 2022-04-19 at 22 50 04

squigglybob:
resolves #1626


squigglybob:
@corsacca not sure why the previous travis builds were failing, looked like something to do with php codesniffer throwing a wobbly.

If this comes backs ok, the new filter has been applied and is working nicely for the admin and ui

corsacca (approved):

corsacca:
Thank you @squigglybob for your hard work on this one!

PR #1669, opened 2022-05-06
Ensure New Rows Adopt Currently Displayed Field Layout

kodinkat:


corsacca:
@kodinkat Location field and typeaheads don't seem to be playing ball an new added rows:
image

corsacca:
Nice work @kodinkat !

PR #1671, opened 2022-05-09
fix: always show the comments and activity checkbox filters

### LINKED ISSUE #1649: Comment types are not showing on other post types
"aouea uo" has a different comment type. This new comment type should show in the comment filters

image

squigglybob:
resolves #1649


squigglybob:
@corsacca this one should be good to go. It makes sure that the comments and activity checkboxes are always showing and never hidden

PR #1675, opened 2022-05-10
Fixing error when mapbox-search-widget isn’t registered.

incraigulous:
Screen Shot 2022-05-10 at 10 00 34 AM

I'm not sure why it isn't registered. The key exists for me locally. I only get this error in our staging environment.


corsacca:
Thanks @incraigulous !

PR #1677, opened 2022-05-11
refactor: modular-list.js

squigglybob:
put top level blocks of code into self describing functions
also move all on page load functionality into one function

### DIFF
--- a/dt-assets/js/modular-list.js
+++ b/dt-assets/js/modular-list.js
@@ -19,6 +19,7 @@
   let current_user_id = wpApiNotifications.current_user_id;
   let mobile_breakpoint = 1024
   let clearSearchButton = $('.search-input__clear-button')
+  let getFilterCountsPromise = null
   const { status_field } = list_settings.post_type_settings
   const { status_key, archived_key } = status_field ? status_field : {}
   const filterOutArchivedItemsKey = `-${archived_key}`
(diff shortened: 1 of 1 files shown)


squigglybob:
@corsacca i don't think that I have changed or lost any functionality on this refactor. If you think it has any smells, then I can do it again, and do each piece of functionality one by one in their own commits to make it even more checkable

squigglybob:
@corsacca one other thing I did notice from the fresh branch from master (before refactoring) that the sort order is now different to how it used to be, and any changes to the sort order aren't saved between refreshes.

Just flagging it up to check if that's the way it should be now :)

dt-assets/js/modular-list.js
```diff
@@ -32,70 +33,96 @@
...
 
-  //determine list columns
-  if ( window.lodash.isEmpty(fields_to_show_in_table)){
-    fields_to_show_in_table = list_settings.fields_to_show_in_table
+  let check_first_filter = function (){
+    $('#list-filter-tabs .accordion-item a')[0].click()
+    $($('.js-list-view')[0]).prop('checked', true)
+  }
+
+  function determineListColumns(fieldsToShowInTable, listSettings) {
```
corsacca, line 119: @squigglybob why pass in `listSettings` when it is available in root?

squigglybob: that's a good question. I was trying to go for the paradigm of having every dependency passed into the function through the params. But not sure if it's really necessary or not.

I was also trying to have the functions have no side effects, like modifying incoming params or global params, but failed in that also.

I'm happy to cut them out if you would prefer to just grab the global?

PR #1681, opened 2022-05-12
fix: add and use a utf-8 capable base64 encoder and decoder 

squigglybob:
courtesy of [webtoolkit.info](http://www.webtoolkit.info/javascript_base64.html#.YnzNAHXRRhE)

this fixes the issue that the modular list query param encoding breaks if any non ascii chars are used in the query


micahmills (approved):
This works on my end. Good work @squigglybob

PR #1688, opened 2022-05-13
#1647 - Support Address Geocoding On API Entry

### LINKED ISSUE #1647: Option to geocode addresses on API entry.
See import plugin as it has an implementation

kodinkat:
fixes #1647


corsacca:
thank you @kodinkat !

PR #1692, opened 2022-05-23
fix: remove regex that mistakenly filters out folders with dots in

### LINKED ISSUE #1689: phpcs ignoring files with folder names with dots in
in a folder that looks like

`disciple.local/wp-content/themes/disciple-tools-theme`

phpcs finds 0 files

however if disciple.local is changed to disciple-local it works fine.

this could be due to the exclude pattern 

`<exclude-pattern>*\.(?!php$)</exclude-pattern>` catching the dot in the folder name and excluding it because it doesn't end in php after the dot in disciple.local

## Suggested fix

To play around with the include/exclude/extensions options to only scan php files with phpcs

squigglybob:
resolves #1689

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -23,7 +23,6 @@
 
   <file>.</file>
 
-  <exclude-pattern>*\.(?!php$)</exclude-pattern> <!-- only check php files -->
   <exclude-pattern>dt-core/dependencies/*</exclude-pattern>
   <exclude-pattern>dt-core/libraries/*</exclude-pattern>
   <exclude-pattern>vendor/*</exclude-pattern>


squigglybob:
@corsacca I have tested this on php 7.4 and 8.0 and both check the same number of files when running `vendor/bin/phpcs -v`

I haven't checked it with unit_phpcs though, so will do that now as well

squigglybob:
@corsacca ok, should be good to go. This one just fixes the dots in folder names. I'm going to open another issue to add a pre-commit hook to run the phpcs on the staged php files as well.

corsacca:
Thanks @squigglybob !

PR #1697, opened 2022-05-24
#776 - Merge Feature For All Post Types

### LINKED ISSUE #776: Merge feature for all post types
Rewrite the merge tool to work with all post types.
Show all the fields with values so the user can best choose which ones to keep or decide not to merge the records.
We would like the user to be better able to 
- compare both records (all fields)
- chose which value to keep if both records have values (text, key select, date, etc)

kodinkat:
fixes #776


corsacca:
On merging 2 starter posts:

image

Private fields throw errors when merging:

image

corsacca:
Awesome work @kodinkat

corsacca (approved):

PR #1703, opened 2022-06-01
#1691 - Adding User 404 Exception Multisite Support

### LINKED ISSUE #1691: 404 when user try adding users
Screen Shot 2022-05-23 at 11 50 02 AM

Multisite settings needs to be enabled. 
https://wordpress.stackexchange.com/questions/242662/create-users-capabilities-on-a-role-on-multisite#:~:text=It's%20easily%20solved%20by%20going,be%20able%20to%20add%20users

Causes current_user_can( "create_users" ) to return false.

Suggestion:
Add message that says "adding new users needs to be enabled in http://site.com/wp-admin/network/settings.php"

kodinkat:
fixes #1691

### DIFF
--- a/dt-users/template-new-user.php
+++ b/dt-users/template-new-user.php
@@ -3,8 +3,8 @@
  * Name: User Management
 */
 
-// Ensure Add New Users' admin setting has been enabled
-if ( current_user_can( 'manage_dt' ) && ! current_user_can( 'create_users' ) ) {
+// Ensure Add New Users' admin setting has been enabled on multi-sites
+if ( is_multisite() && ! get_site_option( 'add_new_users' ) ) {
     get_header();
     ?>
     <div class="large-6 medium-6 small-6"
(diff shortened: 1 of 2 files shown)


dt-users/template-new-user.php
```diff
@@ -3,6 +3,27 @@
  * Name: User Management
 */
 
+// Ensure Add New Users' admin setting has been enabled
+if ( current_user_can( 'manage_dt' ) && ! current_user_can( 'create_users' ) && ! DT_User_Management::non_admins_can_make_users() ) {
```
corsacca, line 7: @kodinkat This is not the same setting that is change in the network section

kodinkat: Reverted back to `create_users` capability check only; which can be tested by toggling Add New Users network setting.

corsacca: @kodinkat Let's still replace `DT_User_Management::non_admins_can_make_users()` with a check for the network setting

PR #1706, opened 2022-06-07
#1685 - Support Tile Title Translations

### LINKED ISSUE #1685: Tile Title translations to showing in tiles
When adding a custom translation for a language to a tile title, this change is not showing up in the tile title

kodinkat:
fixes #1685

### DIFF
--- a/dt-posts/dt-posts.php
+++ b/dt-posts/dt-posts.php
@@ -1594,7 +1594,7 @@ class DT_Posts extends Disciple_Tools_Posts {
         return $columns;
     }
 
-    private static function adjust_post_tile_labels( $tiles ): array {
+    private static function translation_post_tile_labels( $tiles ): array {
         $adjusted_titles = [];
         foreach ( $tiles as $tile_key => $tile ) {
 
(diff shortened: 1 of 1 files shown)


dt-posts/dt-posts.php
```diff
@@ -1594,12 +1594,26 @@ public static function get_default_list_column_order( $post_type ){
...
+
+        return $adjusted_titles;
+    }
 
     public static function get_post_tiles( $post_type, $return_cache = true ){
         $cached = wp_cache_get( $post_type . "_tile_options" );
-        if ( $return_cache && $cached ){
-            return $cached;
+        if ( $return_cache && $cached ) {
+            return self::adjust_post_tile_labels( $cached );
```
corsacca, line 1616: @kodinkat If we do the translation before the cache is set, then we don't need to call the function in 2 different places.

dt-posts/dt-posts.php
```diff
@@ -1594,12 +1594,26 @@ public static function get_default_list_column_order( $post_type ){
         return $columns;
     }
 
+    private static function adjust_post_tile_labels( $tiles ): array {
```
corsacca, line 1597: @kodinkat, instead of "adjust" can we call this function "translation_post_tile_labels"?

dt-posts/dt-posts.php
kodinkat: Ok, will take a look further upstream so as to translate prior to cache update....

corsacca: @kodinkat just in this function.

corsacca:
Thanks @kodinkat !

PR #1707, opened 2022-06-08
fix: change min max comparisons to parse to int

squigglybob:
@corsacca this should fix the number comparison issue


corsacca:
@squigglybob if it enter in a number below the min, it changes the number to the min, but does not do an api call

squigglybob:
ah yeah, that would be a good point to do an api call... I had been assuming that they were already on the min, and so going below wouldn't do anything.

good catch

squigglybob:
hey @corsacca 

This should be good to go.

I even got the chance to use a nice little closure to prevent the update happening twice 👯‍♂️

PR #1708, opened 2022-06-09
Fix tile label translations

corsacca:
@kodinkat I noticed this function and the a key that was missing:
dt_get_custom_tile_translations()

Thoughts vs https://github.com/DiscipleTools/disciple-tools-theme/pull/1706 ?


kodinkat (approved):
Logic holds true and is a neater approach! Therefore happy for the closure of #1706 PR.

PR #1730, opened 2022-06-29
feat: enhance the global array of languages

squigglybob:
and use in place of long array of languages in dt_get_available_languages


squigglybob:
@prykon

I've fleshed out the global languages array to include native_name, flag, and rtl, but don't have time to put in the details of the flags, native names, rtl info etc.

Would you be able to do a pass over the array and flesh out the missing info, native_name, flag and rtl info?

I've got it to where I need it to be to carry on with the prayer campaigns plugin.

squigglybob:
The `$translations` in the function `dt_get_available_languages` has been removed
image

And is now pulling in the updated list of languages from the `$global_languages_list` via the function `dt_get_global_languages_list`
image

squigglybob:
Here is where the global languages list is first seen, and is the bit that we are working on, so this should still do the same thing as before. ( on the custom lists screen )
image

The list of languages in your user settings page is the one that is now going to be using the global list of languages instead of it's shorter list in the back end. The list you see will still be reduced down to those that have translations in the theme.

I guess we should check that all of the languages that were in that list before are also in the new list.
image

`dt_get_available_translations` is also used in other places as well such as:
any translation cells in the admin side for custom fields, custom lists, custom tiles etc.

prykon:
Global languages list array sent to @squigglybob for review.

squigglybob:
Shall we just output the native language here, rather than the english one as well?
image

PR #1734, opened 2022-07-04
Feat/allow other pages to use translation modal

squigglybob:
Adds a filter to allow other admin pages to use translation modal scripts

Also adds the translation modal html into it's own function to be re-used around the site.


kodinkat (commented):
Nice, as logic still holds true....

Just a thought, could `dt_display_translation_dialog()` be added to a pre-existing admin file?

squigglybob:
I was looking at the other files, and none of them are generic function files, so made one...
Do you have any suggestions of which one it would fit in?

PR #1736, opened 2022-07-18
Adds a post type check on an ajaxComplete event.

micahmills:
This fixes a bug when using the Prayer Request plugin and for proposed updates in the Meetings plugin. When those post types were created from a contact or group record page it caused the name of the record to update in the details summary tile. It wasn't making the change in the DB, just adjusting the DOM. Previously the assumption was that if there was a record update from a record page it was related to the current record, in the case of these plugins this was not the case.


corsacca:
Thanks @micahmills !

PR #1738, opened 2022-07-19
Advanced Search Icon style fix in RTL languages

micahmills:
Advanced Search Icon overlaps the text input in RTL languages. This fixes it


package.json
```diff
@@ -63,7 +63,6 @@
     "@babel/eslint-parser": "^7.18.2",
     "babel-eslint": "^10.1.0",
     "browserslist": "^4.17.0",
-    "caniuse-lite": "^1.0.30001255",
```
corsacca, line 66: @micahmills and this part?

micahmills: Yeah grunt/npm said caniuse was out of date. I reflexively updated it.

corsacca: @micahmills i'm confused. Does the browser check still work with that removed? (on the lists page i think)

micahmills: From what I can see caniuse is only being used by the browserlist which is used by autoprefixer on our CSS as part of a gulp task. check-browser-version.js is using modernizr which is unrelated to this. 

Just to be clear the caniuse-lite NPM package is still being installed it is just now a dependency of the browsers list package.

corsacca:
Thanks @micahmills !

PR #1740, opened 2022-07-21
Workflow action to set current date

cairocoder01:
Adds a checkbox below the datepicker to use current date. When checked, the datepicker is cleared. When a date is selected, the checkbox is unchecked.

Potential issue is that it can cause an infinite loop when it continuously changes a date field because the change of the field would trigger the workflow again. I had to add a condition to my workflow to check if the date field was empty. This could cause unexpected behavior for users who don't know the need for the condition. I tried to change the date to the current date without the time, but it still seems to be having the issue.

### DIFF
--- a/dt-workflows/workflows-execution-handler.php
+++ b/dt-workflows/workflows-execution-handler.php
@@ -478,7 +478,7 @@ class Disciple_Tools_Workflows_Execution_Handler {
                 $updated[ $field_id ] = $value;
                 break;
             case 'date': // $value to be of epoch timestamp int type
-                $updated[ $field_id ] = $value === 'current' ? strtotime(date("Y-m-d", time())) : $value;
+                $updated[ $field_id ] = $value === 'current' ? strtotime( gmdate( "Y-m-d", time() ) ) : $value;
                 break;
         }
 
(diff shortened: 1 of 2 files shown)


dt-workflows/workflows-execution-handler.php
```diff
@@ -473,11 +473,13 @@ private static function action_update( $field_type, $field_id, $value ): array {
             case 'text':
             case 'number':
             case 'boolean':
-            case 'date': // $value to be of epoch timestamp int type
             case 'key_select':
             case 'user_select':
                 $updated[ $field_id ] = $value;
                 break;
+            case 'date': // $value to be of epoch timestamp int type
+                $updated[ $field_id ] = $value === 'current' ? strtotime( gmdate( "Y-m-d", time() ) ) : $value;
```
corsacca, line 481: `strtotime( gmdate( "Y-m-d", time() ) )` can just be `time()`

cairocoder01:
That's what I initially had, which includes the current time. I changed it
to this in order to just use the current date without the time hoping to
fix the looping issue. It made no difference though. I can probably change
it back so that the time is more accurate.

corsacca:
@cairocoder01 awesome I'll wait for the change and then this is good to go!

cairocoder01:
@corsacca Done. I had done it right away but forgot to push I guess.

corsacca:
Thanks @cairocoder01 !

PR #1748, opened 2022-08-01
#1743 - Workflow Actioned Following Conditional Field Trigger

### LINKED ISSUE #1743: Workflow - Action Triggered Only After Conditional Field Update
Only execute actions if conditional fields are updated and evaluate to true; so as to avoid endless loops.

kodinkat:
fixes #1743


corsacca:
Thanks @kodinkat !

PR #1754, opened 2022-08-04
DT Refactor

prykon:
Starting a theme refactor...
The goal is to tidy up code written a while back and make it easier for maintaining and new collaborators to engage with it.


prykon:
@corsacca If you want, we can accumulate a few commits (each commit, a refactored file) and review them one by one.

dt-assets/functions/cleanup.php
```diff
@@ -229,19 +224,16 @@ function dt_disable_wp_emoji() {
...
 
 function dt_disable_emojis_tinymce( $plugins ) {
+    $plugins_array = [];
     if ( is_array( $plugins ) ) {
-        return array_diff( $plugins, array( 'wpemoji' ) );
-    } else {
-        return array();
+        $plugins_array = array_diff( $plugins, array( 'wpemoji' ) );
     }
+    return $plugins_array;
```
corsacca, line 231: @prykon this is simpler
```
    if ( is_array( $plugins ) ) {
        return array_diff( $plugins, array( 'wpemoji' ) );
    }
    return [];
```

prykon:
@corsacca dt-assets/functions/cleanup.php has comments on the filters inside the `dt_start()` function but then redundant comments on each of the callback functions.

**Example:**
Screen Shot 2022-08-05 at 12 10 01

Screen Shot 2022-08-05 at 12 10 33

**Should we remove...?**
- The comment on the actions/filters
- The comment on the callback function
- Both
- Neither

Or should we change the callback function names for descriptive ones that render the comments irrelevant?

dt-assets/functions/cleanup.php
corsacca: @prykon why keep the redundant variable?

corsacca:
@prykon i'd leave both.

dt-assets/functions/cleanup.php
prykon: @corsacca Updated.

corsacca:
thank @prykon

PR #1755, opened 2022-08-05
#1752 - List Filters Typeahead Find Field

### LINKED ISSUE #1752: List Filters: Typeahead to find the field
image

Change the "filter name" to be a typeahead used to search the fields list.
Setting the filter name here is not needed.

Search the field name and key for matches.

kodinkat:
fixes #1752


archive-template.php
```diff
@@ -773,31 +773,29 @@ class="dt_date_picker" data-delimit="end"
 
         <div class="grid-x grid-padding-x">
             <div class="cell small-4 filter-modal-left">
-                <button class="button button-cancel clear" data-close aria-label="Close reveal" type="button">
+                <button class="button button-cancel" data-close aria-label="Close reveal" type="button" style="margin: 5px 0 5px 0;">
```
corsacca, line 776: @kodinkat can we pull these into scss instead?

PR #1757, opened 2022-08-08
#373 - Default Workflow To Auto Select In Church/Group Following Group Addition

### LINKED ISSUE #373: Adding a contact to a group could trigger automatic ticking of “In Church/Group” under progress
<a href="https://github.com/burntsun"></img></a> **Issue by [burntsun](https://github.com/burntsun)**
_Monday Nov 20, 2017 at 14:06 GMT_
_Originally opened as https://github.com/DiscipleTools/disciple-tools/issues/293_

----

> If a contact is listed in a group (under Connections) would it be possible to automatically tick the “In Church/Group” button under Progress?

This is an interesting idea that is worth considering.

kodinkat:
fixes #373

### DIFF
--- a/dt-contacts/base-setup.php
+++ b/dt-contacts/base-setup.php
@@ -386,7 +386,7 @@ class DT_Contacts_Base {
     public static function dt_record_admin_actions( $post_type, $post_id ){
         if ( $post_type === "contacts" ){
             $post = DT_Posts::get_post( $post_type, $post_id );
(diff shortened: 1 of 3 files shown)


corsacca:
@kodinkat I think this should also work when a contact is added to a group from the group record.

corsacca:
@kodinkat is this now 2 separate workflows?
Can it part of the same workflow under the contacts post type?

PR #1759, opened 2022-08-10
Fix error on strategist profile

corsacca:


prykon (approved):
This solves the issue.

PR #1767, opened 2022-08-22
#1676 - Only Show Published Post Related Activities

### LINKED ISSUE #1676: Activity Log shows trashed items
As seen in the attached image, the `Activity Log` is currently displaying results from records from updated fields that were updated when the records were sent to the trash.

Screen Shot 2022-05-01 at 17 11 41

kodinkat:
fixes #1676


corsacca:
Thanks @kodinkat !

PR #1770, opened 2022-08-24
Usage: Nomination

ChrisChasm:
Usage: Nomination for collection country codes for contacts, groups, churches.
Adds country codes so that DT can have a better idea for usage at a country level.


corsacca:
@ChrisChasm do we need to separate contacts, groups and churches? Could lump them together to be more specific

ChrisChasm:
@corsacca We could do that. The regions function in this class was attempting to do that, but most people do not set the region of focus, so this data isn't useful. 

We would be able to understand the different between contact and group tracking of the system. Differentiating the group and church might not be as valuable. 

I did like your nomination of surfacing the "opt out" to a clickable configuration in the admin area. That makes sense regardless.

corsacca:
Todo:
- [ ] Setting to opt out.

ChrisChasm:
Added opt-out metabox in the security tab
Added languages


Screen Shot 2022-09-21 at 12 09 16 PM

Screen Shot 2022-09-21 at 12 09 27 PM

PR #1772, opened 2022-08-26
Mapping v2b

ChrisChasm:
Part 2 of 3
- Replaces the Levels tab in the Mapping admin area with a full location grid install
- Removed the country by country location install for this simpler method. 
- Changed the asset location of the theme data to a google cloud bucket instead of Github.


corsacca:
@ChrisChasm Why do all of them instead of just country by country?

PR #1793, opened 2022-09-13
Fix/save sort order of list

### LINKED ISSUE #1680: Save the sort order of the modular list between refreshes
It would be nice to have the sort order of the modular list stay consistent between refreshes of the page, or revisits to a list of records.

Also since some recent changes to the modular list, the sort order has changed to sort by creation date descending

squigglybob:
resolve #1680


corsacca:
Thank you @squigglybob !

PR #1796, opened 2022-09-15
feat: link field upgrades

### LINKED ISSUE #1775: Link type upgrades
Suggestions for maybe separate Issue/PR

- Add button could be a dropdown with the types. 

- If there are no types should there be a message saying to create one?

- Auto add if there is only one selection

- Combine inputs under one label:
image

_Originally posted by @corsacca in https://github.com/DiscipleTools/disciple-tools-theme/issues/1687#issuecomment-1230640792_

squigglybob:
resolves #1775


squigglybob:
@corsacca Hey there. This update to the link type should be done I think. I haven't touched the issue of adding in a warning if there are no link types available.

I'm thinking that as we are now creating a default link type when a link field is created, hopefully this situation won't happen ( although it could if they deleted their last link type )

corsacca:
@squigglybob 
Dropdow opening outside of screen :(
image

It might be helpful to use the foundation menu dropdown

PR #1804, opened 2022-09-21
Bulk send for group post types

ChrisChasm:
Added a filter and extra loop for bulk send to be enabled on group style post types. For example, the groups or streams post types can have a field ('reporter') added as a connection type to contacts. This would allow bulk sending of magic links to all the contacts connected through this connection field. 

The default field name for this connection type is 'reporter' but can be changed using the filter, and can be extended to support different plugins that are offering magic links. 

The use case for this is to solve distribution of magic links to group oriented post types. 

Next steps:
Add support for Groups
Add support in the admin area for fields to be enabled as distribution connection fields, instead of just 'reporter', i.e. subassigned, or coaches, etc.


dt-reports/magic-url-endpoints.php
```diff
@@ -121,7 +132,24 @@ public function email_magic( WP_REST_Request $request ) {
             }
 
             // check if email exists to send to
-            if ( ! isset( $post_record['contact_email'][0] ) && empty( $params['email'] ) ) {
+            $emails = [];
+            if ( isset( $params['email'] ) && ! empty( $params['email'] ) ){
```
ChrisChasm, line 136: If email is provided, go with the email.

dt-reports/magic-url-endpoints.php
```diff
@@ -121,7 +132,24 @@ public function email_magic( WP_REST_Request $request ) {
...
+                $emails[]   = $params['email'];
+            }
+            else if ( isset( $post_record['contact_email'][0] ) ) {
```
ChrisChasm, line 139: If the single field for a contact "contact_email" field is available, go with the email found here. This will serve contacts.

dt-reports/magic-url-endpoints.php
```diff
@@ -121,7 +132,24 @@ public function email_magic( WP_REST_Request $request ) {
...
+                $emails[]  = $post_record['contact_email'][0]['value'];
+            }
+            else if ( isset( $post_record[$target_field] ) && ! empty( $post_record[$target_field] ) ) {
```
ChrisChasm, line 142: Else if the target connection field (default 'reporter') exists, then loop through and lookup available emails found in these contacts and send links to these individuals.

dt-reports/magic-url-endpoints.php
```diff
@@ -113,6 +113,17 @@ public function email_magic( WP_REST_Request $request ) {
...
+        /**
+         * Bulk Field Filter
+         *
+         * This filter allows for targeting a specific contact connection field to loop through and send emails.
+         * The default field name is 'reporter', but this can be overwritten and pointed towards assigned_to, subassigned_to,
+         * coaching, or another custom contact connection field.
+         *
+         * @param (string)
+         */
+        $target_field = apply_filters( 'dt_bulk_email_connection_field', 'reporter' );
```
corsacca, line 125: @ChrisChasm the target_field will depend on the post type. Maybe we can pass that in.
Each magic link might want to specify their own connection field. can we add which magic link is being used as a param too?

PR #1813, opened 2022-10-03
#1718 - Show/Hide Tiles Based On Field Selection 

### LINKED ISSUE #1718: Show or hide tiles based on field selection
Allow a tile to be lined to a `key_select` field or a `multi_select` field.
if a certain option is selected, then show the tile.

kodinkat:
fixes #1718


corsacca:
@kodinkat lets do the multi option version, and make `$tile['displayed']` an array instead of a string

corsacca:
@kodinkat 
We already have the "hidden" parameter. So always and never are redundant.
Maybe we can move the hidden parameter down next to this setting. And make Visible, hidden, custom a radio.
If custom is chosen, then it shows the fields picking wizard. And hidden is set to false.

lets rename `$tile['displayed']` to  $tile['display_conditions']  and make it look like this:
```php
[   
  'operator' => 'or' (could be 'and' if "Displayed if all specified fields are present")
  'conditions' => [ 
    [ 'key' => 'tags', 'value' => 'bob' ],
    [ 'key' => 'status', 'value' => 'active' ],
  ]
]
```
This lets us expand with other parameters down the line if we want.

corsacca:
There is also an error showing when opening the settings for a new tile

corsacca:
This is looking good @kodinkat 

image
image
image

The hidden selection should still use the `['tile]['hidden] = true` option to not break existing setups.
My "other" tile has `"hidden" = true` but still displays

Add: Select field and field options. The tile will be displayed when these are selected on the record.

Rename "Displayed if all specified fields are present" -> "Require all selected condition to be true"

Help Modal
Lets add a ? by `The display` that opens a modal with simple instructions explaining
- hidden
- visible
- custom
  - how to create select the option
  - what "Require all selected condition to be true" means

corsacca:
@kodinkat 
this point seems to not be true (yet)
The hidden selection should still use the ['tile]['hidden] = true option to not break existing setups.

corsacca:
Thanks @kodinkat !

PR #1818, opened 2022-10-11
Disappearing More Menu Items Fix

kodinkat:


corsacca:
@kodinkat can you describe the changes proposed here?
It looks like Chris pushed a fix for the current bug

kodinkat:
> @kodinkat can you describe the changes proposed here?
> It looks like Chris pushed a fix for the current bug

Hey @corsacca - sure thing; fleshed out the js logic a little; as an alternative to the css approach. Also, ensured the `peoplegroups` menu display flag is observed, whilst generating the more dropdown list items.

Happy to close PR, if no longer required...?

PR #1830, opened 2022-10-21
#1795 - Ability To Configure Comment Types

### LINKED ISSUE #1795: Ability to configure comment types
Add custom items to this list:
image

kodinkat:
fixes #1795

### DIFF
--- a/dt-contacts/base-setup.php
+++ b/dt-contacts/base-setup.php
@@ -649,7 +649,7 @@ class DT_Contacts_Base {
         if ( $post_type === 'contacts' ){
             $channels = DT_Posts::get_post_field_settings( $post_type );
             foreach ( $channels as $channel_key => $channel_option ) {
-                if ( $channel_option['type'] !== 'communication_channel' ){
+                if ( ( ! isset( $channel_option['is_channel'] ) && ( $channel_option['type'] !== 'communication_channel' ) ) || ( isset( $channel_option['is_channel'] ) && ! $channel_option['is_channel'] ) ) {
                     continue;
                 }
                 $enabled = !isset( $channel_option['enabled'] ) || $channel_option['enabled'] !== false;
--- a/dt-core/admin/menu/tabs/tab-custom-lists.php
+++ b/dt-core/admin/menu/tabs/tab-custom-lists.php
@@ -261,17 +261,18 @@ class Disciple_Tools_Tab_Custom_Lists extends Disciple_Tools_Abstract_Menu_Base
                 </thead>
                 <tbody>
                     <?php foreach ( $fields as $channel_key => $channel_option ) :
-                        if ( $channel_option['type'] !== 'communication_channel' ){
+                        if ( ( ! isset( $channel_option['is_channel'] ) && ( $channel_option['type'] !== 'communication_channel' ) ) || ( isset( $channel_option['is_channel'] ) && ! $channel_option['is_channel'] ) ) {
                             continue;
                         }
 
                         $enabled = !isset( $channel_option['enabled'] ) || $channel_option['enabled'] !== false;
                         $hide_domain = isset( $channel_option['hide_domain'] ) && $channel_option['hide_domain'] == true;
-                        if ( $channel_key == 'phone' || $channel_key == 'email' || $channel_key == 'address' ){
+                        if ( $channel_key == 'contact_phone' || $channel_key == 'contact_email' || $channel_key == 'contact_address' ){
                             continue;
                         } ?>
 
                     <tr>
+                        <input type="hidden" name="channel_fields[<?php echo esc_html( $channel_key ) ?>]">
                         <td><input type="text" name="channel_label[<?php echo esc_html( $channel_key ) ?>][default]" value="<?php echo esc_html( $channel_option['name'] ?? $channel_key ) ?>"></td>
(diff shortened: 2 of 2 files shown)


corsacca:
Screenshot
image

The goal isn't to add more communication channels.

The comment types are a combination of:
- the existing communication channels
- plugin added types via the `dt_comments_additional_sections` filter

What we would like is a new table, that list the combination of those 2 and lets the user:
- add new comment types
- modify labels and translations for existing comment types.

We should consider excluding the communication channels from this new table since they are already editable.

corsacca:
@kodinkat  i'm not seeing the comment types added fia `dt_comments_additional_sections`

kodinkat:
> @kodinkat i'm not seeing the comment types added fia `dt_comments_additional_sections`

Stored alongside other custom settings within `dt_field_customizations` option and loaded within `add_comm_channel_comment_section()` function; which is assigned to `dt_comments_additional_sections` filter.

Are any custom types being shown within Comments & Activity section?

Screenshot 2022-11-29 at 16 35 15

corsacca:
The Label column is picking up the translation instead of the default label:
image

Don't allow disabling the non custom types
image
A record will always have activity.

corsacca:
The comment type order got changed, which means the default comment type when adding a comment on a contact is no longer "comment"
image

dt-core/admin/menu/tabs/tab-custom-lists.php
```diff
@@ -261,17 +440,18 @@ public function channels_box(){
...
                     <?php foreach ( $fields as $channel_key => $channel_option ) :
-                        if ( $channel_option['type'] !== 'communication_channel' ){
+                        if ( $channel_option['type'] !== 'communication_channel' ) {
                             continue;
                         }
 
                         $enabled = !isset( $channel_option['enabled'] ) || $channel_option['enabled'] !== false;
                         $hide_domain = isset( $channel_option['hide_domain'] ) && $channel_option['hide_domain'] == true;
-                        if ( $channel_key == 'phone' || $channel_key == 'email' || $channel_key == 'address' ){
+                        if ( $channel_key == 'contact_phone' || $channel_key == 'contact_email' || $channel_key == 'contact_address' ){
```
corsacca, line 449: This removes them from the Communication Channels list?

dt-core/admin/menu/tabs/tab-custom-lists.php
```diff
@@ -241,6 +248,178 @@ public function process_user_profile_box() {
...
+
+                    // Persist key prefix.
+                    $custom_contact_fields[$type]['key_prefix'] = $custom_key_prefix;
+                }
+            }
+
+            // Handle the adding of new comment types.
+            if ( ! empty( $_POST['add_type'] ) ) {
+                $label = sanitize_text_field( wp_unslash( $_POST['add_type'] ) );
+                $key   = mb_strimwidth( dt_create_field_key( $custom_key_prefix . $label ), 0, 20 );
```
corsacca, line 312: We don't use mb_ functions as that extension might not be installed.

dt-contacts/base-setup.php
```diff
@@ -661,10 +661,95 @@ public function add_comm_channel_comment_section( $sections, $post_type ){
                     'label' => esc_html( $channel_option['name'] ?? $channel_key )
                 ];
             }
+
+            // Extract custom comment types.
+            $custom_field_options  = dt_get_option( 'dt_field_customizations' );
```
corsacca, line 666: @kodinkat I think this would be simpler if it wasn't save in the custom fields array. A comment type isn't a custom field.
Lets move it to it's own option that soly stores custom comments types. Then merge those options  over the filtered ones.

dt-core/admin/menu/tabs/tab-custom-lists.php
```diff
@@ -241,6 +248,178 @@ public function process_user_profile_box() {
...
+                        if ( isset( $_POST['type_labels'][$type][$langcode] ) ){
+                            $translated_label = sanitize_text_field( wp_unslash( $_POST['type_labels'][$type][$langcode] ) );
+                            if ( ( empty( $translated_label ) && !empty( $custom_contact_fields[$type]['translations'][$langcode] ) ) || !empty( $translated_label ) ){
+                                $custom_contact_fields[$type]['translations'][$langcode] = $translated_label;
+                            }
+                        }
+                    }
+
+                    // Persist key prefix.
+                    $custom_contact_fields[$type]['key_prefix'] = $custom_key_prefix;
```
corsacca, line 305: Getting a key_prefix undefined on first load.

dt-core/admin/menu/tabs/tab-custom-lists.php
kodinkat: I believe this was previously the case, although; the key reference was wrong. Are phone/email/address to be included by default?

corsacca:
@kodinkat can we remove the `activity` entry? The user does not need to be able to select it.
image

kodinkat:
> @kodinkat can we remove the `activity` entry? The user does not need to be able to select it.
> image

Yep, that's doable - I'll circle back to this one! :smile:

corsacca:
@kodinkat here too
image

corsacca:
Thank you @kodinkat

PR #1833, opened 2022-10-24
#1826 - Updating Comment With No Change Throws Error

### LINKED ISSUE #1826: updating a comment with no change throws an error
Create a comment.
Click on edit button to edit comment
Click on update and throws error like in the image

image

kodinkat:
fixes #1826


corsacca:
@kodinkat if the wp_update_comment returns 0, then it means the update failed. Why is it failing?

PR #1846, opened 2022-10-31
Change the request referrer header to `strict-origin-when-cross-origin` which is the new default since 2022

corsacca:
Mapbox needs the referrer header in order to work when restricting the key to only work on certain urls.
See:
https://docs.mapbox.com/mapbox-gl-js/guides/browsers-and-testing/#referrer-policies

The current policy is `same-origin` which means any outside link a D.T user clicks on to open does not include the referrer header.
image

For mapbox api token restriction to work we would need to change the referrer-policy to `strict-origin-when-cross-origin`
image

### DIFF
--- a/dt-core/configuration/restrict-site-access.php
+++ b/dt-core/configuration/restrict-site-access.php
@@ -209,7 +209,7 @@ function dt_security_headers_insert() {
         header( 'X-XSS-Protection: 1; mode=block' );
     }
     if ( !$referer_disabled ){
-        header( 'Referrer-Policy: strict-origin-when-cross-origin' );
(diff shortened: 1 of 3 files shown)


micahmills:
This will fix the mapbox problem, my concern is that it will also leak the DT instance domain in the referrer to any outside links. I don’t fully know where all the mapbox domains are being called but is it possible to keep the header same origin and only change the header for the mapbox links in a per request basis https://javascript.info/fetch-api#referrer-referrerpolicy

corsacca:
@micahmills thanks for the suggestion

PR #1848, opened 2022-11-02
#1836 & #1816 - Workflow Updates

### LINKED ISSUE #1816: Workflows: Ability to remove
Remove tags, connections, etc
Unset fields like dates, etc

### LINKED ISSUE #1836: Workfows: Implement Mapbox locations

kodinkat:
fixes #1836 
fixes #1816

### DIFF
--- a/dt-core/admin/js/dt-utilities-workflows.js
+++ b/dt-core/admin/js/dt-utilities-workflows.js
@@ -1204,6 +1204,12 @@ jQuery(function ($) {
           config['order'] = 'asc';
         }
 
+        // Determine search on focus setting.
+        if (typeahead['general']) {
+          config['searchOnFocus'] = typeahead['general']['search_on_focus'];
+          config['minLength'] = 2;
+        }
+
         // Instantiate...!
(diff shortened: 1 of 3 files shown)


corsacca:
On a record:
image
2 words when choosing a location in the workflows is not working.
image

corsacca:
@kodinkat 
When 2 fields are present, it won't remove the faith milestone if the Coach of group is not set.
image

corsacca:
@kodinkat 
Why is the best result way at the bottom?
image

should be at the top:
image

kodinkat:
> @kodinkat Why is the best result way at the bottom? image
> 
> should be at the top: image

@corsacca Ok, so; by default typeahead appears to display results accordingly, by query hit relevance. However, within core, the relevance flag seems to be ignored; as the results list is manually built & displayed.

corsacca:
@kodinkat can we mimic the core here too?

kodinkat:
> @kodinkat can we mimic the core here too?

I'll take a look to see how best to work this in....

PR #1859, opened 2022-11-14
Dup records check on create

corsacca:


kodinkat (commented):
@corsacca dups logic holds true for me whilst submitting new records via webform. If a phone or email match is found, then matched record is updated. However, no match, then a new contacts record is created - see attached screenshots.

Quick sanity check - I'm thinking it's just my instance, as I've noticed both `contact_phone` and `contact_email` comms_channel fields are no longer shown within contacts record? Are you experiencing this also...?

Screenshot 2022-11-15 at 12 50 11
Screenshot 2022-11-15 at 12 49 27

corsacca:
@kodinkat I don't know where your email and phone fields went. They on vacation?

PR #1897, opened 2022-12-14
Allows custom labels for the Status and Details tile.

micahmills:
Changes the current label to the default_label, if the user changes the custom label it will use that custom label and its translations

### DIFF
--- a/dt-posts/dt-posts.php
+++ b/dt-posts/dt-posts.php
@@ -1807,8 +1807,8 @@ class DT_Posts extends Disciple_Tools_Posts {
         }
         $tile_options = dt_get_option( 'dt_custom_tiles' );
         $default = [
-            'status' => [ 'label' => __( 'Status', 'disciple_tools' ), 'tile_priority' => 10 ],
-            'details' => [ 'label' => __( 'Details', 'disciple_tools' ), 'tile_priority' => 20 ]
+            'status' => [ 'default_label' => __( 'Status', 'disciple_tools' ), 'tile_priority' => 10 ],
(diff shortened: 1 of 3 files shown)


dt-posts/dt-posts.php
```diff
@@ -1807,8 +1807,8 @@ public static function get_post_tiles( $post_type, $return_cache = true ){
         }
         $tile_options = dt_get_option( 'dt_custom_tiles' );
         $default = [
-            'status' => [ 'label' => __( 'Status', 'disciple_tools' ), 'tile_priority' => 10 ],
-            'details' => [ 'label' => __( 'Details', 'disciple_tools' ), 'tile_priority' => 20 ]
+            'status' => [ 'default_label' => __( 'Status', 'disciple_tools' ), 'tile_priority' => 10 ],
```
corsacca, line 1810: @micahmills Let's keep the default to be just `label` then the mobile will still know where to get the name of the tile.

single-template.php
```diff
@@ -126,7 +126,11 @@ function dt_display_tile( $tile, $post ): bool {
                     <?php if ( isset( $tiles['status'] ) && empty( $tiles['status']['hidden'] ) ) : ?>
                     <section id="contact-status" class="small-12 cell bordered-box">
                         <h3 class="section-header">
-                            <?php echo esc_html__( 'Status', 'disciple_tools' )?>
+                            <?php if ( isset( $tiles['status']['label'] ) && !empty( $tiles['status']['label'] ) ) {
```
corsacca, line 129: Is there a case currently where a tile does not have the label param set?

PR #1899, opened 2022-12-16
fix: check for selection_id existing before closing list

squigglybob:
fixes console log error when hitting enter in mapbox widget after selecting an item in the list


squigglybob:
@corsacca This is something I came across when working the search box into a modal in PG for setting a user's location.

Using just the keyboard I typed in the box, navigated to a selection with the keys, and hit enter to select it.

Then I hit enter again, and an error showed up as the `selection_id` variable in `close_all_lists` in `mapbox-search-widget.js` was undefined.

corsacca:
Thanks @squigglybob

PR #1901, opened 2022-12-20
feat: add time picker field option and show in front end

### LINKED ISSUE #1895: Date Time field type
It's time for a new field type :)

## The problem

DT currently has a date field type, but it can only set the date, not the time.
Prayer Global laps care about the time as well as the date, but if the date is changed in the DT backend using the datepicker it will set the time to be midnight of that date, thus overwriting any time aspect of the timestamp that was originally there.

For custom laps that say want to start at 11am on a Sunday morning this is an issue if they are subsequently edited in DT.

## Implementations in plugins

The trainings plugin implements a DateTime field type, but this is not currently in core.

The Prayer.Global porch needs a DateTime field for the lap start/end datetime.

Zume 5.0 is going to need a DateTime field as well.

## Proposal

I propose adding in a new field type using the html input type `<input type="datetime-local" />`

This will be useful across several projects and hopefully also within the DT theme itself.

## Time estimate

The number field took about 8 hours to implement back when I was still finding my way around.

I have left myself a trail of tagged comments of places in the code that need updating to add a new field type, so this should take significantly less time to implement. It's also hopefully not too much more complicated than the number field type

I estimate it to take 1 or 2 days at most.

@corsacca @ahillbilly what say you?

squigglybob:
resolves #1895

### DIFF
--- a/dt-core/admin/menu/tabs/tab-custom-fields.php
+++ b/dt-core/admin/menu/tabs/tab-custom-fields.php
@@ -498,6 +498,30 @@ class Disciple_Tools_Tab_Custom_Fields extends Disciple_Tools_Abstract_Menu_Base
 
         <?php endif; ?>
 
+        <?php if ( $field['type'] === 'date' ) : ?>
+
+            <?php
+                $custom_fields = dt_get_option( 'dt_field_customizations' );
+                $custom_field = $custom_fields[$post_type][$field_key] ?? [];
+            ?>
+
+            <h3><?php esc_html_e( 'Field Options', 'disciple_tools' ) ?></h3>
+            <table id="date_options">
+                <tr>
+                    <td style="vertical-align: middle">
+                        <?php esc_html_e( 'Enable Time picker', 'disciple_tools' ) ?>
+                    </td>
+                    <td>
+                        <input type="checkbox" name="enable_time_picker" <?php echo isset( $custom_field['enable_time_picker'] ) && $custom_field['enable_time_picker'] === 'on' ? 'checked' : '' ?> />
+                    </td>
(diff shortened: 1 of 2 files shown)


corsacca:
@squigglybob is this still in draft mode?

squigglybob:
yars, I think so. I can't remember why I didn't finish it, but it could probably use some double brains on it to work out the best way forward and what is necessary.

I was working on it, so that the PG laps, would be able to see/change the start time of a lap, but then it dropped lower in priority and I forgot to come back to it I think.

squigglybob:
hey @corsacca finally got this one going 😀

squigglybob:
hahaha, I thought I was done, but then we decided to split date field with optional time picker into
date field
and
datetime field

Let's split these babies up!!!

squigglybob:
@zdmc23 There has been a new field type in the process of being added, so you're aware of it from the mobile side of things

squigglybob:
## Still todo

- [ ] add docs for the datetime field type
- [ ] add tests for the datetime field type
- [ ] test with private fields
- [ ] test with time travel roll back

squigglybob:
Hey @corsacca,

This one is done and tested in the unit tests, and checked with record time travel.

I noticed that the curly brackets are being left around the timestamps in the record.

Is that intentional? Shall I create a new ticket to deal with that or do it in here?

squigglybob:
I removed them 😆 
It should'nt affect any situations where there aren't any brackets either, so should be ok

corsacca:
thanks @squigglybob !

squigglybob:
woop, that was a long time in the coming 😉

PR #1906, opened 2023-01-05
feat: allow use of seperate definition to WP_DEBUG to turn on logs

squigglybob:
I've hit this several times over the last couple of days.

I would like to just turn on the DT logs without turning on the full WP_DEBUG experience which shouldn't be left on in production and can risk users getting strange errors on the front and back end of the WP site.

I propose adding another definition such as DT_WRITE_LOG to allow the system to write dt logs without having to turn on WP_DEBUG.

Obviously this shouldn't be left on either, but it allows you to get a partial debug experience in production if needed without affecting the user

### DIFF
--- a/dt-core/global-functions.php
+++ b/dt-core/global-functions.php
@@ -39,7 +39,11 @@ if ( ! defined( 'DT_FUNCTIONS_READY' ) ){
          * @param $log
          */
         function dt_write_log( $log ) {
-            if ( true === WP_DEBUG ) {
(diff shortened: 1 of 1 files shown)


squigglybob:
This is not needed as you can just make sure that the wp-config.php
contains this incantation to turn on debug without getting errors/warnings on screen

```
define( 'WP_DEBUG', true );
define( 'WP_DEBUG_LOG', true );
define( 'WP_DEBUG_DISPLAY', false );
define( 'SCRIPT_DEBUG', false );
```

Issue #1866, opened 2022-11-22
Integration Request: Make (Integromat)


corsacca:
Make (Integromat) app or integration to easily connect to D.T

PR #1907, opened 2023-01-10
#1866 - Make.com Integration

kodinkat:
- fixes: #1866


kodinkat:
The following enhancements have been made to the existing DT Site Link functionality; to better aid Make.com integration.

- Added long-lived bearer tokens; with adjustable expiry dates.
- Connection type updated to support all available post types, in addition to just `contacts`.

Screenshot 2023-01-10 at 12 21 37

> Still awaiting feedback from Make.com regarding the unlocking of their [Custom IML Functions](https://docs.integromat.com/apps/app-structure/iml-functions) feature; which is better suited for the dynamic integration of DT post types.

Issue #1871, opened 2022-11-29
Icon selector typeahead and overflow


squigglybob:
## Overflow

It shouldn't have an external horizontal and vertical scroll
![image](https://user-images.githubusercontent.com/9816214/204522548-568dd3d0-264a-40b8-a139-0d5220aaa366.png)

The icon selector modal shouldn't have any overflow in the x or y direction.
I imagine this is because the modal is a fixed width, and the internal has a min-width or something

The tooltip on hovering on an icon sometimes increases the vertical overflow as well.

## Typeahead

Pressing any key, ctrl, shift, etc. not just letters, triggers the typeahead and causes the icons to reload.

The typeahead should only respond when something is actually typed.
Maybe the typeahead could have a debounce feature, where it doesn't refresh the icons on every new key press, but waits til the user has finished typing

**To Reproduce**

Go to WP admin --> DT settings --> fields --> edit a field --> click change icon --> click in typeahead --> press shift/ctrl etc.

**Expected behavior**

The icons shouldn't refresh so often.

**Disciple.Tools Theme:**
- version: 1.33.0

squigglybob:
It also shouldn't have a microbreak warning floating over it in a screenshot 😆

PR #1911, opened 2023-01-13
#1871 - Icon Selector Typeahead & Overflow

kodinkat:
- fixes: #1871 

> Unable to re-create floating micro-break warning?


squigglybob:
looks good on my comp 🙂 Nice work

image

There is a wierd gap though in the icons 😆 ... sorry...

PR #1913, opened 2023-01-17
refactor: move erase_data function into new DT_Contacts_Utils class

squigglybob:
moves the erase_data for GDPR purposes out into a utility class.

This will enable other porches/pages etc. (e.g. prayer.global ) to use this erase data GDPR functionality


squigglybob:
@corsacca wasn't sure where you'd be happy to have the function.
Looked at the other options in the dt-contacts folder, but none of the other files made sense as a home for this function

corsacca:
Thanks @squigglybob. I'm not sure the best spot either. So there is good.

Issue #1872, opened 2022-11-29
To view this map install the mapbox key


corsacca:
In the metrics section. 
Some maps are visible when a mapbox key is added. But Otherwise the menu items are hidden.

We want to keep the menu items visible, load the page, but instead of loading the map we'd display a message saying:
"To view this map a mapbox key is needed, click *here* to add"

PR #1914, opened 2023-01-18
#1872 - To View Map Metrics, Install Mapbox Key

kodinkat:
- fixes: #1872

### DIFF
--- a/dt-metrics/combined/mapbox-maps.php
+++ b/dt-metrics/combined/mapbox-maps.php
@@ -59,7 +59,7 @@ class DT_Metrics_Mapbox_Combined_Maps extends DT_Metrics_Chart_Base
                 'translations' => [],
                 'settings' => [
                     'map_key' => DT_Mapbox_API::get_key(),
-                    'no_key_html_msg' => sprintf( __( 'To view this map, a mapbox key is needed; click <a href="%s">here</a> to add.', 'disciple_tools' ), site_url( '/wp-admin/admin.php?page=dt_mapping_module&tab=geocoding' ) ),
+                    'no_map_key_msg' => _x( 'To view this map, a mapbox key is needed; click here to add.', 'install mapbox key to view map', 'disciple_tools' ),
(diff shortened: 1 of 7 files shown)


dt-metrics/combined/mapbox-maps.php
```diff
@@ -62,6 +59,7 @@ public function scripts() {
                 'translations' => [],
                 'settings' => [
                     'map_key' => DT_Mapbox_API::get_key(),
+                    'no_key_html_msg' => sprintf( __( 'To view this map, a mapbox key is needed; click <a href="%s">here</a> to add.', 'disciple_tools' ), site_url( '/wp-admin/admin.php?page=dt_mapping_module&tab=geocoding' ) ),
```
corsacca, line 62: @kodinkat we try to avoid passing html to the translator.
Instead we can pass the whole `<a href="link">__('here', 'dt')</a>` as the var.
And let's use `_x( 'to view...', 'to view ... click here to add', 'disciple_tools' )` so the translator can see what is happening.

corsacca:
This shape looks good.
Let's also add it to the `/user-management/mapbox-map/` map

dt-metrics/combined/mapbox-maps.php
corsacca: @kodinkat that way works too :D

corsacca:
@kodinkat the map menu item doesn't show up whet mapbox not present

kodinkat:
> @kodinkat the map menu item doesn't show up whet mapbox not present on the user-management page

@corsacca I've just sanity checked and I'm seeing the `Coverage Map` menu item, with the prompt to install mapbox key... Is there another view I've missed?

Screenshot 2023-01-20 at 09 44 25

corsacca:
@kodinkat interesting. It now shows up. Pardon my insanity.

kodinkat:
> @kodinkat interesting. It now shows up. Pardon my insanity.

Sweet! Pardon granted!! 😉

PR #1916, opened 2023-01-25
Load Location Typeahead No Element Fix

kodinkat:
See: https://github.com/DiscipleTools/disciple-tools-survey-collection/issues/23#issuecomment-1403585611


corsacca:
Thank you @kodinkat

PR #1917, opened 2023-01-26
feat: include Jwt_Auth library into dt-core libraries

squigglybob:


squigglybob:
forget that, I should have checked the mobile app plugin 😆  It's already well protected against loading the class twice, so all good

Issue #1904, opened 2023-01-04
number field shouldn't change on scroll


squigglybob:
scrolling the page when hovering over the number field changes the number in the field.

Expected behaviour.
Either scroll the number field or the page, Not both.

Fix by preventing change to number field on page scroll

PR #1919, opened 2023-01-27
#1904 - Number Field Should Not Change On Scroll

kodinkat:
- fixes: #1904


corsacca:
@kodinkat can we add something like: https://stackoverflow.com/a/63872227/6640717 to the render method so it is available anywhere the number field is displayed in D.T or in a magic link?

corsacca:
@kodinkat that will add that whole script for each number field?
How about we use?
`<input type="number" onwheel="return false;" />`

PR #1923, opened 2023-02-02
feat: add beta firebase SSO login

### LINKED ISSUE #1922: SSO options in core
# SSO Options in Core

Leaning on Firebase Authentication as a provider for SSO

This will also leverage the new JWT auth library that has just been added to core

## V1 Spec

- [x] add endpoints for login and checking token authentication
- [x] add shortcodes for SSO login UI, and logging out
- [x] add SSO admin area under settings for managing config and options
- [x] allows logging in via normal WP cookie experience
- [x] allow logging in using JWT token
- [x] deal with user logged into WP as admin if using tokens
- [x] saves config at multisite level if a multisite
- [x] saves rest of config at subsite level --> will need a refactor of the update/delete options, and working out saving into potentially two places
- [x] only allow super admins to edit firebase config info
- [x] syncs SSO signon with existing WP users.
- [x] allows various SSO options to be toggled
- [x] allows options of where to redirect
- [x] remove small writing in login shortcode
- [x] create dt user as multiplier

## Testing on DT

- [x] allow using SSO for logging in to admin
- [x] add custom login / register pages
- [x] add terms and conditions and privacy pages
- [x] refactor custom login fields to work with dt_login fields
- [x] allow enabling/disabling of the custom login page
- [x] if firebase email option is checked, don't show email and password option
- [x] use WP/DT option to allow people to register or not.
- [x] by default don't turn on firebase email option
- [x] turn login method into a filter that plugins can activate
- [x] Docs, Docs, Docs, Docs
- [x] double check site settings
- [x] subsites grab main site login options as their defaults
- [x] if no social logins are enabled then don't show firebase login
- [x] change wording of urls, to be paths

## V2 Spec

- [ ] allow users to sign on per sub site in multisite
- [ ] allow sign-in options to be reordered
- [ ] add option to disallow auto-registering
- [ ] use email address as display name if they don't have a display name when using SSO
- [ ] is there a way to sign into firebase in the admin area and it get the info it needs directly from the firebase console?
- [ ] add option to login with SSO when creating new password as part of user invite process
- [ ] add option to choose default role when user registers themselves
- [ ] allow redirect param in custom login page
- [ ] if user is using firebase email signon then site needs to handle the case that they haven't verified their email yet
- [ ] allow network admin to network disable identity provider options

## RoadMap

```mermaid
flowchart LR
a(BETA in core):::red --> A(Test in PG)
a --> B(Test in DT)
a --> C(Test in Porches)
a --> D(Test in Zume5)
A:::orange --> b(Release in core):::green
B:::orange --> b
C:::orange --> b
D:::orange --> b

classDef red fill:#D00, color:#000, border:unset
classDef orange fill:#DA0, color:#000, border:unset
classDef green fill:#0D0, color:#000, border:unset
```

squigglybob:
resolves #1922


squigglybob:
this is ready for testing properly on staging

squigglybob:
- [ ] It would be nice to set all of the settings at a site wide level and have them only enabled for super admins
  - [ ] as part of this, debug why 
- [ ] clarify that the login and redirect urls are just paths, and give an example

Issue #1934, opened 2023-02-08
Group Member List Ordering


kodinkat:
Alphabetically sort listed group members by Name:

![Screenshot 2023-02-08 at 11 00 35](https://user-images.githubusercontent.com/19330800/217511423-55e6ebc6-e90e-44a4-aa35-0cbd885a78c2.png)

PR #1939, opened 2023-02-10
#1934 - Group Member List Ordering

kodinkat:
- fixes: #1934

### DIFF
--- a/dt-groups/groups.js
+++ b/dt-groups/groups.js
@@ -143,7 +143,7 @@ jQuery(document).ready(function($) {
         m.leader = true
       }
     })
-    post.members = window.lodash.sortBy(post.members, ["leader", "post_title"])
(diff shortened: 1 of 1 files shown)


dt-groups/groups.js
```diff
@@ -143,7 +143,7 @@ jQuery(document).ready(function($) {
         m.leader = true
       }
     })
-    post.members = window.lodash.sortBy( post.members, ["leader"])
+    post.members = window.lodash.sortBy( post.members, ["post_title"])
```
corsacca, line 146: @kodinkat can it do leader, and then member name?

corsacca:
I'm not sure the order is always working:
image

kodinkat:
> I'm not sure the order is always working:
> image

@corsacca - Yep, it seems to be a caps thing; as uppercase will be sorted before lowercase. I'll explore the API further to try and find a suitable option/workaround; which ignores casing..

corsacca:
Thanks @kodinkat !

Issue #1874, opened 2022-11-29
Ability to copy tiles and fields settings to another instance.


corsacca:
Also:
Ability to set tiles and fields globally for a whole multisite

PR #1946, opened 2023-02-17
#1874 - Export/Import D.T Settings

kodinkat:
fixes: #1874
---

## Summary

The Admin Utilities section has now been extended, to support the following new features:
- D.T Settings Exports
- D.T Settings Imports

### Filter & Action Hooks

The following filter/action hooks have also been introduced:
- __dt_export_services__
  - A filter which returns a list of export services; which, by default, are Tile & Field Settings.
  - External plugins and other D.T settings can also be hooked in, by following the implemented framework pattern.
- __dt_export_payload__
  - Filter event which carries out the actual settings extraction, to be included and stored within downloaded json file.
- __dt_import_services__
  - A filter which returns a list of import services; which, by default, are Tile & Field Settings.
  - Like above, additional services can be integrated.
- __dt_import_services_details__
  - This filter provides import services, with the ability to display additional import-configuration details; enabling extra control over the import process.
- __dt_import_payload__
  - Action hook; which handles the actual importing of uploaded configuration json data.

### Exports Tab

- Provides the option of cherry-picking services to be exported.
- Exported data is automatically downloaded into a json file, ready for transportation and uploading into the target D.T instance.

Screenshot 2023-02-28 at 16 50 42

### Imports Tab

- Provides the ability to upload D.T json configuration files; which have been exported from other D.T instances.
- Once configuration data has been uploaded, the ability to cherry-pick actual options at a granular level, is provided.
- Setting options already existing within the target D.T instance, are disabled.

Screenshot 2023-02-28 at 16 51 26

Screenshot 2023-02-28 at 16 52 49

Screenshot 2023-02-28 at 16 54 10

@corsacca once merged, I'm happy to update documentation, so as to capture new export/import feature.


corsacca:
image
Selecting Both does not give me the options to selected tiles AND fields as the same time?
image

corsacca:
Could be one tab, since imports only has a small section?

corsacca:
image
Make it a radio import all and select fields.
Import all can open the selections underneith all check
Import custom can open then unchecked.

Maybe tiles and fields each has it's own box underneith that can be expanded and collapsed

kodinkat:
> image
> Selecting Both does not give me the options to selected tiles AND fields as the same time?
> image

@corsacca 

> Missing instructions....

Cool, I'll work in some wording...

> Select all buttons.

Will work these in across both details and general services sections.

> Relocate service details section.

Could do, however, given the framework nature of this feature, I envisage the main services section list growing, as more services (plugins, workflows, magic-links, etc) come online; who wish to make use of the export/import functionality. Based on that use case, you would need to always scroll down to the bottom to view details, instead of it appearing immediately to the side - Happy to discuss this one further, once other UI changes are in..

> Selecting Both does not give me the options to selected tiles AND fields as the same time?

Clicking on the service title displays the corresponding details section; which then allows you to select further.

kodinkat:
> Could be one tab, since imports only has a small section?

@corsacca 

Again, given the framework nature of this feature, the services list section has scope to grow. Also, as the general flow is dependant on page refreshes, it would be cleaner and easier to follow logic if both exports and imports are kept within their respective swimlanes (tabs).

kodinkat:
> image Make it a radio import all and select fields. Import all can open the selections underneith all check Import custom can open then unchecked.
> 
> Maybe tiles and fields each has it's own box underneith that can be expanded and collapsed

@corsacca 

I believe the other UI changes may satisfy these requirements. Again, given the framework nature of the export/import feature; some sections would need to remain uniform, to minimise any friction for other services wanting to integrate with export/import service.

Again, I believe the next round of UI changes will satisfy some of these requirements; but, happy to discuss further once in..

corsacca:
Option: We could make the columns take up 50% of the space.

corsacca:
@kodinkat When enabling the overwrite, uncheck the "select all" so it can be checked again.
image

corsacca:
I think we need to work on the workflow a bit.
I think by default the User would expect all settings to be imported for each service.
As it is is now the user would need to carefully look at the list, figure out to enable overwrite, make sure all fields are checked. And then repeat for each service.

PR #1947, opened 2023-02-19
Fix translations

mikeallbutt:


corsacca:
Thanks @mikeallbutt .
`esc_html_e` means echo, escape and translate. So `echo` is not needed.

PR #1952, opened 2023-02-23
use `list_all_` permission for viewing compact list

cairocoder01:
The `list_all_[post_type]` permission was not properly being used at all and was mis-named in the code. This updates `get_viewable_compact` to check the `list_all_` permission and then skip further permission checks in `search_viewable_post`


corsacca:
Thank you @cairocoder01 !

PR #1954, opened 2023-02-24
Revised Initial User Language Setup Flow

kodinkat:


dt-users/user-initial-setup.php
```diff
@@ -96,6 +91,29 @@ function dt_user_initial_setup_modal(): void{
...
+
+                // Post updated language locale.
+                makeRequest("POST", `user/update`, {
+                    'locale': jQuery('#locale').val()
+                }, 'dt/v1/')
+                .done(response => {
+
+                    // Assume all is well in the world, refresh page and close modal!
+                    modal.foundation('close');
+                    location.reload();
```
corsacca, line 109: @kodinkat is the reload necessary if the lang did not change?

Issue #1953, opened 2023-02-23
Search compact unit tests


corsacca:
Add unit tests for the typeahead search that test that the search does not return more record that the user has access to.

PR #1961, opened 2023-03-01
#1953 - Viewable Compact Search Unit Tests

kodinkat:
- fixes: #1953


corsacca:
@kodinkat, 
Let me know if this is already in place:
A different user creates a contact with name "John". Current user can not access it.
A search for "John Doe" returns the expected contacts (Including John Bob Joe).

kodinkat:
> @kodinkat,
> Let me know if this is already in place:
> A different user creates a contact with name "John". Current user can not access it.
> A search for "John Doe" returns the expected contacts (Including John Bob Joe).

@corsacca 

Hmm, I think these may have been missed... I'll circle back and double-check.....

corsacca:
Pushed a commit to fix the typeahead
image

image

PR #1972, opened 2023-03-07
feat: add param to dt_get_url_path to include host

squigglybob:
easily leave host in the path when using this function

@corsacca let me know what you think

I'm happy to copy and paste a version of this into PG in necessary


corsacca:
Looks good

Issue #1955, opened 2023-02-28
Click on multiselect to create filter


corsacca:
The typeahead multiselect like the "Sources" field.
Let's make those clickable.

PR #2000, opened 2023-03-22
#1955 - Enabled Clickable Typeahead Multi Select Filter Queries

kodinkat:
- fixes: #1955


corsacca:
@kodinkat see last commit to get field name

PR #2011, opened 2023-03-28
Fixed typeahead bug

prykon:


dt-core/admin/admin-settings-endpoints.php
```diff
@@ -196,7 +196,6 @@ public static function get_post_fields() {
                 }
             }
         }
-        header( 'Content-Type" => application/json' );
         echo json_encode( $output );
```
corsacca, line 199: this would probably be better as just a return $output

PR #2021, opened 2023-03-30
check permission to install or delete plugins

corsacca:


prykon (commented):
New permissions work as expected.

prykon (approved):

PR #2022, opened 2023-03-30
Fixed delete button bug

prykon:
After deactivating a plugin the 'delete' button would show even if user didn't have install plugin permissions. (Nothing happened on click if no permissions)

Button doesn't appear anymore if user has no install plugin permissions.


corsacca:
Thanks @prykon !

PR #2034, opened 2023-04-11
New settings delete tile

prykon:
Added 'delete' feature for custom tiles.

Will do:
- Delete custom tiles

Will not do:
- Delete default tiles
- Delete custom tile field options


prykon:
Corresponds to this issue: https://github.com/orgs/DiscipleTools/projects/13/views/1?pane=issue&itemId=23991705

corsacca:
Let's make save and delete be on the same row
image

prykon:
@corsacca Delete feature only appears for custom tiles. Will add the feature that moves its fields to the 'No Tile' section and make the 'Delete' link be on the same row.

prykon:
Example of animation for fields moving into the 'No Tile / Hidden' menu when their parent tile gets deleted.

https://user-images.githubusercontent.com/36511666/233102466-8282827d-249f-4fde-b191-39e5eab281d1.mov

PR #2035, opened 2023-04-12
Added 'field type' label to Edit Field modal

prykon:
Screenshot 2023-04-11 at 21 11 38


corsacca:
@prykon lets use the labels instead of the keys: 
image

prykon:
@corsacca We're using 'Dropdown' in the legacy wording. Would you like to change it to 'Key Select'? 'Dropdown' is more intuitive, IMO...

Screenshot 2023-04-17 at 08 59 49

corsacca:
Correct, we'd want it to say Dropdown.

I'm thinking we might want to build a Field Type array in the D.T core that could be reference from anywhere in the system.

Issue #2020, opened 2023-03-30
Extensions (D.T)


corsacca:
Show error when api call failed:
![image](https://user-images.githubusercontent.com/24901539/228871145-2f2bd8b1-4ca3-48a9-963e-b277735f3d9e.png)

PR #2038, opened 2023-04-13
Make error message appear on false responses

prykon:
If the plugin install, delete, activate or deactivate response returns false, an error message is shown.

https://user-images.githubusercontent.com/36511666/231820496-ff45d9cf-85dc-4340-9633-92827eb23094.mov


prykon:
Addresses issue https://github.com/DiscipleTools/disciple-tools-theme/issues/2020

corsacca:
Thanks @prykon

PR #2057, opened 2023-05-05
Added field key check functionality before adding field

prykon:
**Functionality example:**
https://user-images.githubusercontent.com/36511666/236450540-705aeed5-56dc-44fc-9b94-9c7b8cd1fd89.mov

**Refactored some code:**
- Fixed input field's name `edit-tile-label -> new-field-name`
- The `new_field_name`, `new_field_type`, and `new_field_private` elements didn't need to have the tile id in their id name because only one modal is displayed at a time so no ambiguity conflicts should arise.
- Other input fields might have the tile id in their id name but only this one was fixed in order to not broaden the PRs scope

### DIFF
--- a/dt-core/admin/js/dt-settings.js
+++ b/dt-core/admin/js/dt-settings.js
@@ -321,117 +321,120 @@ jQuery(document).ready(function($) {
 
         $.each(fields, function(field_index, field_key) {
             var field = window['field_settings']['post_type_settings']['fields'][field_key];
+            if( !field?.type ){
+              return;
+            }
 
-                var icon_html = '';
-                if ( field['icon'] ) {
-                    icon_html = `<img src="${field['icon']}" class="dt-icon lightgray"></img>`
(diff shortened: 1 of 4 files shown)


corsacca:
Nice: 
image

dt-core/admin/js/dt-settings.js
```diff
@@ -179,9 +179,9 @@ jQuery(document).ready(function($) {
                     sortable_field_options_ordering.push(option_element.dataset['fieldOptionKey']);
                 });
                 API.update_field_options_order(post_type, field_key, sortable_field_options_ordering).promise().then(function(new_order){
-                    var old_order = window.field_settings.post_type_settings.fields[field_key].default;
+                    var old_order = window['field_settings']['post_type_settings']['fields'][field_key]['default'];
```
corsacca, line 182: @prykon why the revert back? :D

PR #2066, opened 2023-05-11
User IDs were being sent as an integer 

micahmills:
User IDs were being sent as an integer instead of a string like contact ids when requesting a compact.


corsacca:
@micahmills For consistency let's move towards IDs always being (int)

PR #2069, opened 2023-05-16
Simple select for bool field

corsacca:


squigglybob (commented):
image

How do we want the boolean to display in the Details summary section?
Would it be better to have Yes/No even if it's false?

At the moment it only shows in the details section if it is true, and not if it is false.

squigglybob:
Also I noticed that you can't create a bool field in the DT settings Fields section. Is that something we want to add as well, or if we are just getting a frontend view of a code only bool field?

corsacca:
Good Catch:
image
image

squigglybob:
Sweet, looks good

PR #2071, opened 2023-05-17
workflows: handle not_equals/not_contain when field is not set

cairocoder01:
When setting a condition for a workflow with the `not_contain` comparison but the field is not set at all, the condition was not being matched. However, if the field is not set, we know that it doesn't contain the given value. This handles the case where the field is not set for `not_contain` and `not_equals`


corsacca:
Thanks @cairocoder01 !

Issue #1538, opened 2021-12-08
activity during 'date range' search feature


oldvill:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
Getting a list of contacts that had an action/activity happen during a time range sounds useful. 
ie:
--Which contacts were made active in december?
--How many and which contact states belief in "time range"
--For my DT instances, "How many people received a bible in 'x' month"

**Describe alternatives you've considered**
the introduction of the 'Fields over Time Charts' has helped but its limited. 
before this, there was no alternative that i knew of

**Additional context**

PR #2073, opened 2023-05-18
#1538 - Activity During Date Range Metric

kodinkat:
- fixes: #1538 
- fixes: #2080 
---

- New Activity During Date Range project metric introduced.
- Metric provides the ability to query post records by field and within a given date range.
- Ability to click-through to listed results post records for more detail.
- Added 2nd level filter support for typeahead based field types: `connection`, `location` & `user_select`.
- Display new value field was updated with.

Screenshot 2023-05-30 at 16 45 31

Screenshot 2023-05-30 at 16 46 19

Screenshot 2023-05-30 at 16 47 06

### DIFF
--- a/dt-metrics/combined/date-range-activity.php
+++ b/dt-metrics/combined/date-range-activity.php
@@ -223,7 +223,7 @@ class DT_Metrics_Date_Range_Activity extends DT_Metrics_Chart_Base
                             $field_types[] = 'connection ' . $p2p_direction;
                         }
                     }
-                    $field_type_sql = "AND field_type IN (" . dt_array_to_sql( $field_types ) . ")";
+                    $field_type_sql = 'AND field_type IN (' . dt_array_to_sql( $field_types ) . ')';
                 }
                 $meta_key_sql = ( $field_type == 'connection' ) ? "AND meta_key LIKE '". esc_sql( !empty( $settings['p2p_key'] ) ? $settings['p2p_key'] : '%' ) ."'" : "AND meta_key LIKE '" . esc_sql( $params['field'] ) . "'";
 
--- a/dt-metrics/combined/date-range-activity.php
+++ b/dt-metrics/combined/date-range-activity.php
@@ -149,18 +149,6 @@ class DT_Metrics_Date_Range_Activity extends DT_Metrics_Chart_Base
         );
 
         register_rest_route(
-            $namespace, '/metrics/dummy_endpoint', [
-                [
-                    'methods'  => WP_REST_Server::READABLE,
-                    'callback' => [ $this, 'dummy_endpoint' ],
-                    'permission_callback' => function(){
-                        return $this->has_permission();
-                    },
-                ],
-            ]
-        );
-
-        register_rest_route(
             $namespace, '/metrics/date_range_activity', [
                 [
                     'methods'  => WP_REST_Server::CREATABLE,
@@ -199,10 +187,6 @@ class DT_Metrics_Date_Range_Activity extends DT_Metrics_Chart_Base
         }
 
         return [];
-    }
-
-    public function dummy_endpoint( WP_REST_Request $request ){
-        return new WP_REST_Response( [] );
     }
 
     public function date_range_activity( WP_REST_Request $request ){
--- a/dt-metrics/combined/date-range-activity.php
+++ b/dt-metrics/combined/date-range-activity.php
@@ -197,6 +197,7 @@ class DT_Metrics_Date_Range_Activity extends DT_Metrics_Chart_Base
             ] );
 
             global $wpdb;
+            // phpcs:disable
             $results = $wpdb->get_results( $wpdb->prepare( "
             SELECT *
             FROM $wpdb->dt_activity_log
@@ -208,6 +209,7 @@ class DT_Metrics_Date_Range_Activity extends DT_Metrics_Chart_Base
             $meta_value_sql
             ORDER BY hist_time DESC;
             ", $params['post_type'], $params['ts_start'], $params['ts_end'] ), ARRAY_A );
+            // phpcs:enable
 
             // Package result findings and return.
             $posts = [];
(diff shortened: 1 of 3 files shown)


corsacca:
@kodinkat 
Enable first layer all fields
image

List will show field changes during the time period.
ex all contact with baptism date change in range
ex all contact the status change in range

Some fields have optional second filter ( all contacts set to status active. all contacts assigned to user y)
-  key select, tags, multisilect, connection have

Please test and confirm fields first lever:

- [ ] multi_select
- [ ] key_select
- [ ] tags
- [ ] communication_channel
- [ ] connection
- [ ] user_select
- [ ] text
- [ ] textarea
- [ ] link
- [ ] date
- [ ] number
- [ ] boolean
- [ ] location
- [ ] location_meta

With optional second filter:
- [ ] multi_select
- [ ] key_select
- [ ] tags
- [ ] communication_channel
- [ ] connection
- [ ] user_select
- [ ] location?

kodinkat:
@corsacca Ready for review...

corsacca:
Error on Baptism Date all time
image

Connection fields seem to be returning values for any connection field.

kodinkat:
@corsacca hopefully all major bugs now smoked out, but let me know if any more are encountered....

corsacca:
@kodinkat 

- can you try adding the demo data from the demo plugin? The activity from connections fields from it don't show.
- Adding and then removing a connection should not show the same thing twice:
image

dt-metrics/combined/date-range-activity.php
```diff
@@ -0,0 +1,370 @@
...
+
+    public function field_settings( WP_REST_Request $request ) {
+        if ( !$this->has_permission() ) {
+            wp_send_json_error( new WP_Error( 'get_field_settings', 'Missing Permissions', [ 'status' => 400 ] ) );
+        }
+        $url_params = $request->get_url_params();
+        return $this->get_field_settings( $url_params['post_type'] );
+    }
+
+    public function render_field_html( WP_REST_Request $request ){
```
corsacca, line 176: @kodinkat how can we steer away from sending html in API calls? 
It's a security risk as the result can't be escaped

dt-metrics/combined/date-range-activity.php
```diff
@@ -0,0 +1,370 @@
...
+                $meta_key_sql = ( $field_type == 'connection' ) ? "AND meta_key LIKE '%'" : "AND meta_key LIKE '" . esc_sql( $params['field'] ) . "'";
+
+                $values = [];
+                foreach ( $params['value'] ?? [] as $value ){
+                    if ( isset( $value['ID'] ) ){
+                        $values[] = $value['ID'];
+                    }
+                }
+                $meta_value_sql = ( !empty( $values ) ? 'AND meta_value IN (' . dt_array_to_sql( $values ) . ')' : "AND meta_value LIKE '%'" );
+                $obj_subtype_sql = ( $field_type == 'connection' ) ? "AND object_subtype = '" . esc_sql( $params['field'] ) . "'" : '';
```
corsacca, line 239: @kodinkat object subtype used to be 'p2p', so this does not pick up on historical data.
image

corsacca (changes requested):

dt-metrics/combined/date-range-activity.php
```diff
@@ -0,0 +1,370 @@
...
+                [
+                    'methods'  => WP_REST_Server::READABLE,
+                    'callback' => [ $this, 'render_field_html' ],
+                    'permission_callback' => '__return_true'
+                ]
+            ]
+        );
+
+        register_rest_route(
+            $namespace, '/metrics/dummy_endpoint', [
```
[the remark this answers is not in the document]
kodinkat: Given `window.METRICS.setupDatePicker()` shape, `dummy_endpoint` acts as a placeholder; used when initialising date range picker element.

dt-metrics/combined/date-range-activity.php
kodinkat: Could capture all field html when initially loading the page; or, refresh page and load on refresh... Personally, I'm not keen on either approach... Alternatively, html could be generated within frontend, but this too would add to code / logic duplication.

dt-metrics/combined/date-range-activity.php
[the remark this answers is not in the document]
corsacca: @kodinkat what about the existing window.METRICS.setupDatePickerWithoutEndpoint() ?

dt-metrics/combined/date-range-activity.php
kodinkat: Historical `p2p` activity entries now captured; however, please note that any `p2p` entries detected within given date range, will be included; as there is no apparent way to filter by field id.

dt-metrics/combined/date-range-activity.php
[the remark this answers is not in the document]
kodinkat: ah, good spot; as I did not see that! :)

dt-metrics/combined/date-range-activity.php
corsacca: @kodinkat I think there are 2options here
We use the meta_key and the field type:
image


Or we write a migration to upgrade the old activity to match the new shape.

kodinkat: @corsacca yep, maybe the first option, re exploring the use of both the meta_key and field_type, to determine actual field_id.... As the 2nd option sounds like a new ticket? ;)

corsacca: Yeah. Let's go for #1 and see where we get.

kodinkat: @corsacca does the latest sql tweak hold true for you?

corsacca: @kodinkat almost. 
We do need to add the field type. Other wise it gets both the  `Baptized` and the `baptized by` field.
image

Safir:

corsacca:
Thank you @kodinkat !

PR #2082, opened 2023-05-31
New settings misc bug fixes

prykon:
**Misc bug fixes:**
- Tile previews normally now
- Added tile preview for textarea fields
- You can't create a tile with an empty name
- You can't edit a tile with an empty name
- You can't create a field with an empty name
- You can't edit a field with an empty name
- You can't create a field option with an empty name
- You can't edit a field option with an empty name
- Removed edit feature from the 'default blank' element in field options


corsacca:
Nice work @prykon !

corsacca:
@prykon why remove the edit on "Default Blank" ?

Issue #2100, opened 2023-06-26
Integrate DT Post Types Plugin Functionality Into Customization (D.T) Beta


kodinkat:
- v1
  - Add new post-type modal, with the following layout:

![image](https://github.com/DiscipleTools/disciple-tools-theme/assets/19330800/3bc93d2f-99ae-4feb-8366-b3654f85a0d8)

- v2
  - Auto assign user(s) to new post type capability roles.

PR #2101, opened 2023-06-26
#2100 - Integrate DT Post Types Plugin Functionality Into Customization (D.T) Beta

kodinkat:
- fixes: #2100 
---

Screenshot 2023-06-28 at 17 28 41

Screenshot 2023-06-28 at 17 31 10

Screenshot 2023-06-28 at 17 31 52

### DIFF
--- a/dt-core/admin/admin-settings-endpoints.php
+++ b/dt-core/admin/admin-settings-endpoints.php
@@ -282,6 +282,8 @@ class Disciple_Tools_Admin_Settings_Endpoints {
                 $admin->add_cap( 'access_' . $key );
                 $admin->add_cap( 'create_' . $key );
                 $admin->add_cap( 'delete_any_' . $key );
+                $admin->add_cap( 'view_any_' . $key );
+                $admin->add_cap( 'update_any_' . $key );
 
                 // Return successful response.
                 $response = [
@@ -386,6 +388,8 @@ class Disciple_Tools_Admin_Settings_Endpoints {
                 $admin->remove_cap( 'access_' . $post_type );
                 $admin->remove_cap( 'create_' . $post_type );
                 $admin->remove_cap( 'delete_any_' . $post_type );
+                $admin->remove_cap( 'view_any_' . $post_type );
+                $admin->remove_cap( 'update_any_' . $post_type );
 
                 $deleted = true;
                 $deleted_msg = 'Deleted ' . $deleted_post_counter . ' out of ' . $deleted_post_count .' posts.';
--- a/dt-core/admin/admin-settings-endpoints.php
+++ b/dt-core/admin/admin-settings-endpoints.php
@@ -18,6 +18,16 @@ class Disciple_Tools_Admin_Settings_Endpoints {
     public function __construct() {
         $this->namespace = $this->context;
         add_action( 'rest_api_init', [ $this, 'add_api_routes' ] );
+
+        add_action( 'after_setup_theme', function (){
+            $custom_post_types = get_option( 'dt_custom_post_types', [] );
+
+            foreach ( $custom_post_types as $post_type_key => $post_type ){
+                if ( class_exists( 'Disciple_Tools_Post_Type_Template' ) ){
+                    new Disciple_Tools_Post_Type_Template( $post_type_key, $post_type['single_name'] ?? $post_type_key, $post_type['plural_name'] ?? $post_type_key );
+                }
+            }
+        }, 100 );
     }
 
     /**
@@ -68,6 +78,22 @@ class Disciple_Tools_Admin_Settings_Endpoints {
             $this->namespace, '/create-new-post-type', [
                 'methods' => 'POST',
                 'callback' => [ $this, 'create_new_post_type' ],
+                'permission_callback' => [ $this, 'default_permission_check' ],
+            ]
+        );
+
+        register_rest_route(
+            $this->namespace, '/update-post-type', [
+                'methods' => 'POST',
+                'callback' => [ $this, 'update_post_type' ],
+                'permission_callback' => [ $this, 'default_permission_check' ],
+            ]
+        );
+
+        register_rest_route(
+            $this->namespace, '/delete-post-type', [
+                'methods' => 'POST',
+                'callback' => [ $this, 'delete_post_type' ],
                 'permission_callback' => [ $this, 'default_permission_check' ],
             ]
         );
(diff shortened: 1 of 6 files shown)


corsacca:
Nice work @kodinkat 
Use: `Display [name] tab in navigation bar` text
Use: `Record type` instead of post type. (at least 2 places).
Use spinner and success checkmark instead of alert modal when saving settings.

corsacca:
Delete button.
Opens modal with text:
Are you sure you want to delete [name] and field settings?
Also delete field settings.

Add check box:
[ ] delete all records of type [type]?

Add code to delete all posts, meta, activity etc of that post type. 
see delete_post() in posts.php
We can use sql like `DELETE FROM $wpdb->dt_notifications WHERE post_id IN ( SELECT ID from $wpdb->posts where post_type = %)`

corsacca:
It looks like we have duplicate settings. Can we merge the `dt_custom_post_types` and `dt_post_type_custom_updates` options into one and use `dt_custom_post_types` option?
Add a "is_custom" => true value when create a record type, the we know it is custom.

dt-posts/custom-post-type.php
```diff
@@ -24,7 +24,7 @@ public function __construct( string $post_type, string $singular, string $plural
         add_filter( 'desktop_navbar_menu_options', [ $this, 'add_navigation_links' ], 20 );
         add_filter( 'dt_nav_add_post_menu', [ $this, 'dt_nav_add_post_menu' ], 10, 1 );
         add_filter( 'dt_templates_for_urls', [ $this, 'add_template_for_url' ] );
-        add_filter( 'dt_get_post_type_settings', [ $this, 'dt_get_post_type_settings' ], 10, 4 );
+        add_filter( 'dt_get_post_type_settings', [ $this, 'dt_get_post_type_settings' ], 1000, 4 ); // Called last within chain to ensure custom updates are applied last.
```
corsacca, line 27: @kodinkat this applies to all post types.
The number is low, so other hooks can add on to this one.
Why should it be 1000?

dt-posts/custom-post-type.php
```diff
@@ -10,23 +10,26 @@ class Disciple_Tools_Post_Type_Template {
     public $singular;
     public $plural;
     public $search_items;
+    public $hidden;
 
-    public function __construct( string $post_type, string $singular, string $plural ) {
+    public function __construct( string $post_type, string $singular, string $plural, bool $hidden = false ) {
```
corsacca, line 15: @kodinkat i don't see any construction with the hidden param

dt-posts/custom-post-type.php
```diff
@@ -86,7 +89,28 @@ public function rewrite_init(){
         add_rewrite_rule( $this->post_type . '/([0-9]+)?$', 'index.php?post_type=' . $this->post_type . '&p=$matches[1]', 'top' );
     }
 
+    public function refresh_custom_post_type_settings(){
+        $post_type_updates = get_option( 'dt_custom_post_types', [] );
```
corsacca, line 93: @kodinkat can this be move to the construct?

dt-core/admin/admin-settings-endpoints.php
```diff
@@ -222,6 +256,156 @@ public static function get_post_fields() {
...
+                    update_option( 'dt_field_customizations', $field_customizations );
+                    wp_cache_delete( $post_type . '_field_settings' );
+                }
+
+                // If specified, proceed with deletion of all associated records.
+                $deleted_post_count = 0;
+                $deleted_post_counter = 0;
+                if ( isset( $params['delete_all_records'] ) && $params['delete_all_records'] ){
+                    global $wpdb;
+                    $posts = $wpdb->get_results( $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE post_type = %s", $post_type ), ARRAY_A );
```
corsacca, line 376: @kodinkat please circle back to using the SQL we talked about or explain why we aren't using it.
This method will not work on a large number of posts.

dt-core/admin/admin-settings-endpoints.php
```diff
@@ -222,6 +256,156 @@ public static function get_post_fields() {
...
+                }
+
+                // Remove custom post type admin capabilities.
+                $admin = get_role( 'administrator' );
+                $admin->remove_cap( 'dt_all_admin_' . $post_type );
+                $admin->remove_cap( 'access_' . $post_type );
+                $admin->remove_cap( 'create_' . $post_type );
+                $admin->remove_cap( 'delete_any_' . $post_type );
+                $admin->remove_cap( 'view_any_' . $post_type );
+                $admin->remove_cap( 'update_any_' . $post_type );
```
corsacca, line 392: @kodinkat these remove_caps are no longer needed with the new caps filters.

dt-core/admin/menu/tabs/tab-customizations.php
```diff
@@ -82,7 +82,16 @@ public function admin_enqueue_scripts() {
         wp_enqueue_style( 'dt_settings_css' );
 
         $post_type = self::get_parameter( 'post_type' );
-        if ( !isset( $post_type ) || is_null( $post_type ) ) {
+        if ( !isset( $post_type ) || is_null( $post_type ) ){
```
corsacca, line 85: @kodinkat is this section needed since the dt-settings bellow has these options?

kodinkat: Yes, otherwise an exception is thrown when attempting to create a new record type prior to selecting any existing ones.

Screenshot 2023-07-06 at 12 08 24

dt-posts/custom-post-type.php
kodinkat: To ensure custom updates are applied at the end of hooks execution chain. Number now reduced to something more sensible, and `contacts` and `groups` `base-setup` classes updated to also check for new updates.

corsacca: @kodinkat the custom post type filter needs to be loaded before the others, so that contacts, groups, etc can build upon it. Are there certain settings that are being overwritten? 
We can add a second `dt_get_post_type_settings` hook with a high value if needed.

corsacca: I added a later filter that will get the labels correctly. I see why that was tricky. Thanks!

dt-core/admin/menu/tabs/tab-customizations.php
corsacca: I moved the code around a little to easily add to the base object.

corsacca:
Thank you for all your work in this one @kodinkat! Nicely done.

corsacca:
will merge in with https://github.com/DiscipleTools/disciple-tools-theme/pull/2106

PR #2103, opened 2023-06-30
fix: clear any WP login cookies before JWT login

squigglybob:
@corsacca 

In PG, there is an issue, where if a user logs in with the admin side which uses PHP, and then comes back to their logged in PG (with JWT token) then it has an issue and can't check the auth properly.

This might be a sledgehammer solution, and not sure if it will have any other repercussions (but I don't think it would)

But would essentially make sure, that if someone is logging in with JWT, that the WP side is logged out first.

This makes sure that the JWT check_auth function continues to work and doesn't lead to any unexpected behaviour where the check_auth doesn't return the user correctly.

### DIFF
--- a/dt-login/login-endpoints.php
+++ b/dt-login/login-endpoints.php
@@ -106,6 +106,9 @@ class DT_Login_Endpoints {
         }
 
         if ( DT_Login_Methods::MOBILE === $login_method ) {
+            /* Make sure that any WP login tokens are cleared before logging in with JWT */
+            wp_logout();
+
             require_once( get_template_directory() . '/dt-core/libraries/wp-api-jwt-auth/public/class-jwt-auth-public.php' );
             $response = Jwt_Auth_Public::validate_token( $request );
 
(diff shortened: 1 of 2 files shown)


squigglybob:
It's not a deal breaker if we can't fix it as it will only affect anyone who logs into the admin side of PG, which is only us.

But not sure if there might be other places it could catch people out as well

dt-login/login-endpoints.php
```diff
@@ -106,6 +106,9 @@ public function check_auth( WP_REST_Request $request ) {
         }
 
         if ( DT_Login_Methods::MOBILE === $login_method ) {
+            /* Make sure that any WP login tokens are cleared before logging in with JWT */
+            wp_logout();
```
corsacca, line 110: @squigglybob , should we add a `is logged in check` first?

squigglybob: Nope, it works fine. I've added the check in

corsacca:
Thanks @squigglybob !

PR #2105, opened 2023-07-04
fix: add firebase Displayname to user manager if not exists

squigglybob:
fixes: user having UID as displayname when logging in with email through SSO login UI on Prayer.Global


squigglybob:
Thanks @corsacca

PR #2110, opened 2023-07-08
Allow whitelisting API endpoints to skip DT auth requirement

cairocoder01:
Certain third party plugins add API endpoints that needs to not have authentication on them that DT enforces by default. There is currently a filter that can whitelist endpoints via plugin, but this adds a simple way in core to add endpoints that should be allowed to pass through.

For reference, my use case is the Wordfence plugin that has a central dashboard that hits an API endpoint for remote management and configuration.


corsacca:
Looks good. Lets move this to the security tab instead of the general one.

cairocoder01:
That makes sense. It's moved now. @corsacca

corsacca:
Thanks @cairocoder01!

PR #2114, opened 2023-07-13
Fix deprecated `FILTER_SANITIZE_STRING`

corsacca:


kodinkat (approved):
Holding true for me! Good job!

Issue #2109, opened 2023-07-07
Custom Role removes permissions from administrator role


cairocoder01:
**Describe the bug**
When adding a new custom role, I noticed that my `administrator` role lost some of the same permissions that I had added to the custom role. After digging around, I found that any permissions that are granted to a custom role will actually be removed from the administrator role and can't be edited in the Roles admin.

**To Reproduce**
1. Create a custom role with `manage_options` permission from the core WordPress options
2. Login as user with new role to see they have access to things with `manage_options` permission.
3. Refresh Roles admin and open Administrator role to see that `manage_options` has been removed from the role
4. Logout and login as admin to see that they have lost the `manage_options` permission

**Expected behavior**
`administrator` role should not be affected by custom roles

**Screenshots**

**Disciple.Tools Theme:**
- version: [e.g 1.42.1]
- setup: single site, local

**Additional context**
When testing with LocalWP, in order to restore admin permissions, I had to open up SSH and use the CLI to reset the permissions once I removed the extra permission from the custom role. `wp role reset administrator`

In testing, I found that the role was being removed from `/dt-core/setup-functions.php:62`. This happens as soon as the theme is loaded.

I had tried to fix the admin role by going into the database directly and updating the table `wp_options` and option named `wp_user_roles` to include the missing permissions. They always got reset again though. This could be an alternative to the CLI reset above once you remove the permission from the custom role.

cairocoder01:
Resolved by #2115

PR #2115, opened 2023-07-13
Don't remove wp caps from the administrator

corsacca:


kodinkat (approved):
Excellent, as this holds up on my end! Good job!

cairocoder01:
resolves #2109 

Tested this and it looks good.

corsacca:
Thanks @cairocoder01 and @kodinkat

PR #2116, opened 2023-07-13
feat: add filters to login allowed js and css

squigglybob:
Allow css / js to be allowed through the magic link net, to change styles of the built in login page


squigglybob:
Hey @corsacca,
Not sure why the wpunit test install is falling over here. 😬

PR #2148, opened 2023-08-04
feat: do an action when the user profile has updated with an existing corresponding contact

squigglybob:


squigglybob:
@corsacca what do you think about using the same do_action hook name in both the update and create areas for a corresponding contact?

squigglybob:
I opted for changing the hook name when creating a contact for a user, as it was providing a contact record rather than a WP_Post object

Issue #2132, opened 2023-07-31
Display People Group Setting


corsacca:
Migrate old setting to new way of saving.
Remove from D.T settings page.

<img width="543" alt="image" src="https://github.com/DiscipleTools/disciple-tools-theme/assets/24901539/7ba3b834-e1f5-4cfe-be40-45acf0b02393">

PR #2164, opened 2023-08-10
Relocated Display People Group Settings To New Custom Setup

kodinkat:
- fixes: #2132


corsacca:
@kodinkat is there the migration part?

corsacca:
@kodinkat this is a core migrations instead of a mapping migration.

PR #2167, opened 2023-08-16
Fix/login fixes

squigglybob:
## Adding parts for Zume5
- [x] check that passwords match
- [x] use login_url dt login field to show the custom login page
- [x] add google captcha section to sso-login fields
- [x] make sso options visible to all users

---
## Security bits

- [x] an option to make registration invite only by default on multisite (on single it's blocking fine if registration on)
- [x] and don't register the user by default if it is a private invite only site
- [x] and don't display registration page or buttons
- [x] turn off subsite registration option by default when site is created


squigglybob:
@corsacca ok, as long as this one passes it's tests, this one is ready to be merged.
I'll be back in 2 weeks if there are any issues with it

squigglybob:
I've moved the last 2 small security fixes for the custom login page into another issue #2193

PR #2179, opened 2023-08-27
Updating next branch to 1.47.4

micahmills:


micahmills:
I am not sure why the PHPCS is failing here but not on the master branch? Is there something I am missing here?

PR #2191, opened 2023-09-05
ML#116 [ CORE ] - Ensure Expired Links Remain Invalid

kodinkat:


dt-reports/magic-url-class.php
```diff
@@ -274,6 +274,12 @@ public function parse_url_parts(){
                 if ( ! empty( $instance_id ) ) {
                     $elements['instance_id'] = $instance_id;
                 }
+
+                // Wider callout to ensure link is still valid.
+                if ( apply_filters( 'dt_ml_obj_link_expired', $elements['instance_id'], $elements['post_id'] ) === true ) {
```
corsacca, line 279: @kodinkat 
`dt_magic_link_continue` will be more reusable for other possible uses.
return false if issue
check !$dt_magic_link_continue

corsacca:
@kodinkat see updated format.

Issue #2196, opened 2023-09-18
Dynamic Metrics Field Activity Creation Date Not Responding


kodinkat:
Creation Date not returning any records within field activity dynamic metrics.

![Screenshot 2023-09-18 at 09 22 13](https://github.com/DiscipleTools/disciple-tools-theme/assets/19330800/22d9bbcf-0d75-45b5-8b56-1caad77b70d6)

PR #2197, opened 2023-09-18
#2196 - Dynamic Metrics Field Activity Creation Date Not Responding

kodinkat:
- fixes: #2196


dt-metrics/records/date-range-activity.php
```diff
@@ -204,6 +205,37 @@ public function date_range_activity( WP_REST_Request $request ){
             $obj_subtype_sql = '';
 
             // Accommodate special cases.
+            if ( ( $field_type == 'date' ) && ( $params['field'] == 'post_date' ) ) {
+
+                // phpcs:disable
```
corsacca, line 210: PHPCS :)

PR #2200, opened 2023-09-21
Login fixes

corsacca:


squigglybob (commented):
Looks good to me bro :)

I think getting the url when it doesn't exist is probably the right thing to do most times, and the $url function paramater allows it to be overridden where needed which is fine too

Issue #1973, opened 2023-03-09
When deleting a user reassign all their contacts to the base user


corsacca:
Hook into:
![image](https://user-images.githubusercontent.com/24901539/224967076-03e143cb-d0d4-441e-ac3a-66379a2a5144.png)

PR #2204, opened 2023-09-21
#1973 - When Deleting A User Reassign All Their Contacts To The Base User

kodinkat:
- fixes: #1973


corsacca:
@kodinkat 
When removing a User from a blog,
the $reassign_id is `0`
So this PR sets the assigned to to `user-0` and shares all the records with user id `0`

currently we check is $reasign_id is `null`. default is `0`
image

PR #2206, opened 2023-09-28
Feat/option to hide login page nav

squigglybob:
@corsacca 

The missing piece for me not understanding Chris's desire for the login page to be within the login magic link, was the ability to control the login page so that there are no other links anywhere else but to do the login/registration and then be redirected back to the magic link.

ergo restricting the user's flow so that they don't get distracted by any navbar/footer links etc.

I've added the necessary code to hide the nav bar when the hide-nav query param is in the url on the custom login page.

setCookie shared js function was a piece I needed for the language switching in zume, but forgot to create a PR for so I've just rolled it into here.

### DIFF
--- a/dt-assets/js/shared-functions.js
+++ b/dt-assets/js/shared-functions.js
@@ -575,11 +575,17 @@ window.SHAREDFUNCTIONS = {
     }
     return "";
   },
-  setCookie(cname, cvalue, exdays = 30 ) {
-    var d = new Date();
-    d.setTime(d.getTime() + (exdays * 24 * 60 * 60 * 1000));
-    var expires = "expires="+d.toUTCString();
-    document.cookie = cname + "=" + cvalue + ";" + expires + ";path=/";
+  setCookie(cname, cvalue, path, exdays ) {
+    let cookie = `${cname}=${cvalue};`
+    if ( Number.isInteger( exdays ) && exdays > 0 ) {
+      var d = new Date();
+      d.setTime(d.getTime() + (exdays * 24 * 60 * 60 * 1000));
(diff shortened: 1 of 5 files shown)


dt-assets/js/shared-functions.js
```diff
@@ -575,6 +575,12 @@ window.SHAREDFUNCTIONS = {
     }
     return "";
   },
+  setCookie(cname, cvalue, exdays = 30 ) {
```
corsacca, line 578: @squigglybob save_json_cookie() also works on strings.

on a subdirectory multisite, would `/` set a cookie for the whole installation?

squigglybob: yes it would and good point. Maybe I should add exdays to save_json_cookie

I think part of why I added setCookie was because there is a getCookie function but no setCookie function, but you're right it's not taking a path parameter

squigglybob: I think we need a setCookie function which doesn't stringify strings by default.
It's the difference between saving a cookie that looks like this
image

and this

image

squigglybob: what about adding exdays to save_json_cookie and refactoring that to use setCookie underneath

squigglybob:
hey @corsacca If the setCookie thing, is holding up this PR I can pull the cookie thing out as a seperate PR to get this one through

corsacca:
@squigglybob should this work?
image

corsacca:
Weirdly, this is the login screen and there is the "login" link in the link section. (separate issue)

squigglybob:
Yeah the home link should probably be hidden as well

squigglybob:
@corsacca home button hidden now

Issue #2203, opened 2023-09-21
Fields Charts - Records change graph


corsacca:
The last line graph should show:
The number of records where the field (or field and option) were set in that month.
If a contact stats is changed to "active" every month, then it would appear in each month.
Let go over each field type and update the logic.

This chart should use the changed-posts enpoint. Lets separate out the logic.

PR #2208, opened 2023-09-29
#2203 - Fields Charts - Records Change Graph

kodinkat:
- fixes: #2203

### DIFF
--- a/dt-metrics/counters/counter-post-stats.php
+++ b/dt-metrics/counters/counter-post-stats.php
@@ -45,8 +45,7 @@ class DT_Counter_Post_Stats extends Disciple_Tools_Counter_Base
                 ];
 
                 foreach ( $post_changes_sql as $changes_sql ) {
-                    // phpcs:disable
-                    $prepared_sql = $wpdb->remove_placeholder_escape( $wpdb->prepare(
+                    $post_changes[] = $wpdb->get_results( $wpdb->remove_placeholder_escape( $wpdb->prepare(
                         "
                                 SELECT
                                     COUNT( posts.id ) AS count,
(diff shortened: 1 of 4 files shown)


dt-metrics/counters/counter-post-stats.php
```diff
@@ -25,6 +25,181 @@ public function __construct() {
...
+
+                $post_changes = [];
+                $post_changes_sql = [
+                    'posts.added > posts.deleted', // Added
+                    'posts.added <= posts.deleted' // Deleted
+                ];
+
+                foreach ( $post_changes_sql as $changes_sql ) {
+                    // phpcs:disable
+                    $post_changes[] = $wpdb->get_results(
```
corsacca, line 49: Nice work on this @kodinkat !
Please use $wpdb->prepare on these queries.

`log.object_note LIKE '%Added%', 1, 0` might throw an error. 

You can do something like `log.object_note LIKE '%s', 1, 0` and pass `'%Added%'` in the prepared variables.

corsacca (changes requested):
Nice work Karl.
Please user $wpdb-prepare on these queries. 

It might throw an error on

dt-metrics/counters/counter-post-stats.php
```diff
@@ -1264,7 +1274,7 @@ public static function get_posts_by_field_in_date_range_changes( $post_type, $fi
 
                     // phpcs:disable
```
corsacca, line 1275: @kodinkat please remove all the phpcs:disable too!

corsacca (changes requested):

corsacca (approved):

corsacca:
Whanks @kodinkat !

Issue #2212, opened 2023-10-05
Metrics > Dynamic > Field Activity


corsacca:
Use server time when creating reports instead of the user's timezone.

This way every user will see the same report data.

![image](https://github.com/DiscipleTools/disciple-tools-theme/assets/24901539/c4717d5e-77d9-454e-aae6-ac061c1a3b08)

Dev:
send `2023-10-01` instead of the timestamp and let the server convert it.

PR #2216, opened 2023-10-10
#2212 - Metrics Dynamic Field Activity Server Time Usage

kodinkat:
- fixes: #2212


corsacca:
@kodinkat do we need to keep both?

kodinkat:
> @kodinkat do we need to keep both?
> image

More so for future-proofing; but we can always re-introduce if needed... I'll remove....

dt-metrics/records/date-range-activity.js
```diff
@@ -190,8 +194,8 @@ function project_activity_during_date_range() {
     url.searchParams.set('record_type', post_type);
     url.searchParams.set('field', field_id);
     url.searchParams.set('value', search_param_value );
-    url.searchParams.set('ts_start', payload.ts_start);
-    url.searchParams.set('ts_end', payload.ts_end);
+    url.searchParams.set('ts_start', ts_start);
```
corsacca, line 197: @kodinkat should this be date_start now?

corsacca:
Thanks Karl!

PR #2250, opened 2023-11-01
#2249 - Support Creation Of User Select Field Types

kodinkat:
- fixes: #2249


corsacca:
Thanks @kodinkat!
The comments and activity section is not listing any activity from changing values in the custom users selects. Can you look into that?

corsacca:
Thanks @kodinkat!

Issue #2255, opened 2023-11-08
User select list only users that have access


corsacca:
When filling out a user select, the returned users from the api should only be users that have access to the relevant post type.

PR #2257, opened 2023-11-13
#2255 - User Select List Only Users That Have Access

kodinkat:
- fixes: #2255


corsacca:
Very nice @kodinkat !
I added a small edge case: https://github.com/DiscipleTools/disciple-tools-theme/pull/2257/commits/2af6af87834cbefc1409959dd704a4d153d08f15

Issue #2235, opened 2023-10-20
Genmap: Infinite loop notice


corsacca:
notice at bottom of chart
Not showing 5 records because of infinite loop.
link to modal? to see problematic records

_Originally posted by @corsacca in https://github.com/DiscipleTools/disciple-tools-theme/issues/2213#issuecomment-1755679656_

PR #2261, opened 2023-11-14
#2235 - Genmap: Infinite Loop Notice

kodinkat:
- fixes: #2235 
  
  - Identifies infinite loops.
  - Provides record links for identified loop elements.

Screenshot 2023-11-14 at 13 58 24


corsacca:
@kodinkat Shows twice?
image

Issue #2231, opened 2023-10-19
List Ability to set default displayed fields 


corsacca:
I would like to set the default columns that I need so that I don't have to always turn different columns on and off:
<img width="849" alt="image" src="https://github.com/DiscipleTools/disciple-tools-theme/assets/24901539/73817e56-6e63-425d-8ba4-86b3822c39f0">

Current behavior:
- the list columns are governed by a "default fields" list
- selecting and unselecting fields changes the visible columns
- the choice is saved in a cookie, when the page is refreshed the same column list is displayed.
- depending on the browser, the cookie might be deleted after 7 days.

Desired behavior:
- Ability to set change the "default fields" list system wide (for all users). This way everyone has the same field list to start with. We can add this in the D.T Customizations settings

Related issues:
https://github.com/DiscipleTools/disciple-tools-theme/issues/2251: Lists: Save user selected fields to disable in the user settings instead of in a cookie
https://github.com/DiscipleTools/disciple-tools-theme/issues/2125: Lists: Save fields to display on a filter by filter basis.

corsacca:
@kodinkat for this one lets simply add a row with a checkbox in the field modal
"default in list table": [ ] 
that sets `'show_in_table' => true,` or false on the field

<img width="720" alt="image" src="https://github.com/DiscipleTools/disciple-tools-theme/assets/24901539/b8b038b2-177a-4bca-b0c6-98bf1e5e43a3">

PR #2263, opened 2023-11-15
#2231 - List Ability To Set Default Displayed Fields [ Simplified ]

kodinkat:
- fixes: #2231 

Screenshot 2023-11-15 at 12 33 43


corsacca:
@kodinkat if the field already has a numerical value, don't overwrite it with a true. 
This way we keep the existing order.

corsacca:
Thanks @kodinkat!

Issue #2262, opened 2023-11-15
Genmap Hide tree


corsacca:
Clicking "hide" hides the child nodes. This works as expected.
Clicking "hide" again to unhide nodes only unhides the direct children, not the whole tree.

Fix: clicking "hide" again would show the whole child tree.
<img width="1404" alt="image" src="https://github.com/DiscipleTools/disciple-tools-theme/assets/24901539/020eefdb-2764-45d2-a222-0f22db57d31c">

PR #2264, opened 2023-11-15
#2262 - Genmap Hide Tree

kodinkat:
- fixes: #2262


corsacca:
Nice work @kodinkat !

PR #2265, opened 2023-11-16
ML#127 - Updated Magic Link Submit Workflow

kodinkat:


corsacca:
@kodinkat 
I'd vote for not waiting 1500 before starting the refresh.

We could display the toast (as long as we want).

This seem to work fine:
image
Or can we remove the callback altogether?

corsacca:
Thanks @kodinkat

Issue #2266, opened 2023-11-16
Limit new custom post types to 20 chars in Customizations (D.T)


corsacca:
![image](https://github.com/DiscipleTools/disciple-tools-theme/assets/24901539/021be523-6e92-45a4-b1bc-e243214e0a9b)

PR #2267, opened 2023-11-16
#2266 - Limit Custom Post Type Key Characters

kodinkat:
- fixes: #2266 

Screenshot 2023-11-16 at 12 57 48


corsacca:
Thanks @kodinkat !

Issue #2269, opened 2023-11-20
Link field extra line creation


corsacca:
When updating a link field multiple values get created.
Steps:
- add value, 
- click out of the field to save, 
- Edit the value again
- Refresh

You will see 2 fields instead of one.

https://github.com/DiscipleTools/disciple-tools-theme/assets/24901539/ff744d71-7a5a-4da7-acd2-06ea0b393025

PR #2271, opened 2023-11-20
#2269 - Link Field Extra Line Creation

kodinkat:
- fixes: #2269

### DIFF
--- a/dt-assets/js/details.js
+++ b/dt-assets/js/details.js
@@ -96,7 +96,7 @@ jQuery(document).ready(function($) {
       $(`#${fieldKey}-spinner`).removeClass('active')
       post = newPost
 
-      // Safeguard against immediate update failures.
+      // Make sure a key exists for the new link field.
       if ( post && post[fieldKey] && post[fieldKey].length > 0 ) {
         let updated_values = post[fieldKey].filter((option) => {
           return (option['type'] === type) && (option['value'] === value);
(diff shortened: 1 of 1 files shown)


dt-assets/js/details.js
```diff
@@ -94,6 +95,19 @@ jQuery(document).ready(function($) {
     rest_api.update_post(post_type, post_id, { [fieldKey]: fieldValues }).then((newPost)=>{
       $(`#${fieldKey}-spinner`).removeClass('active')
       post = newPost
+
+      // Safeguard against immediate update failures.
```
corsacca, line 99: @kodinkat can we clarify these comments
what is an update failure?
Is this just `//make sure a key exists for the new link field?`

dt-assets/js/details.js
```diff
@@ -94,6 +95,19 @@ jQuery(document).ready(function($) {
...
+      if ( post && post[fieldKey] && post[fieldKey].length > 0 ) {
+        let updated_values = post[fieldKey].filter((option) => {
+          return (option['type'] === type) && (option['value'] === value);
+        });
+
+        // Assuming no duplicate values; process first and only filtered hit.
```
corsacca, line 105: @kodinkat Is this: add missing keys to the the new link field inputs?

kodinkat: Flow updates the meta-id of the link input which generated the event, with newly created meta-id. This ensures any immediate updates, are assigned to correct link input and not to a new/duplicated input field.

PR #2272, opened 2023-11-20
Firebase Login: style background to match normal login

cairocoder01:
- Change background to same gray as standard WP login
- Remove H1 heading that doesn't exist on standard WP login
- Remove "Home" link at bottom of login (login page is the home page if you're not logged in)
- Align "Remember Me" and "Login" button on same line, like standard WP login
- Update URLs sent to Firebase for TOS and Privacy Policy

### DIFF
--- a/dt-login/login-page.php
+++ b/dt-login/login-page.php
@@ -48,7 +48,6 @@ class Disciple_Tools_Login_Base extends DT_Login_Page_Base
         ?>
         <style>
             body { background: #f0f0f1; }
-            #login > .grid-x > .callout h1 { display: none; }
             .login-remember { float: left; }
             .login-submit { float: right; }
             #firebaseui-auth-container { clear: both; }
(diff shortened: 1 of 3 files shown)


corsacca:
Changes. New on left:

image

@cairocoder01 

> Change background to same gray as standard WP login

👍 

> Remove H1 heading that doesn't exist on standard WP login

We'd like to at least keep the heading when user registration is enabled and the user clicks "Register". 

> Remove "Home" link at bottom of login (login page is the home page if you're not logged in)

We want this button for sites where a porch/landing page is used like prayertools: https://ramadandemo.pray4movement.org/. The best solution moving forward might be to make a setting to enable the home button.

> Align "Remember Me" and "Login" button on same line, like standard WP login

Awesome

> Update URLs sent to Firebase for TOS and Privacy Policy

Thanks!

cairocoder01:
@corsacca That makes sense for those two items. I've reverted those changes. It might be nice to have configuration options for both of them. For the sake of time, I didn't do that in this PR.

squigglybob (approved):
All looks good to me.

I think we will need a filter/shortcode args for adjusting the tosUrl and privacyPolicyUrl
I'll add that as another issue

squigglybob:
Thanks @cairocoder01 🎉

PR #2277, opened 2023-11-23
Fix/dt login security fixes

### LINKED ISSUE #2193: Small SSO security fixes
## On sso-login settings tab tab-sso-login.php

- [x] hide firebase keys for non super admins

## In custom login page login-template.php

- [x] remove registered css tag so that there is no link to DT in the sources section of the page.
- [x] echo out dt css into custom login page

squigglybob:
resolves #2193

### DIFF
--- a/dt-login/pages/base.php
+++ b/dt-login/pages/base.php
@@ -30,11 +30,11 @@ class DT_Login_Page_Base
     }
     public function header_style(){
         ?>
+
         <style>
-            body {
-                background: white;
-            }
+            <?php require __DIR__ . '/../../dt-assets/build/css/style.min.css' ?>
         </style>
+
         <?php
     }
     public function _browser_tab_title( $title ){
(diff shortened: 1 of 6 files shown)


corsacca:
@squigglybob 
image

Did this first message show for you?

squigglybob:
> @squigglybob image
> 
> Did this first message show for you?

ooh, I don't know if I saw/noticed that one

squigglybob:
> Are we ok with obfuscation vs not being able to discover the value? (i think so, not a huge issue) image

Hmm, that's a good point, it doesn't change the `.value` of the input.
I can make it put in some fake data as the value as well

dt-login/pages/base.php
```diff
@@ -30,11 +30,11 @@ public function _header(){
     }
     public function header_style(){
         ?>
+
         <style>
-            body {
-                background: white;
-            }
+            <?php require __DIR__ . '/../../dt-assets/build/css/style.min.css' ?>
```
corsacca, line 35: @squigglybob 
this includes ALL of the 30 kB of D.T css in the header.
It is just for this UI, right:
image

squigglybob: I could sift the CSS, for what is needed for this/ write some CSS to make it look like this 😆

squigglybob: I've created a stripped down css for the login page, that has no dependencies on foundation either.

I wasn't able to test the page where you create a new password after receiving the lost password email. 😬 
it was taking me to the wp-login new password page instead

corsacca:
Thank you @squigglybob !

PR #2282, opened 2023-11-29
Custom login rest

corsacca:
The goal is to only load SSO settings when needed.


squigglybob:
This may be a seperate thing to your PR here, but I can't register 😱 it says missing email or password even though I have given them correctly

Issue #1643, opened 2022-05-03
Sort dropdown icons


corsacca:
This sort dropdown could use some icons.
Suggestions?
![image](https://user-images.githubusercontent.com/24901539/166421944-a2cdfe28-90f7-445d-9512-d0e79faa8560.png)

corsacca:
Look for icons in https://materialdesignicons.com/

EthanW96:
@corsacca should this issue be closed now that #2286 has been merged?

corsacca:
Yes! Thanks.

PR #2286, opened 2023-12-03
#1643 - Sort dropdown icons

EthanW96:
#1643 

Hi! I added icons from [https://pictogrammers.com/library/mdi/](https://pictogrammers.com/library/mdi/) which is the rebranded [https://materialdesignicons.com/](https://materialdesignicons.com/). No apparent changes to licensing (continues to be [open-source](https://pictogrammers.com/docs/general/license/)) 

I feel good about the calendar icons but I'm not certain about representing "modified" with the clock icon. As I was looking through the icons that was my instinct but I'm open to switching it out for something else. 

Also, the arrows seem backwards, but the desc/asc match up with the names of the icons so I left it. I can reverse the icons if that's preferred.


corsacca:
Thanks @EthanW96 
I'm sorry, I missed including some context. We have mdi already included in the theme.
So we should be able to do something like `<i class="mdi mdi-help-circle"></i>` without needing to download and include the asset. 

Good to know about the rebrand.

corsacca:
image
I think the icons and the direction makes sense. we can easily switch them around if we decide otherwise :D

corsacca:
Thanks @EthanW96

Issue #1878, opened 2022-11-29
UI for the boolean field


corsacca:
Yes / No? Checkbox?

corsacca:
Add boolean field to new field options

PR #2288, opened 2023-12-04
#1878 - UI for the boolean field

EthanW96:
As I mention in the issue we have some additional features that might be helpful.

- Ability to enter specific words into the "Yes" and "No" fields. Sometimes different words convey meaning better for boolean fields. I'm also wondering if a rework like this could help with language translation options?
- Ability to set a default instead of defaulted to "No".
- Ability to pick a checkbox only (no language) or Yes/No dropdown

Here is the result of the current commit:

image

### DIFF
--- a/dt-assets/js/modular-list.js
+++ b/dt-assets/js/modular-list.js
@@ -797,7 +797,6 @@
           let field_value = window.lodash.get( record, field_key, false )
           if ( field_key !== "favorite" && field_settings.type === "boolean" ) {
             field_value = window.lodash.get( record, field_key )
-            console.log(field_key + ' ' + field_value)
           }
 
           /* breadcrumb: new-field-type Display field in table */
--- a/dt-core/global-functions.php
+++ b/dt-core/global-functions.php
@@ -604,8 +604,7 @@ if ( ! defined( 'DT_FUNCTIONS_READY' ) ){
                 if ( isset( $post[$field_key] ) ) {
                     $selected = !empty( $post[$field_key] ) ? 'selected' : '';
                 } else {
-                    // Set a default value here, for example:
-                    $selected = $fields[$field_key]['default'] === true ? 'selected' :''; // Assuming 'Yes' is the default value
+                    $selected = $fields[$field_key]['default'] === true ? 'selected' :'';
                 }
                 ?>
                 <select class="select-field" id="<?php echo esc_html( $display_field_id ); ?>" <?php echo esc_html( $disabled ); ?>>


EthanW96:
@corsacca nice that's awesome! Should I set this Boolean field up as a component somewhere? Or is this issue on hold till the refactor is complete?

dt-core/admin/menu/tabs/tab-custom-fields.php
```diff
@@ -1538,6 +1540,15 @@ private function process_add_field( $post_submission ){
                     'customizable' => 'all',
                     'private' => $field_private,
                 ];
+            } elseif ( $field_type === 'boolean' ){
+                $new_field = [
+                    'name' => $post_submission['new_field_name'],
+                    'default' => [],
```
corsacca, line 1546: @EthanW96 default here should probably be `false`

EthanW96: Cool, I've adjusted this be marked as false by default.

EthanW96:
All sounds good to me. I did wonder how the display of the boolean field worked so seamlessly without additional code written on my part. I'm still learning where everything is located in the codebase. It's good to know that https://github.com/DiscipleTools/disciple-tools-theme/commit/8e1ab90bc14a43e62d7f8400c1e13c24eada90d0 exists.

EthanW96:
@corsacca I realized while looking at this line:
https://github.com/DiscipleTools/disciple-tools-theme/blob/8e1ab90bc14a43e62d7f8400c1e13c24eada90d0/dt-core/global-functions.php#L599
The display of the boolean field doesn't actually handle a default (I could enter anything in the review you pointed out above) so https://github.com/DiscipleTools/disciple-tools-theme/pull/2288/commits/54b2064eacf536e9556da2d69117aa34d6ac255b adds that functionality. When the default is being set it must be a strict `true` or `false`.

corsacca:
Implementing the default might be a bit tricky.

I added this field in contacts/base-setup.php

```
    $fields['bool_test'] = [
        'name' => __( 'Boolean Test', 'disciple_tools' ),
        'type' => 'boolean',
        'default' => true,
        'tile' => 'status',
    ];
```
with default is true.

It shows the field with YES selected:
image

But, when we change the value to NO and refresh the page it show YES.
This is because the `!empty()`  is figuring out the selected with show false, so then the "default" is chosen.

EthanW96:
Good call out. I've made some adjustments to that logic and to the modular list so boolean fields that are defaulted to true show correctly. Are there other places in D.T. that boolean fields may appear that need additional logic for defaults?

Issue #2292, opened 2023-12-11
View to se scheduled jobs.


corsacca:
In WP Admin > Utilities. Next to email logs.

Create a page to view the queue_jobs table.

This will help the admin know if there are pending jobs (likes emails to be sent).

Show the overall count, the job names, attempts. Don't show the job contents.

Create a button to trigger processing the jobs from this admin page. We do this in the facebook plugin.

PR #2305, opened 2023-12-16
#2292 - View to see scheduled jobs

EthanW96:
@corsacca Here is my first pass at #2292 for seeing scheduled jobs.

- [x] Show the overall count
- [x] Show job names
- [x] Show job attempts 
- [x] Don't show the job contents
- [x] Button to trigger processing the jobs

Extra
- [x] Show cron schedule to user
Next to the button that triggers processing jobs I added some text to indicate the wp_queue cron schedule timing. You can test this by adjusting the interval default here:  
https://github.com/DiscipleTools/disciple-tools-theme/blob/f397541a1871d7d7d5c3864eed00d45d913126c6/dt-core/libraries/wp-queue/WP_Queue/class-queue.php#L62
I tried to add a "countdown" for when the cron schedule would trigger the next processing but [wp_next_scheduled](https://developer.wordpress.org/reference/functions/wp_next_scheduled/) didn't appear to give me accurate info and I didn't know where else to turn.

I wasn't positive where to place the JS to trigger the make_admin_request so I guessed. Feel free to let me know it needs to be located somewhere else.

### DIFF
--- a/dt-core/admin/js/dt-utilities-scripts.js
+++ b/dt-core/admin/js/dt-utilities-scripts.js
@@ -55,6 +55,15 @@ jQuery(document).ready(function ($) {
     })
   })
 
+  $('.process-jobs-button').on('click', function() {
+    make_admin_request("GET", 'process_jobs').then(status => {
+        $(`#process-jobs-loading-spinner .loading-spinner`).addClass( "active" )
+        if (status.success === true) {
+            $('.process-jobs-result-text').html('Done!')
+            $(`#process-jobs-loading-spinner .loading-spinner`).removeClass( "active" )
+        }
+    })
+})
   /**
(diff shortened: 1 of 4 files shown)


corsacca:
Fantastic work @EthanW96!
Tables displays the jobs and the "Process Jobs" button cleared the queue.
I like the added row in the Overview tab.

I changed:
> "Every 5 Minutes the queue is processed"
-> The queue is scheduled to be processed: Every 5 Minutes
The queue schedules the cron to run every 5 minutes. But it will depend on the server setup as to how often the cron job process is called. We ask Admins to set up their own cron jobs as the WP default is that cron jobs are only triggered when a user visits a page.

corsacca:
Thanks @EthanW96 !

Issue #2296, opened 2023-12-13
Lists: Sort by decimal number not working:


corsacca:
![image](https://github.com/DiscipleTools/disciple-tools-theme/assets/24901539/70837ef0-acf2-4ac1-a92c-4390dd5f635d)

PR #2306, opened 2023-12-18
#2296 - Lists Sort By Decimal Number Not Working

kodinkat:
- fixes: #2296

### DIFF
--- a/dt-assets/js/modular-list.js
+++ b/dt-assets/js/modular-list.js
@@ -800,7 +800,7 @@
           }
 
           /* breadcrumb: new-field-type Display field in table */
-          if ( field_value ) {
+          if ( field_value !== false ) {
             if (['text', 'textarea', 'number'].includes(field_settings.type)) {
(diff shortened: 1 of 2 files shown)


corsacca:
> DECIMAL(18,4)
@kodinkat what do the 18 and 4 stand for?

kodinkat:
> > DECIMAL(18,4)
> > @kodinkat what do the 18 and 4 stand for?

@corsacca to do with [precision](https://dev.mysql.com/doc/refman/8.0/en/precision-math-decimal-characteristics.html) - number of digits on either side of decimal point.

corsacca:
Is 18 the largest it will go? (why 18 and 4?)

kodinkat:
> Is 18 the largest it will go? (why 18 and 4?)

@corsacca  No reason, just arbitrary numbers... Max on the left side (of decimal point) is 65 and 30 on the right...

Issue #2036, opened 2023-04-13
CSV Export


corsacca:
Move CSV export from https://github.com/DiscipleTools/disciple-tools-list-exports into core
  
##  Current behavior:
  
 Clicking "CSV list" and downloading the file will only download a csv with the 100 showing contacts and only with the selected fields.
  
![image](https://github.com/DiscipleTools/disciple-tools-theme/assets/24901539/b2b0f5fc-fb8e-403d-b4ad-29d847de2313)

![image](https://github.com/DiscipleTools/disciple-tools-theme/assets/24901539/f9f8dc1d-8bda-4959-a45e-982b682480cf)

## We want 
- the csv to include all the record's fields.
- the file to include all the the records that match the filter, instead of just the 100 that are shown.

Design decision:
Download all the records locally and create the csv via js.
or Create the csv in PHP and make the file available.

I think creation the file via js will work better in this case.

## Notes
Let the theme declare the List Exports tile that the plugin can hook into.
Make sure strings are translated.
Let the modal or download button say how many records will be in the csv. (Download 1430 records).
Remove no longer needed sections: "The following fields will be exported:" and "show list".
Make sure all the field types are downloading correctly.

Lets do bbc email list, phone number list and map list after the this PR is merged.

PR #2307, opened 2023-12-21
#2036 - CSV Export Core Integration

kodinkat:
- fixes: #2036 
---
 
Screenshot 2023-12-21 at 11 34 13

Screenshot 2023-12-21 at 11 34 39

Screenshot 2023-12-21 at 11 38 53

Screenshot 2023-12-21 at 11 36 28


corsacca:
Awesome! @kodinkat 

It looks like the current implementation only will export 1000 records.
If the list contains more than 1000, we need it to make more api calls to get the rest.

The list of fields looks rough. I know that is how the plugin shows them. Would you be able to try a different display?
Related, can we make the first csv field be the record ID,  and the second be the name?

Thank you much!

corsacca:
@kodinkat note, you can omit the "fields_to_return" param to return all fields.

corsacca:
@kodinkat for 1111 contacts 4 api calls are being made instead of 3. The last one returns empty. Can you fix that?

corsacca:
@kodinkat 
- Missing the ID part of location exports:
image

- Filter out array fields from exporting to the csv

- Exclude private fields. (a next issue will be to make this a setting that can manually be included).

corsacca:
Looking Good! Thank you @kodinkat !

Issue #750, opened 2019-08-20
Create an email template


corsacca:
Email are currently send as plain text.
Improve emails with an html template.
Code html as to handle mixing Arabic and English text. Direction set in the same way as in the web interface (with dir="auto").

corsacca:
Explore copying template from https://github.com/nlemoine/wp-better-emails

PR #2314, opened 2024-01-09
#750 - Create An Email Template

kodinkat:
- fixes: #750 
---

Screenshot 2024-01-09 at 15 06 43

Screenshot 2024-01-10 at 13 43 22

Screenshot 2024-01-10 at 13 43 43

Screenshot 2024-01-10 at 13 44 12

Screenshot 2024-01-10 at 13 44 35

Screenshot 2024-01-10 at 13 45 27

Screenshot 2024-01-10 at 13 46 26


corsacca:
@kodinkat can we use the already built in `make_clickable()` function ?

Also we need to check that the email is not already an html message. I think we can use $phpmailer->ContentType

corsacca:
Thanks @kodinkat !

Issue #2319, opened 2024-01-11
List Filters: Ability to filter for text and communication channel fields


corsacca:
Set a custom text to search for
Checkbox for "any value"
Checkbox for "no value set" (not set or empty string)

PR #2322, opened 2024-01-15
#2319 - List Filters: Support Text & Comms-Channel Field Filters

kodinkat:
- fixes: #2319 
---
 
Screenshot 2024-01-15 at 17 06 53


corsacca:
@kodinkat Text field filtering api calls don't seem to be working for me working

corsacca:
Todo:
when the user fills out the input field, default to searching fields matching the text:
image

corsacca:
@kodinkat 
The filter label in this case. Can it be adjusted to reflect the filtered text instead of "All".

image

corsacca:
@kodinkat 
image
Is returning all contacts.

Should only return contacts with non empty values for the field.

Let's use for that search `['*']` instead of `-*`. (and update documentation).

corsacca:
@kodinkat 
No sure how i missed this, but 
image

Isn't showing any results? 
image

corsacca:
@kodinkat 
this one isn't working any more:
image

corsacca:
@kodinkat we want to try to move away from lodash.
Would you be willing to replace the 2 lodash functions ?

corsacca:
This is looking fantastic!

dt-assets/js/modular-list.js
```diff
@@ -1081,6 +1081,42 @@
           date.end = end
         }
         search_query.push({[field]: date})
+      } else if ( type === "text" || type === "communication_channel" ){
+        let value = '-*';
```
corsacca, line 1085: @kodinkat `-*` isn't a valid format, lets change the default

corsacca:
Thank you @kodinkat!

PR #2324, opened 2024-01-16
feat: add redirect_to attribute to login shortcode

squigglybob:
@corsacca
This is needed to fix being able to pass through a custom redirect_to param to overrule the default one in the SSO settings

This fixes Zume where after SSO login/register it would go only to the dashboard rather than to the redirect_to query param in the url when they logged in.

### DIFF
--- a/dt-login/login-shortcodes.php
+++ b/dt-login/login-shortcodes.php
@@ -12,11 +12,14 @@ add_shortcode( 'dt_firebase_login_ui', 'dt_firebase_login_ui' );
 function dt_firebase_login_ui( $atts ) {
 
     $default_lang = 'en';
+    $default_redirect_to = DT_Login_Fields::get( 'redirect_url' );
     $atts = shortcode_atts( [
         'lang_code' => $default_lang,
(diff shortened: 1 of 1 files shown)


dt-login/login-shortcodes.php
```diff
@@ -12,11 +12,14 @@
...
 
     $default_lang = 'en';
+    $default_redirect_to = DT_Login_Fields::get( 'redirect_url' );
     $atts = shortcode_atts( [
         'lang_code' => $default_lang,
+        'redirect_to' => $default_redirect_to,
     ], $atts );
 
     $lang_code = $atts['lang_code'];
+    $redirect_to = $lang_code;
```
corsacca, line 22: @squigglybob i'm not sure i follow the logic here.
The redirect will *always* be the lang code?

squigglybob: 😆 doh. That is not what's it's meant to be, and I have no idea how it was even working!!!

corsacca:
THanks @squigglybob!

Issue #2317, opened 2024-01-11
Activity Log field key


corsacca:
Add a field key to the activity log table.
Especially useful for the communication channel and links fields.
Write a migration to add the field key to the existing activity records.

PR #2326, opened 2024-01-22
#2317 - Activity Log Field Key

kodinkat:
- fixes: #2317


corsacca:
Thanks @kodinkat 

Please make sure the field_key gets saved for new activity as well.
Double check that all fields type were included in the migration. It looks like connection fields weren't.


`contact_address_95f_details` just resolves to `details`
Lets only add valid fields in the field_key, so lets leave it empty for now

corsacca:
Going to close this for now in favor of using the object_subtype

PR #2334, opened 2024-01-24
Non-Latin CSV Export bug fix

### LINKED ISSUE #2333: List Export Non-latin text
When exporting a list with non-latin text (Arabic, Hebrew, etc.) the names are shown as encoded in the CSV file.

This may be related to to #2332

micahmills:
removes escaping the csv so that non-latin languages work in the csv export.

We need to determine if this opens us up to any security concerns.

Resolves #2333


kodinkat (changes requested):

micahmills:
I initially tried the shared function and it didn't work but that was when I was trying to escape the whole CSV string instead of each field. I have updated it to use the escapeHTML function in shared functions.

micahmills:
This should be good to go now

kodinkat (changes requested):
@micahmills I'm no longer seeing the previous exceptions; thanks for that.. However, exported csv file, no longer contains double-quotes or new-line carriage-returns?

Screenshot 2024-01-26 at 10 51 55

micahmills:
@corsacca and @kodinkat can I ask what may be an obvious question. Do we need to escape this data at all? My gut says yes, but I can’t think of an attack vector taking this into a CSV file. Can you?

corsacca:
:D I thought this PR was adding escaping

micahmills:
In master now it runs the whole string through the (probably deprecated) .escape function. That  encodes non-latin text and if we looked closely probably breaks on line breaks and edge cases like apostrophes and commas. 

This was an attempt to fix the non-Latin character thing but keep the escaping. 

So the question is do we need to escape it at all?

micahmills:
Ok the current changes removes escaping. If there are double quotes it caused a problem so I found that you can double double quote (https://stackoverflow.com/a/769675). So there is a check on each item to see if there is a double quote and then replaces it with a double double quote (quadruple quote, is that a thing?)

I am not sure about carriage returns, or even where that would show up in the lists.

corsacca:
Assuming escaping isn't needed, that would simplify things a lot.
We'll want to double check.
I plan an giving it a test with https://github.com/DiscipleTools/xss-attack

corsacca:
@micahmills I don't think each cell needs to be escaped.
i think we could just go with
`csv_download_link.href = "data:text/csv;charset=utf-8," + csv;`
Does this pass tests for non latin scripts?

corsacca:
My on field quick test says yes:
image

instead of:
image

micahmills:
That works for non-Latin but if you have a name with a double quote it will fail. I had something like "Tom and it will break.

micahmills:
The console error was because some items were coming in as the boolean value of true instead of a string. I added code that casts the item variable as a string before running the include check.

micahmills:
If we just remove the escaping of double quotes in each item an unclosed double quote in a name or field will break the CSV ( in a name like `"tom`)

corsacca:
I'm starting to see. Here is weird behavior:

image

Without the double quotes escaping Tom is missing a quotes and the next row the quotes is at the end:

image

With the double quotes escaping something weird happens with `"tom"`
image

corsacca:
@micahmills 
`escapeditem = item.replaceAll('"', '""')` instead of 'escapeditem = item.replace('"', '""')'
seems to do it for me. Can you see any issues with that?

micahmills:
Good catch replaceAll is what we want not replace.

corsacca:
Thanks @micahmills !

PR #2347, opened 2024-01-29
#2341 - Update Activity Log Object_Subtype Column To Be Used As Default Field Key

kodinkat:
- fixes: #2341


corsacca:
@kodinkat 1000 contacts with emails, phone numbers etc with generate 1000s of API calls in this migration.
This will likely timeout on a large database before it is able to finish

Can we do something like
Update object_subtype = " object_subtype - 4 chars "
Where field_type = 'communication_channel'
AND action = field_update
AND object_subtype NOT LIKE %_details
AND object_subtype LIKE contact_%_%

And another one for links

corsacca:
Nice work @kodinkat 
Edge issue if the migration is run again (this can happen)
First migration:
image
Correctly becomes
image

But then on the second migration:
image

corsacca:
Thank you @kodinkat !

Issue #2337, opened 2024-01-26
Escaped chars


corsacca:
In filter name:
![image](https://github.com/DiscipleTools/disciple-tools-theme/assets/24901539/2a161a3d-d3c6-4664-91b3-25e2469c900c)

PR #2350, opened 2024-01-30
#2337 - Support Escaped HTML Characters

kodinkat:
- fixes: #2337


corsacca:
@kodinkat 
It looks like the issue is from double escaping.

Unescaping could have bad consequences.

Issue #2340, opened 2024-01-26
List Exports > Map


corsacca:
Add sidebar listing record viewed in the current zoomed in map

![image](https://github.com/DiscipleTools/disciple-tools-theme/assets/24901539/7faa2a41-402c-4ba8-8f91-1a9f4dc7598a)

corsacca:
Next issue:
open list filtered to current viewed records

corsacca:
Copy code from https://github.com/ZumeProject/zume-public-heatmaps

PR #2359, opened 2024-02-07
#2340 -  List Exports Map

kodinkat:
- fixes: #2332 
- fixes: #2340 
---

Screenshot 2024-02-07 at 10 08 11


corsacca:
The filtered list is meant to pick up on the bounds of the map instead of where the mouse is hovering.

corsacca:
@kodinkat 
I'm still getting the double scroll bars.
image

Todo:
Can we use the existing modals?
#modal-large for the CSV and list exports
#modal-full for the map

That should deal with one of the scroll bars. 

For the second: Only scroll the right panel instead of the whole page when many records are present.

corsacca:
@kodinkat add loading spinner on BCC Email List and Phone list
image

corsacca:
Thanks for the Map updates!

Issue #2366, opened 2024-02-09
Disable autofill on new user creation


corsacca:
Keep last pass and other password managers from autofilling 
Email, Username and password (or any field) when creating a new user.

<img width="696" alt="image" src="https://github.com/DiscipleTools/disciple-tools-theme/assets/24901539/3d190fc8-94c6-48aa-9439-74340accbf07">

PR #2367, opened 2024-02-09
#2366 - Disabled Autocomplete

kodinkat:
- fixes: #2366


corsacca:
@kodinkat this is good add and we'll keep it.
But password managers like lastpass do not respect Autocomplete.

I think the best option might be to change the name/ids of the fields.
Please also test before/after.

PR #2386, opened 2024-02-22
feat: add filter to override dt_custom_dir_attr

squigglybob:
Where a porch is using it's own language switching method, this allows that porch to prevent the html attrs from being switched to match the WP user locale


squigglybob:
@corsacca I've added this override so that we can control the html lang attributes for zume better.
The dt_custom_dir_attr function was overriding the ones created for the zume ui language selection with the user locale dir attributes.

PR #2390, opened 2024-03-05
ML#136 - User Magic Links

kodinkat:
Screenshot 2024-03-05 at 15 50 24

Screenshot 2024-03-05 at 15 49 08


corsacca (changes requested):
Thanks @kodinkat
On the user-management records:
Display only the magic links of type **user**. I'm seeing the templates show up there too.

corsacca:
@kodinkat 
On a new user.
None of these magic links work
image

The first one is just: `http://domain.lwp/smart_links/user_contacts_updates/`

kodinkat:
@corsacca 

Screenshot 2024-03-15 at 14 48 44

dt-users/users-endpoints.php
```diff
@@ -168,8 +168,8 @@ public function switch_preference( WP_REST_Request $request ) {
 
     public function app_switch( WP_REST_Request $request ) {
         $params = $request->get_params();
-        $user_id = get_current_user_id();
-        if ( isset( $params['app_key'] ) && ! empty( $params['app_key'] ) && $user_id ) {
+        $user_id = !empty( $params['user_id'] ) ? $params['user_id'] : get_current_user_id();
```
corsacca, line 171: @kodinkat what is keep a user from enabling or disabling other users apps?
(implement permissions)

dt-reports/magic-url-setup.php
```diff
@@ -82,6 +82,11 @@ private function add_app_row( $post_type, $app ){
         } else {
             $key = dt_create_unique_key();
             update_post_meta( get_the_ID(), $meta_key, $key );
+
+            // Ensure user meta options are also kept in sync.
+            if ( !empty( $record['corresponds_to_user'] ) ) {
```
corsacca, line 87: @kodinkat, this force enables all the apps.

corsacca:
Original issue: https://github.com/DiscipleTools/disciple-tools-bulk-magic-link-sender/issues/136

corsacca:
@kodinkat these magic links are not only available to the adminintrators and D.T admins as requested:
image

(also check the user management side)

corsacca (changes requested):

dt-reports/magic-url-setup.php
kodinkat: > @kodinkat, this force enables all the apps.

@corsacca - ensures whenever a new key is created and updated at the post meta level; it's also reflected at the user meta level too; which is queried within user management view.

corsacca: @kodinkat auto creating the key enables the app for the user.
We don't want this to happen automatically, this is why the user has the toggles.
Can you think of a good way around this issue?
In the end i'd rather remove them from this list then have them be auto enabled.

kodinkat: @corsacca let me know how the latest tweaks holds up for you; which now ensures (in addition to correct permissions) there is a valid app key set in order to be displayed within record's frontend app list.

App keys are now solely handled within the user management section.

dt-users/users-endpoints.php
corsacca: @kodinkat multipliers can update their own preferences, but should not be able to update someone else's.
Here the multiplier is blocked from this update.

corsacca:
@kodinkat 
These are displaying activate or not based on if the magic link key is saved on the contact instead of the user.
image

See
image

dt-users/users-endpoints.php
corsacca: @kodinkat sorry, i wasn't clear here. We need to think through all the users here, not just the multipliers.

Simply allow the $user_id to be set manually here if the user can `manage_dt`, otherwise return an error if the $user_id is not the current logged in user

corsacca (approved):

corsacca:
Thanks @kodinkat !!

PR #2392, opened 2024-03-06
Filter Count Alignment

kodinkat:
Screenshot 2024-03-06 at 14 27 06


corsacca:
@kodinkat can this be done with css and hiding the overflow instead of char length?
Char length won't do as well on different screen sizes.

corsacca:
@kodinkat why the added extra space?
image

corsacca:
@kodinkat the -7px margin feels hacky.

Try making the .list-view {
  display: flex;
  column-gay: 0.2rem
  }
and add `flex-grow: 1` on the middle element (.list-view__text)

corsacca:
Thank you @kodinkat !

PR #2403, opened 2024-03-13
#2399 - Fix Workflow Conditions Not Firing For Key Select Field Types

kodinkat:
- fixes: #2399


corsacca:
@kodinkat

 What happens when step 2 has not value set?
 
image

Issue #2395, opened 2024-03-11
Ability to upload media to be stored securely.


corsacca:
For profile pictures, voice messages, pictures in comments, etc.

Ideally media would not be stored on wordpress. (we think this is not secure).
Media would be stored in S3, or backblaze.

cairocoder01:
Another cloud storage provider to consider is Cloudinary. I think they have a pretty decent free plan that would let a lot of people use it who don't have a lot of resources.

Issue #2001, opened 2023-03-24
Ability to record voice messages on a contact record


corsacca:
![image](https://user-images.githubusercontent.com/24901539/227463512-60577a52-4969-49d2-9248-3ebc44601ed2.png)

Bonus: auto transcripts.

corsacca:
Notes:
One of the big issues is storage of pictures and audio.
The uploads folder does not feel very secure and most hosting don't give a lot of space.
This highlights a need for media to be able to be stored in s3 or backblase or similar. with the links to the audio stored in D.T

cairocoder01:
To support various cloud storage providers, it would be smart to setup plenty of hooks in order to allow custom plugins to support other providers. I did this with my data reporting plugin (https://github.com/cairocoder01/disciple-tools-data-reporting), so that might be worth looking at. It provides basic HTTP functionality and then there are 2 plugins that create providers for [Google](https://github.com/cairocoder01/disciple-tools-data-reporting-provider-gcp) and [Azure](https://github.com/cairocoder01/disciple-tools-data-reporting-provider-azure).

corsacca:
Yes! hook would be great so we can build this in collaboration with this feature: https://github.com/DiscipleTools/disciple-tools-theme/issues/2395

PR #2411, opened 2024-03-15
Add audio playback to comments (via meta data)

cairocoder01:
This is related to #2001 and #2395, but doesn't include the ability to record/upload audio. 

This looks for comment meta data with key `audio_url` and displays an audio player below the comment if so. It expects the URL is fully-qualified, and it can support multiple audio files per comment (though this is probably unlikely).

The REST API is also updated for create/read/update to be able to retrieve/add meta data. This will include existing reaction meta data if it exists. This allows external systems that send data into DT via API to specify any audio content that they may have.

image


corsacca:
Nice work @cairocoder01 !

Added a comment via;
`makeRequestOnPosts('POST', 'contacts/1378/comments', { "comment":"a", "comment_type":"comment", "meta": {"audio_url": "https://www2.cs.uic.edu/~i101/SoundFiles/BabyElephantWalk60.wav"}})`

And 
`makeRequestOnPosts('POST', 'contacts/1378/comments', { "comment":"a", "comment_type":"comment", "meta": {"audio_url": ["https://www2.cs.uic.edu/~i101/SoundFiles/BabyElephantWalk60.wav", "https://www2.cs.uic.edu/~i101/SoundFiles/Fanfare60.wav"]}})`

Would be nice to add these to the docs.

Comments list API returns:
```
          "comment_reactions": {
                "audio_url": [
                    "https:\/\/www2.cs.uic.edu\/~i101\/SoundFiles\/BabyElephantWalk60.wav",
                    "https:\/\/www2.cs.uic.edu\/~i101\/SoundFiles\/Fanfare60.wav"
                ]
            },
            "comment_meta": {
                "audio_url": [
                    "https:\/\/www2.cs.uic.edu\/~i101\/SoundFiles\/BabyElephantWalk60.wav",
                    "https:\/\/www2.cs.uic.edu\/~i101\/SoundFiles\/Fanfare60.wav"
                ]
            }
```
We'll need to limit reactions to just reactions. Could we just have "comment_meta" and let the front end parse out the reactions?

Wondering if comment_meta should return the meta_id, so we can later directly update or delete comment meta via a comment meta api. 
```
            "comment_meta": {
                "audio_url": [
                    { id:123, value:"https:\/\/www2.cs.uic.edu\/~i101\/SoundFiles\/BabyElephantWalk60.wav"},
                    { id:124, value:"https:\/\/www2.cs.uic.edu\/~i101\/SoundFiles\/Fanfare60.wav"}
                ]
            }
```

Tested an update with:
`makeRequestOnPosts('POST', 'contacts/1378/comments/1085', { "comment": "b","meta": {"audio_url": ["https://www2.cs.uic.edu/~i101/SoundFiles/a.wav", "https://www2.cs.uic.edu/~i101/SoundFiles/b.wav"]}})`

cairocoder01:
@corsacca Great thoughts. I updated the `get_post_comments` API to return the meta IDs and also properly split out regular meta data from reactions.

I was going to do similar for add/update endpoints, but the WP function `get_comment_meta()` doesn't actually return the meta IDs. So I'm not sure if we want to do extra work to grab those meta IDs with a custom query or not.

I'll submit a PR for the documentation site.

corsacca:
This looks great @cairocoder01! Thank you.

> I was going to do similar for add/update endpoints, but the WP function get_comment_meta() doesn't actually return the meta IDs. So I'm not sure if we want to do extra work to grab those meta IDs with a custom query or not.

Lets tackle this one when when it is needed.

Issue #2401, opened 2024-03-12
Zoomed map persist


corsacca:
All contacts > Open in Map > Zoom In > Open Zoomed Map Records.
Refresh the page and the current view is lost.

Other custom filters the current view is saved on refresh

corsacca:
Thanks @kodinkat!
I think we discussed the query labels also being updated:
![image](https://github.com/DiscipleTools/disciple-tools-theme/assets/24901539/fa28f2d8-cd05-40c3-b337-0eb897a73b94)

Currently they stay the same.

PR #2413, opened 2024-03-18
#2401 - Persist Zoomed Map Query Settings

kodinkat:
- fixes: #2401


corsacca:
Thanks @kodinkat!
I think we discussed the query labels also being updated:
image

Currently they stay the same.

corsacca:
Thank you @kodinkat!

Issue #2408, opened 2024-03-15
WP Admin Settings Import -> Select all tiles and fields


corsacca:
Button to select all tiles and fields: 
![image](https://github.com/DiscipleTools/disciple-tools-theme/assets/24901539/2a61cd9f-cee4-4aa6-9e70-3a9463d4f211)

PR #2414, opened 2024-03-19
#2408 - Admin Import - Select All Tiles & Fields

kodinkat:
- fixes: #2408 
---

Screenshot 2024-03-19 at 11 39 19


corsacca:
Masterful @kodinkat!

PR #2419, opened 2024-03-23
coding exercise answers

CptHappyHands:
Looks like my linter changed a lot of the " to '  and vice versa. But besides that figure I would provide an explanation on what I did: 

I only saw 3 lodash functions in that particular file.  

In this one, declared a new const and made it equal to everything but .dropdown from the original. This allowed me to declare another const, value, and check if .dropdown exists. If it does, great, it will equal .value, otherwise it'll equal 'all'.
```
                // return window.lodash.get(window.Typeahead['.js-typeahead-location_grid'].filters.dropdown, 'value', 'all')
                const { dropdown } =
                  window.Typeahead[".js-typeahead-location_grid"].filters;
                const value = dropdown?.value ?? "all";
                return value;
```

 
In this case, I used the .sort function. This will check the start_date property in descending order.
```
  // list = window.lodash.orderBy( list, [ "start_date" ], "desc")
  list.sort((a, b) => new Date(b.start_date) - new Date(a.start_date));
```

Here, I declared a new const, color, get the same info that the lodash was getting previously, and use it to set the background-color. 

```
    const color =
      window.wpApiSettingsPage?.workload_status_options?.[name]?.color;
    selected.css("background-color", color);
    // selected.css('background-color', window.lodash.get(window.wpApiSettingsPage, `workload_status_options.${name}.color`))
```

### DIFF
--- a/dt-assets/js/settings.js
+++ b/dt-assets/js/settings.js
@@ -1,56 +1,56 @@
-jQuery(document).ready(function() {
-  window.current_user_lookup = window.wpApiSettingsPage.current_user_id
-  load_locations()
+jQuery(document).ready(function () {
+  window.current_user_lookup = window.wpApiSettingsPage.current_user_id;
+  load_locations();
+});
 
-})
-
-function app_switch (app_key = null) {
-  let a = jQuery('#app_link_' + app_key)
-  a.empty().html(`<span class="loading-spinner active"></span>`)
-  window.makeRequest('post', 'users/app_switch', { app_key })
-    .done(function(data) {
-      if ('removed' === data) {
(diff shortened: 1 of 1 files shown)


corsacca:
This looks great @CptHappyHands, thank you!
Would you be willing to resubmit the PR without the linting changes?

CptHappyHands:
> This looks great @CptHappyHands, thank you! Would you be willing to resubmit the PR without the linting changes?

Thanks for checking that. I believe I got all of the linting changes reverted. Let me know if there is anything else I missed

corsacca:
Thank you @CptHappyHands !

PR #2421, opened 2024-03-26
Fix to new contact error(title-needed)

gp-birender:
The error(title needed) showed up when we added a name to the contact, as we are using the dt-text to render the name on this form.. we added a few lines here then it started working to add contact but the selected values for edit form elements are not coming..we also need to check that.  @micahmills

### DIFF
--- a/dt-assets/js/new-record.js
+++ b/dt-assets/js/new-record.js
@@ -91,10 +91,11 @@ jQuery(function($) {
   })
 
   /* breadcrumb: new-field-type Add the new link type data to the new_post array */
-  $(".js-create-post").on("submit", function() {
+  $(".js-create-post").on("submit", function (e) {
     $(".js-create-post-button").attr("disabled", true).addClass("loading");
-    // new_post.title = $(".js-create-post input[name=title]").val()
-    new_post.name = $("#name").val();
+
+    new_post.title = $(".js-create-post [name=name]").val();
+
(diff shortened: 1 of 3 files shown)


corsacca:
@gp-birender thanks for the PR.
To help me understand. Did you create a new field called "name" ?
"name" is already a default field for all post types.

gp-birender:
@corsacca - No, we were using jQuery code `new_post.title = $(".js-create-post input[name=title]").val()`  to get the value but it's not getting posted and we were getting **' title needed error'** while posting. so, we have checked the integrated web component named(dt-text) for the name field to get the value out of that so that the form gets submitted.

gp-birender:
Today, I have made some changes to the previous commit by getting name from `new_post.title = $(".js-create-post [name=name]").val()`

gp-birender:
@micahmills - earlier we made changes in the DT theme JQuery code..and we will now make changes in the LIt code.

micahmills:
This seems to be working. I would like to move away from the jQuery code creating the new post, but it works for now. 

I was seeing some PHP errors and was able to fix it by changing line 136 of template-new-post.php to   `render_field_for_display( $field_key, $post_settings['fields'], [ 'post_type' => $dt_post_type ] );`

micahmills:
I think we resolved this issue in a different PR so I am closing this one. If it needs to be reopened let me know

PR #2429, opened 2024-04-03
Added wildcard character compatibility for DT API Whitelist

prykon:
Currently, a user could only whitelist endpoints with static paths. This meant that if the endpoint passes a parameter through the URL, it couldn't be whitelisted.

This upgrade allows users to use the star (*) wildcard character in order to whitelist dynamic rest api endpoint paths.

**Screenshot**
Screenshot 2024-04-03 at 12 37 24

### DIFF
--- a/dt-core/configuration/restrict-rest-api.php
+++ b/dt-core/configuration/restrict-rest-api.php
@@ -109,6 +109,14 @@ function dt_dra_only_allow_logged_in_rest_access( $access ) {
         $authorized = true;
     }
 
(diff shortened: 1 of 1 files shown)


prykon:
@cairocoder01  Glad others find it useful as well!

corsacca:
Thanks @prykon and @cairocoder01!

Would be awesome to add a couple lines of example to that page to show what is possible. @prykon would you be up for doing that?

PR #2444, opened 2024-04-24
CT#22 - Support Notification Numbers

kodinkat:


corsacca:
@kodinkat It looks like the last 2 will try to display even if the channels are not enabled?

corsacca:
@kodinkat awesome!
Thinking this through.

image

Change `[work number]` to "none set".

Lets make this a filter so the theme does not need to keep track of what channels are added via plugins.

The twilio plugin will hook into the filter and add the SMS and Whatsapp line.

By each line add a edit icon that opens the existing edit modal.

corsacca:
Thanks @kodinkat !

Issue #2409, opened 2024-03-15
Customizations (D.T) > Custom Fields > Location


corsacca:
Ability to create a custom location fields (on custom post types)

PR #2446, opened 2024-04-24
#2409 - Support Creation Of Custom Location Fields

kodinkat:
- fixes: #2409


corsacca:
@kodinkat do we not support the ability to create mulitple location fields?

corsacca:
Thanks @kodinkat 
Ok. So this gets the default location grid field working.
If mapbox (and google) is enabled we also want the `location_grid` to work too.

corsacca:
@kodinkat 
I think the next step on this one is allowing the creation of a second field using the `location_meta` type.

```
$fields['location_grid_meta2'] = [
    'name'        => __( 'Locations or Address2', 'disciple_tools' ),
    'type'        => 'location_meta',
    'tile'      => 'details',
    'mapbox'    => false,
    'hidden' => false,
    'in_create_form' => true,
    'icon' => get_template_directory_uri() . '/dt-assets/images/map-marker-multiple.svg?v=2',
];
```

You can start by simply adding this to the contact's fields. There will need to be changes to make it operatable:
image

dt-mapping/location-grid-meta.php
```diff
@@ -99,10 +99,10 @@ public static function validate_location_grid_meta( &$location_grid_meta = null
 
         public static function get_location_grid_meta_by_id( $grid_meta_id ) {
             global $wpdb;
-            return $wpdb->get_row( $wpdb->prepare( "SELECT * FROM $wpdb->dt_location_grid_meta WHERE grid_meta_id = %d", $grid_meta_id ), ARRAY_A );
+            return $wpdb->get_row( $wpdb->prepare( "SELECT * FROM $wpdb->dt_location_grid_meta WHERE grid_meta_id = %d OR postmeta_id_location_grid = %d", $grid_meta_id, $grid_meta_id ), ARRAY_A );
```
corsacca, line 102: I don't get this one

corsacca:
It seems like the `location_grid_meta2` field is still dependent on a `location_grid` field.
So to create a second location_grid_meta field, we would need to also create a second `location_grid` field.

Seem like this could get messy.

2 solutions: 
We merge the 2 fields into 1. THe geolocation key would change the UI and API workflow.
We auto create the needed location grid field when creating a location grid meta field.

corsacca:
We decided to go with this strategy instead: https://github.com/DiscipleTools/disciple-tools-theme/issues/2494

Issue #2309, opened 2024-01-05
comment type @mentions notifications


corsacca:
insert_notification_for_comment only sends a notification to @mentioned users if the comment is of type 'comment'

kodinkat:
It would seem it's a size restriction issue; as the wp_comments table currently has a column comment_type `varchar(20)` size constraint; which will fail for longer types.

![Screenshot 2024-05-09 at 13 12 53](https://github.com/DiscipleTools/disciple-tools-theme/assets/19330800/b651dd18-eb6c-4307-b693-a047755d12ac)

PR #2469, opened 2024-05-09
#2309 - Comment Type @Mentions Notifications 

kodinkat:
- fixes: #2309


corsacca:
@kodinkat it appears this will enable a notification for all new comments, regardless of type.
We only want notifications on non "comment" types if the comment includes an @mention

corsacca:
Thank you @kodinkat !

PR #2484, opened 2024-05-21
feat: add where function to dt-reports class

squigglybob:
Adds functionality to the class in dt-reports.php to find reports where some things are true.


squigglybob:
@corsacca I've added in this function so I can pull out reports that only match on certain fields.

I'm thinking actually that this could be more powerful if it mimicked something like laravels where function where you can add in comparison functions.

so instead of an assosciative array which assumes equality in the where clause, it would have an array of arrays

```php
Disciple_Tools_Reports::where( [
   [ 'column_name', '=', $value ],
] );
```

and the middle value of the array could also be '!=', '>', '<', 'not' etc.

dt-reports/reports.php
```diff
@@ -310,6 +310,54 @@ public static function get( $value, $type ) : array {
...
+            }
+
+            $new_sql .= $wpdb->prepare( ' %1$s = ', $key );
+
+            if ( is_numeric( $value ) ) {
+                $new_sql .= '%d';
+            } else if ( is_string( $value ) ) {
+                $new_sql .= '%s';
+            } else {
+                dt_write_log( __METHOD__ . ' $value is of type ' . gettype( $value ) . ' but should be a string or number' );
```
corsacca, line 340: @squigglybob can we return an error instead?

squigglybob: sure :)

squigglybob: done

squigglybob:
Are there any other changes that need to be made in light of our recent conversation about keeeping things consistent?

corsacca:
hey @squigglybob 
```
Disciple_Tools_Reports::where( [
   [ 'column_name', '=', $value ],
] );
```
This does not seem to match the code.

We also talked about matching the format in https://developers.disciple.tools/theme-core/api-posts/list-query

squigglybob:
> hey @squigglybob
> 
> ```
> Disciple_Tools_Reports::where( [
>    [ 'column_name', '=', $value ],
> ] );
> ```

That was an example of how I was thinking it could be. It's not currently how it works.

> 
> This does not seem to match the code.
> 
> We also talked about matching the format in https://developers.disciple.tools/theme-core/api-posts/list-query

That was the thing I was thinking of.  I'll look at getting it matching that format next

squigglybob:
@corsacca hey bro, when you're back in the saddle, it would be good to connect on this one.

Currently the where function matches the list_posts search and query for the very simple cases (except for putting the values inside an array, which could be coded to cope with string/array depending on what result is desired)

 could we merge this one in, and I can create a new ticket to upgrade the where clause to have more of the complicated features like ORing etc.

The reports table and posts are quite different in the way they work, so it would be good to talk through exactly how we want the search query params to work.

corsacca:
@squigglybob 

The function name `where` seems a bit confusion. Maybe we use: `list_reports()` instead.

The current format for the `$args` or `$where` is:

```
[ 'post_id' => 3, 'type' => 'key' ]
```
@todo, add a comment for expected args format

This format does not support the expansion of the operations ( '!=', '>', '<', 'not' etc. )
So when/if we do want to add them we'd need to refactor the args and change the places using the function across multiple plugins.

I'd rather we build this out nicely. We can use direct $wpdb calls until it is done, which might be simpler anyways.

squigglybob:
@corsacca I should have implemented what we talked about straight away. 😬 Hopefully  I remembered what we talked about
Should be done if i did 😁

corsacca:
Thanks @squigglybob!

Issue #2495, opened 2024-05-28
Personal Genmapper: restrict to the records i'm responsible for


corsacca:
In general this will be the assigned to.

For contacts this includes the subassigned.

PR #2504, opened 2024-05-30
#2495 - Metrics - Personal - Gen-Map - Restrict To Records Responsible For

kodinkat:
- fixes: #2495


corsacca:
@kodinkat 
Can the root node show the users contact record? instead of "...."
image

corsacca:
Thanks @kodinkat!

PR #2516, opened 2024-06-14
Add color variables to css

cairocoder01:
This copies the SASS color variables to CSS variables to allow their easy re-use within plugins


corsacca:
@cairocoder01, are these the same colors as:
https://developers.disciple.tools/style-guide ?

cairocoder01:
@corsacca They are now. I didn't know that style guide existed. This was just exposed those color variables from the foundation theme. I have now updated the secondary and success colors in the foundation theme to match the style guide. The guide doesn't have anything defined for warning or alert colors though, so I left those as the foundation defaults.

corsacca:
Thanks @cairocoder01!

PR #2517, opened 2024-06-17
Added custom Comments action to workflows

jlamanskygitt:


cairocoder01:
As context for other reviewers, I requested @jlamanskygitt to create a workflow action that would add a comment to a post. The next feature of this will include putting tokens for users in the comment based on a field on the record. This would allow triggering a custom notification to a user based on a change on the record.

cairocoder01:
@jlamanskygitt This is looking good and does add a comment as it should. Nicely done.

One thing we need to do is make sure it doesn't add the same comment over and over again. There is a function in `workflows-execution-handler.php:438` called `already_executed_actions()` that determines if an action has already been executed or not. I think we need to add some logic in there for this comment addition to fetch the list of comments for this post (`DT_Posts::get_post_comments($post_type, $post_id, false);`) and check if the action's comment has already been added to the post. If so, we should not execute the action again.

What I did to test this was to create a custom multi_select field on contacts with two possible values: Option 1 and Option 2. I created a workflow that said if Option 1 was selected, add a comment. So when I selected Option 1, it added a comment as it should. Then, when I also selected Option 2 (so that both were selected), it added the comment again. I then unchecked Option 1, and no new comment was added (as expected). Once I selected Option 1 again, the comment was added a third time.

If we follow this same procedure, I would expect that the comment would only be added once (when Option 1 was first selected).

dt-workflows/workflows-execution-handler.php
```diff
@@ -489,7 +499,7 @@ private static function already_executed_actions( $actions, $post, $post_type_se
     private static function process_action( $field_id, $action, $value, $post, $post_type_settings ) {
         if ( ! empty( $field_id ) && ! empty( $action ) && ! empty( $value ) ) {
 
-            if ( isset( $post_type_settings['fields'][ $field_id ]['type'] )) {
```
cairocoder01: This comment can be removed

dt-workflows/workflows-execution-handler.php
```diff
@@ -464,6 +464,16 @@ private static function already_executed_actions( $actions, $post, $post_type_se
                         $current_state = self::condition_contains( $field_type, $field, $action->value );
                         break;
                 }
+            } else if ( $action->field_id === 'comments' ) {
+                $post_comments = DT_Posts::get_post_comments($post['post_type'], $post['ID'], false);
+                //$current_state = in_array($action->value, $comments, false);
+                //$comment->comment_content;
```
cairocoder01, line 470: Clean up this commented code

cairocoder01 (commented):
This works as I would expect it. After cleaning up some of these code comments, I think it's ready for review by @corsacca

corsacca:
Awesome!
Nice work @jlamanskygitt!

image
image

corsacca:
@jlamanskygitt, @cairocoder01 
Is there supposed to be a conditions option for comments in step 2?
image

cairocoder01:
@corsacca No, there isn't supposed to be a condition. It's just used in the actions, though I know it ends up appearing in the conditions too. Do you think it should be removed from there?

corsacca:
@cairocoder01 we could hide it from step 2, 
Or implement a simple: If the comment "contains" this text

cairocoder01:
@corsacca I'm not keen on adding a "contains" condition at this point. I would actually just think that we should exclude any field from the conditions list if it doesn't actually have any condition options for the second field.

corsacca:
Yeah. That sounds good.
This PR also need to pass the prettier check.

corsacca:
Thanks @jlamanskygitt!

Issue #2494, opened 2024-05-28
Merge location_grid and location_grid_meta into one field


corsacca:
the `location` field 
- uses the predefined `dt_location_grid` table to list options
- saves the grid_id in the postmeta table for records and usermeta for users.

the `location_meta` field
- uses the mapbox or the google geocoding api to list options
- finds the corresponding location_grid and saves the grid_id in the postmeta table for records and usermeta for users.
- saves the extra meta in the `dt_location_grid_meta` table (lat, lng, level, label, etc)

The location_meta field requires the location field. So in essence the location_meta in an extension of the location field. These could be considered the same field with different UI and workflows.

We ran into blocks In trying to let the admin create a location_meta field on custom post types. See [pr](https://github.com/DiscipleTools/disciple-tools-theme/pull/2446).

A custom location_meta requires a custom location_grid and we'd like to not have to create 2 fields to make this work.

So can we build out the `location` field to have 2 modes: normal, or with a geolocation key.

PR #2519, opened 2024-06-20
#2494 - Merge Location_Grid And Location_Grid_Meta Into A Single Field

kodinkat:
- fixes: #2494 
---

#### Further Design Discussion Points

- Need to determine which field should be updated when record contains multiple meta location fields and `contact_address` field populated.

- Once proposed shape has been merged; need to identify plugins impacted by changes and in need of updates:
  - DT_Webform
  - DT_Import


corsacca:
@kodinkat If i use the API to set the location_grid to Paris as so:
image

The location field stays empty:
image

On the contact the location grid field acquires 2 values:
image

kodinkat:
403 Forbidden exception encountered whilst trying to directly access API; which I believe is the result of an abscent nonce. As a result, no updates are made to location field.

Screenshot 2024-07-25 at 11 49 45

Recent updates now ensure multiple postmeta location grid entries are no longer made.

Issue #2488, opened 2024-05-23
D.T Workflows: updates when connections are triggered on other records


corsacca:
This scenario does not seem to trigger:
If contact A is subassigned to Multiplier B then we want the "Does follow up" milestone to be added on Multiplier B.

![image](https://github.com/DiscipleTools/disciple-tools-theme/assets/24901539/bd8f0465-3317-4536-bdcb-a5541520d369)

PR #2522, opened 2024-06-27
#2488 - DT Workflows - Updates When Connections Are Triggered On Other Records

kodinkat:
- fixes: #2488


corsacca:
Thanks @kodinkat !

PR #2524, opened 2024-07-03
Integrations of Dt-multiselect-button-group into DT Theme

Ashima-Arora:
Hi @micahmills 

We have completed the following tasks into this -

1. Integrated Dt-multiselect-button-group into Create new contact form and accurate values posted into API. (dt-core\global-functions.php)
2. Also, Removed JQuery code which was earlier responsible for posting New contacts form data to API, not that is achieved via JavaScript and invoking API from web components API services.
3. Added 'isset()' function into access_module at line: 465. This was added while API hits in case of a single select field - **Contact Status** for fetching data and selecting the desired value throws an error. 

@gp-birender @vijenderrr

### DIFF
--- a/dt-assets/js/new-record.js
+++ b/dt-assets/js/new-record.js
@@ -77,22 +77,6 @@ if (
         window.location = response.permalink;
       });
     });
-
-  // Select all elements with class '.form-fields' and 'dt-comm-channel'
-  document
-    .querySelectorAll('.form-fields dt-comm-channel')
-    .forEach(function (element) {
-      element.addEventListener('change', function (event) {
-        let element_index = event.target.id;
-        let valueAttri = event.target.value;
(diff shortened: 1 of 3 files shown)


micahmills:
The most recent commits makes the multiselect button group appear appropriately, but it doesn't appear to be saving in either the new contact form or the existing contact page.

Ashima-Arora:
@micahmills 

Few changes in component service were not merged. we have created new [PR](https://github.com/DiscipleTools/disciple-tools-web-components/pull/86). Kindly verify once.

micahmills:
I Just merged that in and that does fix some of the issues. I am seeing it work on the new record page, and the data loads properly in the contact record page, but if  change the value on the record page it isn't being saved properly because on a refresh it goes back to the previous value.

PR #2525, opened 2024-07-08
#2523 - Ensure Geocoded Locations Are Copied When Transferring Contact To Another Instance

kodinkat:
- fixes: #2523


corsacca:
@kodinkat Is it possible to also transfer the whole meta data? like label, lat, long, level, etc?
image

corsacca:
@kodinkat transfer on the right should match source on the left.
the location grid id is different and 2 are created on the transfer
image

corsacca:
Thank you @kodinkat!

Issue #2313, opened 2024-01-09
Write a check to test of object caching is enabled.


corsacca:
Frustrating experience:
Updating settings in D.T, then going to the front end and not seeing that changes being reflected.

This is often due to object caching.

On the WP Admin Dashboard. Display a warning if object caching is enabled.

--- 
Will `wp_cache_flush()` work?

PR #2529, opened 2024-07-12
#2313 - Object Caching Enabled Check

kodinkat:
- fixes: #2313 
---
 
Screenshot 2024-07-12 at 10 49 10

### DIFF
--- a/dt-core/admin/js/dt-utilities-scripts.js
+++ b/dt-core/admin/js/dt-utilities-scripts.js
@@ -956,16 +956,41 @@ jQuery(document).ready(function ($) {
         value: Date.now(),
       }).then((secondary_ts) => {
         // Next, fetch most recently stored cache test value.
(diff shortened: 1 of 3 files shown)


corsacca:
@kodinkat 
I think this solution addresses local object caching instead of server level object caching. Is that right?

kodinkat:
> @kodinkat I think this solution addresses local/wordpress object caching instead of server level object caching. Is that right?

Yes, the cache for running WP instance, should be cleared accordingly; as hosting server provider approaches towards object cache management may vary.

corsacca:
This Issue is primarily to do with server side object caching as wordpress caching does not seem to be an issue.
Hence the access to a server with server side object caching enabled.

corsacca:
I'm closing this PR for now. It seems like we might need a different approach to figure this one out.

corsacca:
Thanks @kodinkat for your time testing the approach

Issue #2527, opened 2024-07-09
Metrics field selects on one row


corsacca:
Make this one line on all the metrics pages where it is used
<img width="419" alt="image" src="https://github.com/DiscipleTools/disciple-tools-theme/assets/24901539/5f6750a6-dacf-4630-a08c-1c00dd7ecc86">

PR #2533, opened 2024-07-16
#2527 - Metrics Field Selects On One Row

kodinkat:
- fixes: #2527 
---

Screenshot 2024-07-16 at 15 35 32


corsacca:
Thank you @kodinkat!

PR #2534, opened 2024-07-16
Fix/login template errors

squigglybob:
@corsacca 

This one fixes the content-type not working correctly when looking at the reset password emails today.

It also fixes the other array offset error happening during the login process


corsacca:
Thanks @squigglybob!!

PR #2536, opened 2024-07-17
SOO Login: allow logging in with firebase ID when email is not provided.

corsacca:


squigglybob:
@corsacca hey bro, could you give some more context for what this PR does please?
Looking at the code it looks like email-less logging in.

Where/how would it be used?

squigglybob (approved):

PR #2537, opened 2024-07-17
SOO Login: setting to persist login

corsacca:
Setting in SSO Login to extend how long a user is logged in for:

image

Default is session or 2 days.


squigglybob:
Amazing, thanks @corsacca

squigglybob:
Looks good.
I stuck 1000 days on it, and the cookie will expire on 2025-01-14 which is 179 days from now, which is pretty good.

This will work for cookie based auth like on zume or DT, but PG uses JSON tokens, so we would need to come up with another way of extending the life of those

PR #2541, opened 2024-07-18
Workflow action: Comment tokens

brady-lamansky-gtt:
When using the add comment option in workflows, you can tag referenced contacts/users in the comment based on any given connection field on the post.


cairocoder01:
@cairocoder01 I'll take a look at this

dt-workflows/workflows-execution-handler.php
```diff
@@ -583,14 +583,87 @@ private static function action_append( $field_type, $field_id, $value ): array {
...
+                // loop through each field, checking if the comment attempts to reference one of the fields
+                foreach ( $fields as $field => $data ) {
+                    if ( str_contains( $value, '{' . $field . '}' ) ) {
+                        $replacement = '';
+                        $current = 0;
+                        $references = $post[$field]; // get array of references from the field
+
+                        // special case: if the field is 'assigned_to', it is not an array of references,
+                        // but an array of the pieces of the 'assigned_to' field
+                        if ( $field == 'assigned_to' ) {
```
cairocoder01, line 606: Instead of using the field name, can we use the field type of `user_select`? People can add multiple user select fields if they so choose.

dt-workflows/workflows-execution-handler.php
```diff
@@ -583,14 +583,87 @@ private static function action_append( $field_type, $field_id, $value ): array {
...
+                $fields = array_filter( $settings['fields'], function( $obj ) {
+                    if ( $obj['type'] == 'connection' || $obj['type'] == 'user_select' ) {
+                        return true;
+                    }
+                    return false;
+                } );
+
+                // loop through each field, checking if the comment attempts to reference one of the fields
+                foreach ( $fields as $field => $data ) {
+                    if ( str_contains( $value, '{' . $field . '}' ) ) {
```
cairocoder01, line 599: I think we can do this in a little different way. Instead of looping over all of the fields, we could look for all of the tokens in the comment text and look for fields that match the token.

To do that, we can use RegEx (regular expressions). These are formulas you can write to parse any kind of text content. In this case, we can look for everything that's wrapped in curly brackets (`{...}`). I usually use https://regex101.com/ to write and test regex. In this case, the expression we can use is `\{(\S+?)\}`.

To use that in PHP, the sample code that that site generates is this:

```
$re = '/\{(\S+?)\}/m';
$str = 'Notify {subassigned} and {assigned_to}';

preg_match_all($re, $str, $matches, PREG_SET_ORDER, 0);

// Print the entire match result
var_dump($matches);
```

You can replace `$str` with the comment text. `$matches` will be an array where each item is a matched token in the comment. Each match will also be an array where the first item is the whole token with brackets, and the second item is just the text within the brackets. That second item should be the field name that you can use to check if there is a field on the post that matches, and then you can do the replacement.

dt-workflows/workflows-execution-handler.php
```diff
@@ -583,14 +583,87 @@ private static function action_append( $field_type, $field_id, $value ): array {
...
 
         return $updated;
     }
 
+    private static function get_references( $field_name, $post ) {
+        $replacement = '';
+        $current = 0;
+        $references = $post[$field_name];
+        foreach ( $references as $reference ) {
+            ++$current;
```
cairocoder01, line 649: Instead of manually tracking the array index, you can get that using this syntax:

```
foreach ( $references as $current => $reference)
```
See https://www.php.net/manual/en/control-structures.foreach.php

cairocoder01 (changes requested):
This is working great. Nice job!

There's a few minor code improvements we can do that I commented on.

Also, this is creating the comment again if I flip the workflow trigger again. I think we need to do some work on the part that checks if the action has already been executed. I'm assuming it is looking for a comment with the text from the workflow, but we need to check for the comment from the workflow with all of the tokens replaced.

We also need some help text for users to know that tokens can be used and which ones. I'll create a separate task for that and we'll do it in a different PR.

dt-workflows/workflows-execution-handler.php
brady-lamansky-gtt: Sounds good to me. The issue was assigned_to only referenced a single record, so it wasn't an array of records. If people add multiple user_select fields, would they also only be able to reference a single user? If so then this would work, otherwise we'll still need a special case for 'assigned_to'

cairocoder01: Yeah, all user_select records only allow selecting a single user, so they will all act the same.

cairocoder01:
Hey @brady-lamansky-gtt, while reviewing your PR, I'd suggest the following code changes:

👉 [**Code Suggestion for #2541**](https://gitkraken.dev/drafts/11d83756-d6fe-42c2-928a-ab4d40721cf1?source=PRComment)
>https://github.com/DiscipleTools/disciple-tools-theme/pull/2541

You can also review and apply these suggestions [locally on your machine](https://gitkraken.dev/link/drafts/11d83756-d6fe-42c2-928a-ab4d40721cf1?type=suggested_pr_change&prEntityId=WyJnaXRodWIiLCJwciIsIjEiLCIiLCJQUl9rd0RPQkotUlJzNTEwUUJnIl0=&source=PRComment).

<details><summary>Learn more about GitKraken Code Suggest</summary>

>[Code Suggest](https://gitkraken.dev/redirect/aHR0cHM6Ly93d3cuZ2l0a3Jha2VuLmNvbS9zb2x1dGlvbnMvY29kZS1zdWdnZXN0P3NvdXJjZT1naXRodWIuY29tJm1lZGl1bT1wcl9jb21tZW50?source=PRSuggest&event=redirectToCodeSuggestPage&provider=github) liberates your code reviews from GitHub's restrictive, comment-only feedback style. As simple as suggesting changes in a Google-doc, provide real code suggestions from where you code, e.g. your IDE, and on anything in your project — not just on the lines of code changed in the PR.

_Join your team on [GitKraken](https://gitkraken.dev/redirect/aHR0cHM6Ly93d3cuZ2l0a3Jha2VuLmNvbT9zb3VyY2U9Z2l0aHViLmNvbSZtZWRpdW09cHJfY29tbWVudA==?source=PRSuggest&event=redirectToIndexPage&provider=github) to speed up PR review._
</details>

cairocoder01:
[Changes-On-2b83b3.patch](https://github.com/user-attachments/files/16372990/Changes-On-2b83b3.patch)

I'm not sure my GitKraken suggestion will work well for others. Here is a patch file that should fix some issues with not replacing all tokens and also cleaning up the variable and function names to be a little clearer.

@corsacca If you're able to apply that patch, I think this is ready for your review. I think @brady-lamansky-gtt is currently traveling and since I'm not the repo owner, I can't apply changes to his fork.

dt-workflows/workflows-execution-handler.php
```diff
@@ -583,14 +584,68 @@ private static function action_append( $field_type, $field_id, $value ): array {
...
+                    }
+                    // if not the last reference, add a comma & space
+                    if ( $current != sizeof( $references ) ) {
+                        $replacement = $replacement . ', ';
+                    }
+                }
+            }
+
+            // replace the {field} in the comment w/ the references
+            return str_replace( $match[0], $replacement, $value );
```
kodinkat, line 645: If outer foreach loop is to remain, consider moving return statement outside of loop; to be the last statement within `get_references()` function.

kodinkat:
Encountered the following exception, whilst attempting to execute workflow:

```sh
PHP Fatal error:  Uncaught TypeError: DT_Posts::add_post_comment(): Argument #3 ($comment_html) must be of type string, null given, called in /disciple-tools-theme/dt-posts/dt-posts.php:1386

Stack trace:

#0 /disciple-tools-theme/dt-posts/dt-posts.php(553): DT_Posts::add_post_comment('contacts', 293, NULL, 'comment', Array, false, true)

#1 /disciple-tools-theme/dt-workflows/workflows-execution-handler.php(527): DT_Posts::update_post('contacts', 293, Array, false, false)

#2 /disciple-tools-theme/dt-workflows/workflows-execution-handler.php(430): Disciple_Tools_Workflows_Execution_Handler::process_action('comments', 'append', '@admin is readi...', Array, Array)
```

On a Faith Milestone update, I was expecting to see the comment: `@admin is reading the Bible!`

cairocoder01:
@kodinkat Yeah, I found those issues too. My patch should resolve them, but I don't have permissions to apply it. And maybe others haven't worked with patches to know how to apply them either. I'd upload the file here, but it won't let me.

Here is a gist with the whole execution handler file after some cleanup and optimizations I did. It should work after that, based on my testing: https://gist.github.com/cairocoder01/347e5b1d57b7cbc4ddfc14095fce3f5e

corsacca:
This is cool!
This worflow
image
Produces:
image

I love how the user is @mentioned if a user is detected.

corsacca:
Question.
Could all field types be supported? 
`Please check {textarea_field} and {custom_text_field}` results in `Please check and`.
This could be an option for a future upgrade.

corsacca:
Sorry for the delay in merging this one in.
Well done @brady-lamansky-gtt!

Issue #2498, opened 2024-05-28
Metrics > Personal Layers Map


corsacca:
Copy the dynamic layer map under personal and refactor to only show:

Records I have access to

PR #2483, opened 2024-05-20
#2443 - Metrics - Gen-Mapper Multiplier Version Of The Dynamic Metrics

kodinkat:
- fixes: #2443 
---

Screenshot 2024-05-20 at 16 08 15

PR #2542, opened 2024-07-19
#2498 - Metrics - Personal Layers Map

kodinkat:
- fixes: #2498


dt-metrics/records/dynamic-records-map.php
```diff
@@ -252,9 +265,33 @@ public function points_geojson( WP_REST_Request $request ) {
...
+            $params['assigned_post_ids'] = self::list_assigned_post_ids( ( ( $params['post_type'] === 'system-users' ) ? 'contacts' : $params['post_type'] ), get_current_user_id() );
+        } else {
+            $params['list_all'] = true;
+            $params['assigned_post_ids'] = [];
+        }
+
+        return $params;
+    }
+
+    private function list_assigned_post_ids( $post_type, $user_id, $fields = [ 'ID' ], $limit = 500000 ): array {
```
corsacca, line 283: @kodinkat,  lets add an extra param post_type_geojson with the restrict param and add the sql directly if it is present.
Only query subassigned of the post_type is contacts.
Also remember that "subassinged" does not use the $user_id, it needs to be the contact id.

corsacca:
@kodinkat, can we separate what layers are added to the personal map and the dynamic map? Right now those are synced.

corsacca:
@kodinkat 
A random multiplier seems to be able to see all contacts:
image

dt-mapping/mapping-queries.php
```diff
@@ -36,7 +36,7 @@ public static function get_by_grid_id( $grid_id ) {
         ", $grid_id ), ARRAY_A );
 
         if ( empty( $results ) ) {
-            $results = [];
+            $results = array();
```
corsacca, line 39: @kodinkat why all these array() changes?

dt-metrics/records/dynamic-records-map.php
```diff
@@ -18,17 +18,21 @@ class DT_Metrics_Dynamic_Records_Map extends DT_Metrics_Chart_Base
     public $slug = 'dynamic_records_map'; // lowercase
     public $js_object_name = 'wp_js_object'; // This object will be loaded into the metrics.js file by the wp_localize_script.
     public $js_file_name = '/dt-metrics/records/dynamic-records-map.js'; // should be full file name plus extension
-    public $permissions = [ 'dt_all_access_contacts', 'view_project_metrics' ];
+    public $permissions = [ 'dt_all_access_contacts', 'view_project_metrics', 'multiplier' ];
```
corsacca, line 21: @kodinkat `multiplier` is not a permission. Please explain

dt-mapping/mapping-queries.php
```diff
@@ -1301,34 +1301,72 @@ public static function post_type_geojson( $post_type, $args = [], $offset = 0, $
         global $wpdb;
 
         //phpcs:disable
+        // Determine if post id filtering should take place.
+        $post_id_filter_sql = '';
+        $shared_user_join_sql = '';
+        $shared_user_condition_sql = '';
+        if ( isset( $args['list_all'], $args['list_all_by_user_id'] ) && !$args['list_all'] ) {
+            $post_id_filter_sql = $wpdb->prepare( "
+            SELECT * FROM
```
corsacca, line 1310: I don't think this outer layer is needed

dt-mapping/mapping-queries.php
```diff
@@ -1301,34 +1301,72 @@ public static function post_type_geojson( $post_type, $args = [], $offset = 0, $
...
+                      (
+                        SELECT api_p.ID
+                        FROM $wpdb->posts api_p
+                        LEFT JOIN $wpdb->postmeta as field_type ON ( field_type.post_id = api_p.ID AND field_type.meta_key = 'type' )
+                        WHERE (field_type.meta_value = 'access' OR field_type.meta_value = 'user' OR field_type.meta_value = 'access_placeholder')
```
corsacca, line 1315: This keeps groups from showing.
We don't need to filter by type for personal contacts

dt-mapping/mapping-queries.php
```diff
@@ -1301,34 +1301,72 @@ public static function post_type_geojson( $post_type, $args = [], $offset = 0, $
         global $wpdb;
 
         //phpcs:disable
+        // Determine if post id filtering should take place.
+        $post_id_filter_sql = '';
+        $shared_user_join_sql = '';
+        $shared_user_condition_sql = '';
+        if ( isset( $args['list_all'], $args['list_all_by_user_id'] ) && !$args['list_all'] ) {
```
corsacca, line 1308: `list_all` seems to be set by the browser.
Consider checking if the user can actually list all. the new genmapper should have code to copy.

dt-mapping/mapping-queries.php
```diff
@@ -1301,34 +1301,72 @@ public static function post_type_geojson( $post_type, $args = [], $offset = 0, $
...
+                      (
+                        SELECT api_p.ID
+                        FROM $wpdb->posts api_p
+                        LEFT JOIN $wpdb->postmeta as field_type ON ( field_type.post_id = api_p.ID AND field_type.meta_key = 'type' )
+                        WHERE (field_type.meta_value = 'access' OR field_type.meta_value = 'user' OR field_type.meta_value = 'access_placeholder')
+                          AND (api_p.post_status = 'publish')
+                          AND api_p.post_type = %s
+                          GROUP BY api_p.ID
+                          LIMIT 0, 50000
+                      ) AS assigned_post_ids
```
corsacca, line 1320: I'm confused at this whole section. It says assigned_post_ids, but no sql to to with assignment?
Could be that just the share filter is needed here.

corsacca (commented):
@kodinkat please sanity check this one

corsacca:
@kodinkat The records map no longer shows up for non admin users.
We should be using the same strategy as: https://github.com/DiscipleTools/disciple-tools-theme/pull/2483

Once that is addressed, please make user the https://multisite.lwp/wp-json/dt-metrics/personal/dynamic_records_map/post_type_geojson endpoint is secure. (Ie, that removing the "slug" param from the api call doesn't let the non admin user have access to all records.

corsacca:
Thank you @kodinkat!

Issue #2551, opened 2024-08-13
Missing Permissions On Role Updates Bug


kodinkat:
When you try to change Permissions for different Roles of different record types, the changes aren't saving correctly. 

If you click Update and then refresh the page, the changes will stick UNTIL you change Permissions on another record type and click Update.

PR #2552, opened 2024-08-13
#2551 - Missing Permissions On Role Updates Bug

kodinkat:
- fixes: #2551


corsacca:
Thanks you @micahmills!

PR #2553, opened 2024-08-15
Help Text for Comments

brady-lamansky-gtt:
This PR is dependent on my previous Workflow action: Comment Tokens PR, so it should be merged afterwards if everything looks good.
https://github.com/DiscipleTools/disciple-tools-theme/pull/2541


corsacca:
@brady-lamansky-gtt thanks for the PR!
For context on new PRs, would you be able to also attach a screenshot (or multiple) of the changes?

brady-lamansky-gtt:
Separated page into two sections, Help Text is the right column (hidden by default)
Screenshot 2024-08-16 at 8 15 56 AM
Screenshot 2024-08-16 at 8 18 00 AM


Added the Help Text button under Step 3 of Workflows
Screenshot 2024-08-16 at 8 17 26 AM

On clicking the button, updates the list to be visible & to contain information from the tab-workflows-docs.php file
Screenshot 2024-08-16 at 8 18 57 AM

The template that is used is the list Help Text I created in tab-workflows-docs.php
Screenshot 2024-08-16 at 8 20 05 AM

Adds applicable fields to the list while populating the Fields dropdown
Screenshot 2024-08-16 at 8 20 44 AM

corsacca:
@brady-lamansky-gtt, Sorry, i mean screenshots of the changed UI or functionality.
Like i did here: https://github.com/DiscipleTools/disciple-tools-theme/pull/2541#issuecomment-2293168296
The code changes are easily accessible in the files changes tab.

brady-lamansky-gtt:
When you reach step 3 of creating a workflow, you will see a button for Help Text.
Screenshot 2024-08-16 at 9 05 17 AM

After clicking it, a tab on the right side of the page will be visible that shows all fields that can be referenced through comment tokens (either user select fields or connection fields). It displays the field name & how to reference it.
Screenshot 2024-08-16 at 9 05 31 AM

dt-core/admin/menu/tabs/tab-workflows.php
```diff
@@ -275,9 +275,17 @@ public function content( $tab ) {
 
             $this->template( 'begin' );
 
+            ?>
+            <div id="post-body-content">
+            <?php
```
cairocoder01, line 280: This is creating a duplicate div. `$this->template('begin')` already adds this, so this can be removed.

dt-core/admin/menu/tabs/tab-workflows.php
```diff
@@ -275,9 +275,17 @@ public function content( $tab ) {
...
             $this->workflows_post_types_section( $selected_post_type );
             $this->workflows_management_section( $selected_post_type );
             $this->workflows_design_section( $selected_post_type );
+            ?>
+            </div>
+
+            <?php $this->right_column() ?>
```
cairocoder01, line 287: Before the right column, we need to add `$this->template('right_column');` to add the HTML for the sidebar.

dt-core/admin/menu/tabs/tab-workflows.php
```diff
@@ -957,6 +971,31 @@ private function sort_workflows_by_name( $workflows ): array {
         return $workflows;
     }
 
+    public function right_column() {
+        ?>
+        <div id="postbox-container-1" class="postbox-container">
```
cairocoder01, line 976: Once the use of `$this->template('right_column')` is updated above, we can remove this container div.

cairocoder01 (changes requested):
This is looking good. Just a few things to clean up.

In addition to the various code comments, you can add the following at the end of the `dt_utilities_workflows_scripts` which will style the sidebar to float at the user's current scroll position. That also means we could remove the code to auto-scroll up to the top when clicking the help text link.

```
wp_register_style( 'workflows-css', false );
wp_enqueue_style( 'workflows-css' );
wp_add_inline_style( 'workflows-css', "
#post-body {
    display: flex;
    gap: 20px;
}
#postbox-container-1 {
    position: sticky;
    top: 2rem;
    height: 100px;
    float: none;
}
" );
```

brady-lamansky-gtt:
@cairocoder01 Two things I'm unsure if I did efficiently:

1. I replaced the hidden helptext button with the working one on line 1629-1633 of dt-utilities-workflows.js & hovering over the button doesn't make the mouse a click icon anymore.
2. I changed the list of fields within the table to be a height of 300px with overflow so that it's scrollable. With the right_column being sticky, many fields went off screen and couldn't be seen. Unsure if 300px is the optimal height

dt-core/admin/menu/tabs/tab-workflows.php
```diff
@@ -82,6 +82,21 @@ public function dt_utilities_workflows_scripts() {
...
+                #post-body {
+                    display: flex;
+                    gap: 20px;
+                }
+                #postbox-container-1 {
+                    position: sticky;
+                    top: 2rem;
+                    height: 100px;
+                    float: none;
+                }
```
cairocoder01, line 98: To better handle the overflow, let's update this to:

```
position: sticky;
top: 2rem;
height: calc(100vh - 2rem);
float: none;
overflow-y: auto;
```

dt-core/admin/menu/tabs/tab-workflows-docs.php
```diff
@@ -0,0 +1,12 @@
+<!-- Here you can create different Help Texts for use in different places -->
+
+<!-- Comments Field -->
+<div id="dt_right_docs_comments_title" style="display: none;">
+    Comments Field
+</div>
+<div id="dt_right_docs_comments_content" style="display: none;">
+    When creating comments, you can insert field values from the current record. This can be helpful for showing what has changed or for tagging users who are selected in a given field. The following fields are supported for replacement using the given replacement tokens.
+    <ul id="dt_right_docs_comments_list" style="height: 300px; overflow: auto;">
```
cairocoder01, line 9: This overflow was a good idea. See my comment on the CSS in the main workflow file for how we can move it to scroll the whole side column. So we can remove these inline styles.

dt-core/admin/menu/tabs/tab-workflows.php
```diff
@@ -743,7 +760,7 @@ private function workflows_design_section_step3() {
                             </tr>
                             <tr id="workflows_design_section_step3_action_value_tr">
                                 <td>
-                                    value:
+                                    value: <a id="workflows_design_section_step3_help_text" type="hidden"></a>
```
cairocoder01, line 763: It's not a good practice to have an empty anchor tag like this. Since this is a placeholder container, it would be better for it to be a `<span>` that we will swap the content of.

dt-core/admin/js/dt-utilities-workflows.js
```diff
@@ -1616,6 +1626,13 @@ jQuery(function ($) {
     response['html'] =
       `<textarea placeholder="Enter text..." style="min-width: 100%;" id="${window.dt_admin_shared.escape(response['id'])}">`;
 
+    let help_button = $('#workflows_design_section_step3_help_text');
+    const new_help_button = `<a id="workflows_design_section_step3_help_text">[<a class="dt-docs"
+        data-title="dt_right_docs_comments_title"
+        data-content="dt_right_docs_comments_content"
+        style="cursor: pointer;">&#63;</a>]</a>`;
+    help_button.replaceWith(new_help_button);
+
```
cairocoder01, line 1635: You can remove the cursor style if you set an `href` attribute on the tag. That is a feature of all browsers that it will only change the cursor if it has an `href`. You can just set it to `href="#"` so it doesn't actually follow the link. The `#` is used to link to a specific HTML element by id on the page and will just go to the top if js doesn't prevent it, like we do with this.

cairocoder01: This help link never gets removed. If you select Comments from the dropdown and then select some other field, the help link is still there. Whenever the dropdown is changed, we should remove this help link. If we change it to a span, you can just remove all the content of the span. And then in this case when the Comments field is selected, you can append this link within the span.

cairocoder01 (changes requested):
@brady-lamansky-gtt I've addressed your 2 questions with a few comments throughout the code. It was good to add the overflow, but we'll just move it to work on the whole side column.

cairocoder01 (changes requested):
@brady-lamansky-gtt  One small syntax fix and then it should be ready for merge

cairocoder01:
This looks good.
@corsacca This is ready from my perspective.

corsacca:
Brilliant. Nice work @brady-lamansky-gtt!

PR #2554, opened 2024-08-19
Site Links: Option for all Permissions

corsacca:


ChrisChasm (approved):
This makes sense. I often find myself using things this way more than not.

PR #2563, opened 2024-08-30
Change "no answer" quick action to "contact attempted".

corsacca:
image


corsacca:
This matches the seeker path bettor and does not link it to having to be a phone call.

PR #2580, opened 2024-09-27
Safeguard Against ValueError Bcrypt Null Chars Exceptions

kodinkat:
Implemented `random_bytes_no_null()` wrapper function, to ensure `random_bytes()` do not contain null characters; which cause a `ValueError` exception; following the addition of the [Bcrypt password must not contain null character](https://github.com/php/php-src/commit/6a5c04d01d#diff-97027f2a7f32316e1a9d2317ecd1abd1eca58b9bb9c30c7929c7a33b93df5ad5R184) check.

Screenshot 2024-09-27 at 13 40 49

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1208382720655232

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -89,21 +89,6 @@ if ( version_compare( phpversion(), '7.4', '<' ) ) {
         }
     } );
 
-    /** Setup key for JWT authentication */
-    if ( !defined( 'JWT_AUTH_SECRET_KEY' ) ) {
-        if ( get_option( 'my_jwt_key' ) ) {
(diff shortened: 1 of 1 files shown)


incraigulous:
Thanks, @kodinkat , this looks to be a good start! I'll be happy to have this merged to see if our tests work even if we have a caught exception. @corsacca, for context, this breaks our CI/CD occasionally, but it will work if we rerun tests with no changes. It will also randomly and intermittently break for us when running local tests.

corsacca:
Thanks guys

PR #2581, opened 2024-09-30
Missing first letters as a capital

mikeallbutt:


corsacca:
@mikeallbutt 
It feels like this one isn't meant to have a capital
image

PR #2584, opened 2024-10-02
Introduced New Filter Hook To Support Custom SSO Login Responses

kodinkat:
Introduced new `dt_sso_login_response` filter hook to support custom SSO login responses.

### DIFF
--- a/dt-login/login-endpoints.php
+++ b/dt-login/login-endpoints.php
@@ -81,25 +81,29 @@ class DT_Login_Endpoints {
         }
         //phpcs:enable
 
+        $error_response = null;
         $user_manager = new DT_Login_User_Manager( $payload );
 
         try {
             $response = $user_manager->login();
         } catch ( \Throwable $th ) {
-            return new WP_Error( $th->getCode(), $th->getMessage(), [ 'status' => 401 ] );
+            $error_response = new WP_Error( $th->getCode(), $th->getMessage(), [ 'status' => 401 ] );
         }
 
         if ( is_wp_error( $response ) ) {
-            return $response;
+            $error_response = $response;
         }
         if ( !$response ) {
-            return new WP_Error( 'login_error', 'Something went wrong with the login', [ 'status' => 401 ] );
+            $error_response = new WP_Error( 'login_error', 'Something went wrong with the login', [ 'status' => 401 ] );
         }
 
-        return new WP_REST_Response( [
+        // Provide opportunity to update response packet.
+        $rest_response = apply_filters( 'dt_sso_login_response', [
             'status' => 200,
             'body' => $response,
-        ] );
+        ], $payload, $error_response );
+
+        return !is_null( $error_response ) ? $error_response : new WP_REST_Response( $rest_response );
     }
 
     /**
--- a/dt-login/login-shortcodes.php
+++ b/dt-login/login-shortcodes.php
@@ -173,7 +173,7 @@ function dt_firebase_login_ui( $atts ) {
           const response = JSON.parse(json)
 
           if ( response.status === 200 ) {
-            const { login_method, jwt } = response.body
+            const { login_method, jwt, redirect_url } = response.body
 
             if ( login_method === 'mobile' ) {
               if ( !Object.prototype.hasOwnProperty.call( jwt, 'token' ) ) {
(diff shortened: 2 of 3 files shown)


corsacca:
@kodinkat can you give an example of how this will be used?

kodinkat:
> @kodinkat can you give an example of how this will be used?

Hey @corsacca, see `dt_sso_login_response` function within corresponding [PR](https://github.com/thecodezone/dt-home/pull/206/commits/878a28121ec31b1d1b91d0af19c1bbbc99c95431); which aims to update `redirect_url` based on login outcome - @incraigulous may be able to expand on some relevant use cases.

corsacca:
@kodinkat how does this compare with using this filter?
`$redirect_url = apply_filters( 'dt_login_redirect_url', $redirect_url );`

kodinkat:
> @kodinkat how does this compare with using this filter? `$redirect_url = apply_filters( 'dt_login_redirect_url', $redirect_url );`

@corsacca could it potentially be expanded to also accommodate sso payload and error data; to support further business logic?

dt-login/login-endpoints.php
```diff
@@ -81,25 +81,29 @@ public function login( WP_REST_Request $request ) {
...
 
-        return new WP_REST_Response( [
+        // Provide opportunity to update response packet.
+        $rest_response = apply_filters( 'dt_sso_login_response', [
             'status' => 200,
             'body' => $response,
-        ] );
+        ], $payload, $error_response );
+
+        return !is_null( $error_response ) ? $error_response : new WP_REST_Response( $rest_response );
```
corsacca, line 106: @kodinkat will return the $error_response if it is not null no matter if $rest_response has been updated via filter.
Is this the desired behavior?

kodinkat: Yep, so as to maintain the refactored logic-flow; $error_response to take priority.... Previously, it was returned immediately when encountered.

kodinkat: I guess, it could be refactored further to ensure if error detected; then no need to apply filter; but, my thinking was to also communicate to others, that an error has occurred..

corsacca: @kodinkat would they not need control over the error response in order apply to their business logic?

corsacca: I might not fully understand the aim. My intuition says just limit it to success responses. Or allow both success and errors to be modified

kodinkat: @corsacca yeah, maybe a more generic shape(s), might be to separate out across two new filters (`dt_sso_login_response_success` / `dt_sso_login_response_error`) - Both of which would return their respective success/error array/obj...

Interested parties can then hook in and process accordingly.... what do you think?

corsacca:
@kodinkat I think the intended workflow would be to pass in or load the SSO login with the desired redirect instead of trying to determine the correct redirect after login.
Is there a specific reason for doing it after login instead of before?

kodinkat:
> @kodinkat I think the default workflow would be to pass in or load the SSO login with the desired redirect instead of trying to determine the correct redirect after login. Is there a specific reason for doing it after login instead of before?

@incraigulous could the proposed flow work for you? If so, we may need to expand the firebase shortcode, to also accept success/error redirect_url override params.

incraigulous:
> > @kodinkat I think the default workflow would be to pass in or load the SSO login with the desired redirect instead of trying to determine the correct redirect after login. Is there a specific reason for doing it after login instead of before?
> 
> @incraigulous could the proposed flow work for you? If so, we may need to expand the firebase shortcode, to also accept success/error redirect_url override params.

I am going to be honest that I feel less informed than both of you here, so I don't feel as if I should be the decision-maker. I will say, one of our original ideas was to pass success/error URLs to the shortcode so I would be in favor of perusing that option.

kodinkat:
> > > @kodinkat I think the default workflow would be to pass in or load the SSO login with the desired redirect instead of trying to determine the correct redirect after login. Is there a specific reason for doing it after login instead of before?
> > 
> > 
> > @incraigulous could the proposed flow work for you? If so, we may need to expand the firebase shortcode, to also accept success/error redirect_url override params.
> 
> I am going to be honest that I feel less informed than both of you here, so I don't feel as if I should be the decision-maker. I will say, one of our original ideas was to pass success/error URLs to the shortcode so I would be in favor of perusing that option.

@corsacca @incraigulous I'll aim to circle back onto this and update firebase shortcode to accommodate new success/error URL parameters.

kodinkat:
@corsacca see new `success_url` & `error_url` firebase shortcode override params.

corsacca:
Thanks @kodinkat!
It looks like this PR is no longer needed.
The user fix is also covered here:
https://github.com/DiscipleTools/disciple-tools-theme/pull/2598

incraigulous:
👍

PR #2597, opened 2024-11-01
Apply filter to notify all for different comment types

cairocoder01:
**Problem:**
When a plugin or external system creates a comment that is not type "comment" and a user has selected in their profile to "receive a notification for any update that happens in the system", that user will not be notified. Those notifications are only triggered on comments with a type of "comment"

**Solution:**
This adds a filter to include other comment types in the list of those that should be notified. So in a plugin that creates a different comment type, it can include that comment type in this filtered list to indicate that it should also trigger notifications.

### DIFF
--- a/dt-notifications/notifications-comments.php
+++ b/dt-notifications/notifications-comments.php
@@ -30,7 +30,8 @@ class Disciple_Tools_Notifications_Comments
         }
 
         $comment_with_users = self::match_mention( $comment->comment_content ); // fail if no match for mention found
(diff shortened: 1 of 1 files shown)


cairocoder01:
To reproduce:
1. Go to a user's profile, scroll all the way down, and select the option to Follow All Contacts
2. Using the API, create a comment on a contact with type "comment"
3. Confirm that the user gets a notification of the new comment
4. Using the API, create a comment on a contact with type "somethingelse"
5. Confirm that the user does not get a notification
6. Add filter from a plugin or code snippet to add "somethingelse" to this filtered list
7. Using the API, create a comment on a contact with type "somethingelse"
8. Confirm that the user gets a notification

corsacca:
Looks good. Thank you @cairocoder01!

PR #2601, opened 2024-11-20
Option to disable site link HTTPS check

corsacca:
This will let site links work on instances behind a VPN that are set up as localhost.
Or with local dev when debug is false.

image


corsacca:
@hansr In the end we went with not adding UI in the settings tab and only making the filter available.

PR #2604, opened 2024-12-06
Option to only install location grid once on a multisite.

corsacca:
Add this to a mu-plugin:
```
add_filter( 'dt_location_grid_table', function ( $table_name ){
    global $wpdb;
    return $wpdb->base_prefix . 'location_grid';
} );
add_filter( 'dt_using_multisite_location_grid_table', '__return_true' );
```
and save 10 to 15MB per subsite by only using one location grid table

To get an sql statement to drop all the location_grid tables:
```
SELECT CONCAT( 'DROP TABLE ', GROUP_CONCAT(table_name) , ';' ) 
    AS statement FROM information_schema.tables 
    WHERE table_name LIKE '%dt_location_grid';
```

### DIFF
--- a/dt-mapping/class-migration-engine.php
+++ b/dt-mapping/class-migration-engine.php
@@ -74,6 +74,12 @@ class DT_Mapping_Module_Migration_Engine
         if ( $target_migration_number >= count( self::get_migrations() ) ) {
             throw new Exception( "Migration number $target_migration_number does not exist" );
         }
+        global $wpdb;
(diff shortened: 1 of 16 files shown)


squigglybob (commented):
I'm not able to test it all works fine.
But it all seems to make sense.

You're making a central locations grid table, so no prefixes needed.
And the migrations don't need to make their own location grid pieces anymore.

I don't fully understand the migration side of things, but looks good

PR #2638, opened 2025-01-10
use ✓ > and '•' in the menu

corsacca:
How does this look?
image


corsacca:
Not heading this direction

Issue #2610, opened 2024-12-12
Not possible to reorganize communication_channels within the details tile.


corsacca:
The fields don't show up in the Customizations (D.T) tab. Changing the order results in the com channels being pushed to the bottom.

<img width="817" alt="image" src="https://github.com/user-attachments/assets/4de61900-6d2e-472e-914b-1bdb480d15a8" />

PR #2646, opened 2025-01-30
#2610 - Not Possible To Reorganise Comm_Channels Within The Details Tile

kodinkat:
- fixes: #2610 

Screenshot 2025-01-30 at 13 04 29


corsacca:
Thanks @kodinkat,
This does point out where the issue is.
I'd like to confirm that the communication channels need to stay as "not customizable".
If not, then changing that param on the com channels will be a better solution than making an exception for them

PR #2650, opened 2025-03-04
Web components 0.7.0

cairocoder01:
This implements the web components that are ready in version 0.7.0 of the web components library. That includes the following components:

- `dt-text` (field type: `text`)
- `dt-textarea` (field type: `textarea`)
- `dt-date` (field type: `date`) (does not handle `datetime`)
- `dt-multi-text` (field type: `communication_channel`) (may be extended in the future to handle `link` as well)
- `dt-single-select` (field type: `key_select`)
- `dt-multi-select` (field type: `multi_select` when using typeahead display)
- `dt-multi-select-button-group` (field type: `multi_select` when using button display)
- `dt-tags` (field type: `tags`)
- `dt-connection` (field type: `connection`)
- `dt-button` - used within `dt-multi-select-button-group` but not replace within the theme at the moment

This has been tested on the contact detail and new contact screens. 

Bulk add of new posts has been removed upon discussion with @corsacca. Code is still there but link to it has been removed.

Javascript files of details and new screens have been cleaned to remove old code that is no longer necessary and has been tested after removal to ensure everything is still working as expected.

Future versions of the component library will implement more components and integrate them with the theme in future PRs.


corsacca:
Awesome work!
Here is an unordered list of comments. Some bugs, some styling consistency and preferences.

There is an empty <p> tag under the contact status, adding uneeded margin.
image

Multiselect buttons font size is bigger than before and bigger than surrounding elements
image

Multiselect items without icons have extra whitespace:
image

Missing padding/margin under multiselect buttons before next field:
image

X button on first empty com channel seems confusing
image

Cannot unselect key select back to "empty" value:
image

After setting baptism date, a modal opens and i get this error:
Trying to access array offset on null in /wp-content/themes/disciple-tools-theme/dt-core/utilities/dt-components.php on line 53

Com field. Adding a value and then clicking remove clears clears the field, but on refresh it value is there again. After refresh the remove button works.
Some issue for the links field.
image


I see this one is for 0.8. Nice. I feel the + button to add is missing from connection and tag fields:
https://github.com/DiscipleTools/disciple-tools-web-components/issues/105

cairocoder01:
@corsacca 
Thanks for the thorough review. All of your comments should be address with these commits and v0.7.1 of the component library.

- `dt-single-select` had extra bottom margin on the select element, which is fixed. And overall status had a `<p>` tag that has been changed to `<div>` to prevent the extra margin intended for paragraphs
- `dt-multi-select-button-group`
    - fix font and padding sizing
    - remove empty icon space
- fixed spacing between all fields. No margin on components themselves. Margin added to cells in tiles.
- fix error when opening baptism modal
- add empty values for single selects when appropriate
- multi-text: did a re-bind of value after saving so new values can be deleted by key
- remove 'x' from first comm channel

corsacca:
Thanks @cairocoder01!

corsacca:
Can we consider leaving in the listeners in the details.js?
Some plugins like the Coaching checklist depend on those:
image

corsacca:
In merging records, fields should be readonly:
image

now:
It is now obvious that these are disabled:
image

And com channel and multiselect fields are editable, connect fields have eternal spinner, etc.

I think this is something to do with the disabled state?

cairocoder01:
I've added back in some of the old field init scripts for the detail screen. Each of them logs a deprecation warning in the console if they are triggered.

I went through the whole merge records process and updated it to work properly with the components.

I also went through the bulk edit feature on the list page, so that should be tested and working as well.

corsacca:
CHanged:
image

to

image

corsacca:
@cairocoder01 I looks like number fields are not updating

corsacca:
Thanks @cairocoder01!

Which field types are still left to do in future PRs?
location, number

cairocoder01:
Remaining to be implemented:

- number
- boolean
- datetime
- user_select
- location / location_meta
- link
- church health metrics

corsacca:
3 2 1 Components merge!
Thank you @cairocoder01!

PR #2654, opened 2025-03-06
Cypress Testing Framework Integration

kodinkat:
Initial support for the [Cypress Testing Framework](https://www.cypress.io), with a sample `Create New Contact` test. Additional potential tests to be integrated:

- Add Bulk Contacts Records.
- Delete Contact Record.
- Delete Group Record.
- Search For Records.
- Update Contact Record.
- Update Group Record.


corsacca:
Thank you @kodinkat!

PR #2656, opened 2025-03-07
force loading the correct .mo language file in magic links

corsacca:


cairocoder01:
This works for me @corsacca

PR #2661, opened 2025-03-19
Pass apiRoot URL to ComponentService

cairocoder01:
Resolves issue for multi-site instances where API requests were going to domain root instead of subdirectory root.


corsacca:
Awesome! Thank you @cairocoder01

PR #2662, opened 2025-03-19
Update components to v0.7.4

corsacca:
@cairocoder01 does this look right updating the components library?


cairocoder01:
Did you update the package.json file? I would expect at least the lock file would show an updated version number for 0.7.4.

corsacca:
yes: https://github.com/DiscipleTools/disciple-tools-web-components/commit/1db20c618cf000839158eeb9d09dacc86a6c46a7
Is the lock file changed with npm install?

corsacca:
Thanks for the clarification. for updating D.T to use the latest components release.
Change the package number in `package.json` and run `npm install`.

cairocoder01:
Yep, that all looks good now

PR #2664, opened 2025-03-26
Allow fetching magic link lang from request header

cairocoder01:
Magic links can currently pull the locale from query, contact, or user. This adds another option to pull from the `HTTP_ACCEPT_LANGUAGE` [request header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Language) which indicates the user's browser language and thus their preferred language.

Since the base class looks at the `translatable` array for which locations to look for the locale, this is completely an opt-in value and shouldn't affect any existing magic links.


cairocoder01:
The CS error seems to be on a file I didn't touch. Can I leave that alone or should it be fixed?

PR #2668, opened 2025-04-09
Support New DT-Import Non-Overwrite Setting

kodinkat:


corsacca:
I'm curious about adding this feature to the `update_post` function. Something like `do_not_overwrite_exsting_fields` param is past in, then it does the check there.

We'll also need to write some unit tests for this. AI should be good at this part.

kodinkat:
> @kodinkat does `dt_ignore_duplicated_post_fields` check if the 2 values are different or if any value exists on the existing record? I'm having trouble following the logic. The previous behavior is that all fields are updated. The new behavior:
> 
> * for single values fields: we don't update the field if it already has a value
> * for multiple value fields: we only add new values if they don't already exist
> 
> Does that sound right?

This is correct, as new logic only updates/adds value if field does not already contain it. Therefore, ignoring any duplicates.

kodinkat:
> I'm curious about adding this feature to the `update_post` function. Something like `do_not_overwrite_exsting_fields` param is past in, then it does the check there.
> 
> We'll also need to write some unit tests for this. AI should be good at this part.

Sounds good, maybe tackle as a new ticket, once functionality has been merged in?

corsacca:
On this one lets go slower.
Lets move the do not overwrite to the update post
and lets create unit tests to check that things are working as expected.

kodinkat:
@corsacca pls see latest shape and let me know if this holds up for you...

corsacca:
@kodinkat The same thing applies for all the other fields too. For example date is being updated even if original has a date.
Please have a look at each field

corsacca:
Thank you @kodinkat!

PR #2687, opened 2025-05-27
Fix bug on new contact screen

cairocoder01:
As a result of the 1.70 release with updates for web components, the location field is broken on the new contact page.

Locations field was showing a plus icon with no empty field to enter a location. When selecting a location, it would just show a spinner because it sent an API request that was failing. This updates the logic to correctly set it up like it previously was working (no plus icon, a default blank input, and no API request).


cairocoder01:
@corsacca Could you look at this one right away? It is affecting my digital responders from being able to input their new contacts correctly.

corsacca:
Thanks @cairocoder01 for the fix!

PR #2689, opened 2025-05-28
Update on location-grid-meta - Checks if and informed location already exists in the contact.

adventureit:
Checks if and informed location already exists in the contact. 
Asking @cairocoder01 review.

### DIFF
--- a/dt-mapping/location-grid-meta.php
+++ b/dt-mapping/location-grid-meta.php
@@ -98,31 +98,28 @@ if ( ! class_exists( 'Location_Grid_Meta' ) ) {
         }
 
         public static function get_location_grid_meta_by_id( $grid_meta_id ) {
-            dt_write_log('get_location_grid_meta_by_id');
+
             global $wpdb;
             return $wpdb->get_row( $wpdb->prepare( "SELECT * FROM $wpdb->dt_location_grid_meta WHERE grid_meta_id = %d", $grid_meta_id ), ARRAY_A );
         }
 
-        public static function get_existing_locations( $post_id, $grid_id ) {
-            dt_write_log('check_existing_locations');
+        public static function checks_existing_location( $post_id, $grid_id ) {
+
             global $wpdb;
 
-            $dates_meta = $wpdb->get_results( "
+            $data_meta = $wpdb->get_results( "
                     SELECT * FROM $wpdb->dt_location_grid_meta
                     WHERE 
                     ( post_id = $post_id )
                     AND  grid_id = $grid_id 
                     ", ARRAY_A);
 
-            return $dates_meta;
+            return $data_meta;
         }        
 
         public static function add_location_grid_meta( $post_id, array $location_grid_meta, $postmeta_id_location_grid = null ) {
             global $wpdb;
             $geocoder = new Location_Grid_Geocoder();
-
-            dt_write_log( 'add_location_grid_meta - $location_grid_meta' );
-            dt_write_log( $location_grid_meta );
 
             self::validate_location_grid_meta( $location_grid_meta );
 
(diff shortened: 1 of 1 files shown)


dt-mapping/location-grid-meta.php
```diff
@@ -98,16 +98,35 @@ public static function validate_location_grid_meta( &$location_grid_meta = null
         }
 
         public static function get_location_grid_meta_by_id( $grid_meta_id ) {
+            dt_write_log('get_location_grid_meta_by_id');
```
cairocoder01, line 101: We should remove all these debugging statements.

dt-mapping/location-grid-meta.php
```diff
@@ -128,6 +147,19 @@ public static function add_location_grid_meta( $post_id, array $location_grid_me
...
+            if ( isset( $post_id ) ) {
+                $grid = self::get_existing_locations( $post_id, $location_grid_meta['grid_id'] );
+
+                dt_write_log( 'get_existing_locations -> grid' );
+                dt_write_log( $grid );
+
+                if( !empty( $grid ) )
+                {
+                    dt_write_log( 'Location already exists.' );
+                    return new WP_Error( __METHOD__, 'This location already exists in the contact.' );
```
cairocoder01, line 159: Should this really throw an error? Or should it just silently fail and log that the location wasn't added because it already exists? What context is this function being used in and what will happen in either the API request or UI when this function throws an error?

cairocoder01 (changes requested):

dt-mapping/location-grid-meta.php
cairocoder01: Ok. That makes sense. I saw it returned an error and thought it threw it, but I guess each of those functions that call it can handle that error in their own way. Sounds good. Thanks for clarifying. We should probably try to test those different functions that use this as best as we can.

corsacca:
Thanks for this PR!

Context:
When adding a location to a record: The grid_id is added to the post_meta
When using mapbox or google geocoding, a line is also added to the location_grid_meta table.

@adventureit Can you help me understand what workflow this PR helps with?
Should we be checking earlier at the grid_id in the post_meta table instead (or as well?)

Do you have handy an example of a address that resolves to 2 different grid ids?
Off the top of my head i can think of one reason. Some locations like Singapore are both a city and a state and the geocoder might sometimes give one and sometimes the other.

adventureit:
image
Example of same location with different grid_ids. Local testings.

cairocoder01:
> image Example of same location with different grid_ids. Local testings.

@adventureit When I look up those two location grid IDs (which you can do in the `wp_dt_location_grid` table on your instance), they look like completely different locations: 
image
image

Can you confirm how they are showing in the UI that they look the same?

adventureit:
Well, I am using the country name on my testings, not a specific place or city.

corsacca:
Hey @adventureit,

We're still needing some clarification on the workflow and use case here.

I did a quick test, if a contact has 2 addresses that map to the same location grid, then only one of them is saved.
Adding both of these does not work:
1430 Broadway, New York, NY, USA
1431 Broadway, New York, NY, USA

If we are specifically looking to avoiding geocoding the same "label" twice, then we might just do a check during the create_post() and update_post() functions.

So "France" and "France" would not be saved twice.

corsacca:
Thanks @adventureit for your initial work on this. I'm going to close this PR for now.
If we come back to this issue, i think we'd want to explore adding a parameter to the api endpoint that would let enable searching for duplicates. It could search for a match on the search meta label "1431 Broadway, New York, NY, USA" and see if has been used before.

PR #2691, opened 2025-05-29
New record hidden field order

cairocoder01:
When showing/hiding fields on the new record page, fields are often shown above the current scroll position, causing users to scroll back up past fields they have already completed to find ones they have not completed.

This uses CSS flexbox and the `order` property to re-order the hidden fields to only show below the banner to show/hide fields (and above the button and error container). It should provide a more streamlined user experience when manually entering new contacts/groups/records.


corsacca:
Thank you @cairocoder01!

PR #2693, opened 2025-06-05
2692 new record duplicate checker

### LINKED ISSUE #2692: Reimplement duplicate checker on contact create
There was a duplicate checker on the phone number and email fields.
When entering an value it would check the db to see if a contact with that value already exists.

cairocoder01:
Resolves #2692 

Re-implement duplicate checks on new record page for phone and contact fields.


corsacca:
Thank you @cairocoder01!

PR #2696, opened 2025-06-13
List fields selector

corsacca:


corsacca:
@kodinkat debounce only makes sense if the input was searching agains't an API.

PR #2709, opened 2025-07-08
Web Components: Add Connection/Tag buttons

cairocoder01:
Implements Web Components v0.7.8 which adds "Add New" icon buttons to dt-tags and dt-connection.


corsacca:
Brilliant!
Thank you @brady-lamansky-gtt  and @cairocoder01!

PR #2715, opened 2025-07-23
Feat/bulk messaging updates

squigglybob:
Adds the ability to add new placeholders for bulk sending of emails.
Additionally allows the headers of the emails to be adjusted
This is handy if you are sending an html message and need to put in the mime type etc. headers to not have the message wrapped with another layer of html


corsacca:
Very nice! Thanks @squigglybob

kodinkat (commented):
@squigglybob 

This PR is a solid improvement to the bulk messaging system, as a cool feature to have. Some suggestions/areas for improvement below:

- __Documentation:__ Where relevant, ensure that the new filters and their usage are well-documented, so that site administrators and developers know how to leverage the new features.

- __Testing:__ Confirm that the changes are covered by tests, especially for edge cases (e.g., sending to multiple recipients, different content types).

- __UI/UX:__ If there is a user interface for configuring placeholders or headers, consider adding tooltips or help text to guide users.

PR #2726, opened 2025-08-12
Add record links to duplicate check

### LINKED ISSUE #2719: Duplicate Check: Link to record
In v1.69.1 and earlier - before web components were implemented - the duplicate check for phone and email on the new record page showed a link to the duplicate contact when one was found:

Image

Since the addition of components and the work to add back in the duplicate check, it no longer provides a link to the duplicate record:

Image

Per a user report of this issue:

> The responders keep asking me about the issue with duplicate checking. Before, when they entered a new contact and came to the phone number, as they entered the number a box would pop up showing them that it was a duplicate. They could click on that box and it took them straight to the contact record that was the duplicate so they could update that one instead of making a new contact. Now, a pop-up comes up saying that it's a duplicate, but they can't click on it. Instead they have to go back to the search bar and search for the phone number and click on that record. I know it may not seem like much, but they enter a lot of contacts and these small steps add up to big frustrations. They are asking if it will be fixed. I told them to wait for the DT update, but it looks like it updated this week and that issue is still there. Do you know if this is something they are planning to go back to? Or can we ask them to? You've told us before that if we are willing to contribute financially they could prioritize it. We are definitely willing to.

cairocoder01:
resolves #2719

When showing error message for duplicate check error, this shows links to the possible duplicates in the error message. 

Updates web-components to latest version (v0.7.12) for corresponding changes.


corsacca:
Thank you @cairocoder01!

PR #2729, opened 2025-08-14
Workflow Actions on Shares

### LINKED ISSUE #2722: Workflow Action: Add/Remove users from Sharing
**Is your feature request related to a problem? Please describe.**
I'm creating a workflow to re-assign new contacts to different dispatchers based on various criteria. I can easily set the Assigned field, but it still leaves the default user in the list of Share users. It would be nice to be able to remove that default user so they don't get any notifications that might use that list of Share users.

**Describe the solution you'd like**
Workflow action on "Sharing." Actions to Add or Remove users from the list.

**Describe alternatives you've considered**
None

**Additional context**
I've thought about another action to just reset that list based on the assigned user. However, If I assign a user in the same workflow, I'm not confident on the order of actions that it would do that correctly with the newly assigned user.

cairocoder01:
resolves #2722

Adds actions to be able to add/remove users from the list of users a post is shared with

### DIFF
--- a/dt-core/admin/js/dt-utilities-workflows.js
+++ b/dt-core/admin/js/dt-utilities-workflows.js
@@ -1077,6 +1077,16 @@ jQuery(function ($) {
           name: 'Add Comment',
         });
         break;
+      case 'share':
+        actions.push({
+          id: 'append',
+          name: 'Add Share',
+        });
+        actions.push({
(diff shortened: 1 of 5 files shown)


corsacca:
Works great! Thanks @cairocoder01!

Issue #2735, opened 2025-08-19
Field with "elements" key breaks the create form


corsacca:
If you create a field with the key 'elements',
the post-type/new won't work.

There is a clash with the js:
```
      Array.from(form.elements).forEach((el) => {
```

PR #2738, opened 2025-08-20
#2735 - Field With Elements Key Breaks The Create Form

kodinkat:
- fixes: #2735


corsacca:
@kodinkat 
It looks like this introduces an issue for the number fields:
there is an extra:
"": 2234

image

Please do a test for each field type

corsacca:
Thanks @kodinkat. Continued fix in: https://github.com/DiscipleTools/disciple-tools-theme/pull/2739

Issue #2743, opened 2025-09-01
Audio Comment Voice Message Transcriptions


kodinkat:
Provide the ability for users to transcribe audio comment voice messages.

PR #2745, opened 2025-09-03
#2743 - Audio Comment Voice Message Transcriptions

kodinkat:
- fixes: #2743
---

Screenshot 2025-09-03 at 17 45 10

Screenshot 2025-09-03 at 17 45 30


corsacca:
Hey @kodinkat, 
The lang selector really only is needed if AI is enabled.
So let's add a hook and add it from the plugin

PR #2749, opened 2025-09-08
Add User Apps tile to user contacts

cairocoder01:
For contacts that are connected to a user account, this will display a User Apps tile to users with the manage_dt permission. This allows admin users to copy and share user app links with their users easier than logging in as that user and going to their user profile.

### DIFF
--- a/dt-reports/magic-url-setup.php
+++ b/dt-reports/magic-url-setup.php
@@ -333,7 +333,7 @@ Thanks!', 'disciple_tools' );
                     <a type="button" class="app-copy dt-tooltip dt-action-button small button copy_to_clipboard"
                             data-value="<?php echo esc_url( $app_link ) ?>">
                         <span class="tooltiptext"><?php esc_html_e( 'Copy', 'disciple_tools' ); ?></span>
-                        <img class="dt-icon" alt="copy" src="https://dtplayground.local/wp-content/themes/disciple-tools-theme/dt-assets/images/duplicate.svg">
+                        <img class="dt-icon" alt="copy" src="<?php echo esc_url( get_template_directory_uri() . '/dt-assets/images/duplicate.svg' ) ?>">
                     </a>
                 </div>
             </div>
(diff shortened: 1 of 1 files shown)


dt-reports/magic-url-setup.php
```diff
@@ -202,5 +298,47 @@ function is_email_format_valid(email) {
...
+                    <a class="app-link dt-action-button small button view"
+                       href="<?php echo esc_url( $app_link ) ?>"
+                       title="<?php esc_html_e( 'link', 'disciple_tools' ) ?>">
+                        <i class="fi-link"></i>
+                    </a>
+
+                    <a type="button" class="app-copy dt-tooltip dt-action-button small button copy_to_clipboard"
+                            data-value="<?php echo esc_url( $app_link ) ?>">
+                        <span class="tooltiptext"><?php esc_html_e( 'Copy', 'disciple_tools' ); ?></span>
+                        <img class="dt-icon" alt="copy" src="https://dtplayground.local/wp-content/themes/disciple-tools-theme/dt-assets/images/duplicate.svg">
```
corsacca, line 336: Hard coded link. Lets us the same iconv as the Magic links

Image

corsacca:
I'm curious if there is a way to combine the 2 tiles and have them use the same UI.
Might be more work that it is worth. I did it for the D.T home app, going to look at what we did there.

I'm on dev, so there are more than a normal instance would have. The tile is already long and we're adding another one.
Hmm. Hard to capture. 
image

cairocoder01:
I agree on somehow consolidating. We would need a way to enable the user apps, since the Home app doesn't show in the Magic Links tile unless a user has enabled it. The User Apps tile allows that, but maybe we could combine it. There's also a minor difference in the first button: for ML tile it opens it in a modal to preview it, but the User App opens it as a normal link. I often want one-click to open it in a new tab, which the normal link allows but not the preview modal. I know I can copy the link, open new tab, and paste, but I'm lazy.

corsacca:
I'd be up for Changing the magic links functionality to open the magic link in a new tab instead of opening it in a modal.
I usually want that too

corsacca:
image

corsacca:
THanks @cairocoder01 
I really like this new accordion layout

Issue #2741, opened 2025-08-29
Show/Hide Archived posts in Advanced Search


cairocoder01:
**Is your feature request related to a problem? Please describe.**
When using the global/advanced search in the main nav bar, results include archived posts. For sites that may have had import issues and/or many archived contacts, this gets very confusing.

**Describe the solution you'd like**
The search on the contact/record list page excludes archived records by default with a toggle to include them. It would be great to include that toggle in the advanced search

<img width="2408" height="1374" alt="Image" src="https://github.com/user-attachments/assets/88e56e59-4666-4467-bb6f-ce9c6902a017" />

**Describe alternatives you've considered**

**Additional context**

PR #2752, opened 2025-09-11
#2741 - Show/Hide Archived Posts In Advanced Search

kodinkat:
- fixes: #2741 
---

Screenshot 2025-09-11 at 16 18 42

Screenshot 2025-09-11 at 16 19 53


cairocoder01:
@corsacca That's correct. They are hidden by default on the list pages (I think), so it would make sense if they were also hidden by default in the search.

corsacca:
Thanks @kodinkat and @cairocoder01!

PR #2756, opened 2025-09-17
Refresh activity after post updated

### LINKED ISSUE #2717: Record details page: field update -> activity in comments.
When a field is updated, the activity should automatically be reflected in the comments area.

cairocoder01:
fixes #2717

Also removes activation of the activity section spinner since each component has their own spinner now.


cairocoder01:
Oops, too forward thinking. I forgot we're not in the future yet. 👍

PR #2758, opened 2025-09-22
dt-location rendering

brady-lamansky-gtt:
@cairocoder01 This & the components PR I made should be everything. I tested it & had no issues on my end. Let me know if you see any problems!


corsacca:
After selecting a location, the typed text should be removed:
image

Issue #2761, opened 2025-09-25
Voice Record Message Error Message


corsacca:
https://github.com/user-attachments/assets/b1e085ed-e0cb-4bcc-bf23-aa53068fe499

No error message is shown if the voice record upload fails. 
(in the save the s3 connection has the wrong key).

PR #2766, opened 2025-10-03
#2761 - Voice Record Message Error Message

kodinkat:
- fixes: #2761 
---

Screenshot 2025-10-03 at 13 22 03

Screenshot 2025-10-03 at 13 33 45


dt-core/dt-storage.php
```diff
@@ -211,12 +211,20 @@ public static function upload_file( string $key_prefix = '', array $upload = [],
...
-            'uploaded_large_thumbnail_key' => $uploaded_large_thumbnail_key,
+                'uploaded' => true,
+                'uploaded_key' => $key,
+                'uploaded_thumbnail_key' => $uploaded_thumbnail_key,
+                'uploaded_large_thumbnail_key' => $uploaded_large_thumbnail_key,
+                'uploaded_msg' => null
             ];
         } catch ( Throwable $e ) {
-            return false;
+            return [
```
corsacca, line 221: @kodinkat. Since an error happened lets return an new error
`return new WP_Error`

In Admin UIs we can display the more details of the error message.
For user facing UI lets keep the error simpler. 
Something went wrong. Please try again or have an admin check the connections setting.

corsacca:
Thanks @kodinkat!

Issue #2767, opened 2025-10-06
List View User Contacts


corsacca:
User contacts are displayed by default in the list view. 
Users are also contacts so that they can be subassigned or used in the baptism and coaching relationship fields.

Having the user contacts there does lead to confusion. And the feedback from most admins is that they would prefer them not to be listed by default.

cairocoder01:
Or we totally remove users from the contact list. Then create a separate Users page. It's still a list of contact record but only of type user. Since there's already the user management page for admins, maybe we merge those features into tiles on the user contact record page to consolidate a little. Normal users can still click through to a user contact record, but they don't see all the advanced user management tiles.

corsacca:
Easiest to start would be a users entry in the filters. And users would be filtered out form the rest by detault.

cairocoder01:
Oh, or I think you were actually saying to add Users to the list of default users, right? That's different than I was thinking.

cairocoder01:
That makes sense and is probably the easiest first step.

PR #2769, opened 2025-10-07
Users in contacts list

corsacca:
Disable this: When in adding a new user, the new user contact record would automatically be marked as coached by the user who invited them. This ins't always accurate or wanted. Better place for a workflow.

Filters:

Add new "users" line to the filters list.

By default the All contacts doesn't show users:

image

Users will be show in the the Users filter.

image

The rest of the filters don't exclude or include users specifically. 

Fixes https://github.com/DiscipleTools/disciple-tools-theme/issues/2348
Fixes https://github.com/DiscipleTools/disciple-tools-theme/issues/2767

### DIFF
--- a/dt-contacts/base-setup.php
+++ b/dt-contacts/base-setup.php
@@ -679,7 +679,23 @@ class DT_Contacts_Base {
                     ]
                 ],
                 'query' => [
+                    'type' => [ '-user' ],
                     'sort' => '-post_date',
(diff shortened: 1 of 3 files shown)


cairocoder01 (approved):
Pulled this down and tested it. Looks good to me!

corsacca:
Thanks @cairocoder01

PR #2770, opened 2025-10-07
Assign access contacts by source

cairocoder01:
This adds the ability to configure a different base user to be assigned to new access contacts based on the Sources field.

image


corsacca:
looks good! Thanks @cairocoder01.

PR #2776, opened 2025-10-15
List Actions column

cairocoder01:
This updates the layout on the list pages of the "List Actions" that are initiated from the "More" menu. This includes Fields selection and Bulk Edit. Moving them out of the List panel helps to visually separate the actions from the list, especially for mobile.

On desktop...
- Actions are shown in right side panel
- Filters column is hidden while actions shown
- Actions split width 50/50 with list so that fields in bulk edit can split across multiple columns and not take a long page height

On mobile...
- Actions are shown in top panel
- Filters tiles are hidden while actions shown

# Mobile
## Current
screencast 2025-10-15 12-16-48

## Updated
screencast 2025-10-15 12-03-02

# Desktop
## Current
screencast 2025-10-15 12-13-52

## Updated
screencast 2025-10-15 12-07-27


cairocoder01:
- Added back blue bkrd to actions panel
- Fix grid layout of bulk send message and bulk app send
- Create XXL (>1440) list layout with filters still visible

> Questions:
> - when doing a bulk edit, would someone want to change the filters?

I think bulk editing is in itself a feature for power users. Power users are generally using desktop resolution sizes. So adding back the filters panel for >1440px should account for that now. Under that size, there isn't the screen real estate to keep all 3 visible (and I don't think people are wanting advanced features like bulk edits on smaller screens). I think it's more important while doing a bulk edit to have a good view of the list while also the action that you're performing. If you need to further filter, it's not too bad to close the actions and adjust filters to get what you want to see/edit. 

## 2560px
image

## 1440px
image

## 1024px
image

## 768px
image

## 425px
image

PR #2777, opened 2025-10-16
Mobile UI: List Filters and List second-bar nav

cairocoder01:
Branched from #2776, so this may need merging/rebasing once that is approved.

Reorganizes the second-bar nav and filters for small devices to free up vertical space.

# Second-bar Layout
- Search input always present
- Change search input clear button to be just overlay "x"
- Create .search-addon style for attached buttons. Move Search action as search-addon
- Increase height to make touch screen targets closer to 50px (recommendation I've seen)
- Remove Add New (moved to floating action button)
- Does not change anything with desktop second-bar, besides change to search input clear button

| Updated | Previous |
| --- | --- |
| image | image |
| image | image |

# Filter Tile as Bottom Sheet
- Moves Split By feature set into main filters tile, right before custom filters
- Move filters aside to bottom sheet style
- default filter tiles to collapsed to prevent mobile flicker

| Updated | Previous |
| --- | --- |
| image | image |


corsacca:
These are awesome changes! Makes the mobile UI so much better. 

Small issue
- When creating a custom filter, the filter sheet stays open. I think it would make sense for the filter sheet to be closed as well, so the user gets to the list that they filtered to. 

Question:
Should we get rid if this intermediate view? And just have mobile and desktop?
image

corsacca:
@cairocoder01, adde a commit for the filter sheet staying open. Let me now what you think.

Issue #2771, opened 2025-10-08
Location grid hierarchy is backwards


corsacca:
<img width="824" height="272" alt="Image" src="https://github.com/user-attachments/assets/26ca0062-42bd-45a7-b0fa-3a1501551e6e" />

PR #2782, opened 2025-10-20
#2771 - Location Grid Hierarchy Is Backwards

kodinkat:
- fixes: #2771


corsacca:
image


Thank you @kodinkat

PR #2783, opened 2025-10-22
Prevent addition of duplicate location_grid_meta

cairocoder01:
Revisiting #2689

After some further testing, I've found a reproducible test case to warrant this duplicate check in the location_grid methods.

## Steps to Reproduce
These steps assume a site configured with Mapbox and optionally Google geocoding keys.

Using the API endpoint for `/wp-json/dt-posts/v2/contacts/{id}`, send a body with the following:

```
{
	"contact_address": [
        {
            "value": "Madrid",
            "geolocate": true
        }
    ]
}
```

Load the contact record page and you will see Madrid added as a geocoded location.

Then, run the exact same API request attempting to add Madrid again.

Expected: Location is not added again because it already exists.
Actual: Exact same location_grid item is added. Looking in the database, you'll see postmeta and location_grid_meta rows with the same grid_id.

## Implemented solution
This code will check for an existing value that matches the geolocated value and prevent a duplicated value if it already exists.


corsacca:
@cairocoder01 Checking the label too would probably handle this case.
This will let us have multiple locations in the same location_grid

PR #2785, opened 2025-10-24
D.T Home Screen Migration Into Core Theme

kodinkat:
Screenshot 2025-10-24 at 17 49 36

Screenshot 2025-10-24 at 17 49 53

Screenshot 2025-10-24 at 17 48 14

Screenshot 2025-10-24 at 17 48 53


kodinkat:
Screenshot 2025-11-05 at 13 42 54

Screenshot 2025-11-05 at 13 43 31

Screenshot 2025-11-05 at 13 43 46

Screenshot 2025-11-05 at 13 44 34

Screenshot 2025-11-05 at 13 44 59

Screenshot 2025-11-05 at 13 45 42 Screenshot 2025-11-05 at 13 46 18

Screenshot 2025-11-05 at 13 46 38 Screenshot 2025-11-05 at 13 46 46

Screenshot 2025-11-05 at 13 47 16 Screenshot 2025-11-05 at 13 47 52

Screenshot 2025-11-05 at 13 50 25 Screenshot 2025-11-05 at 13 50 44

cairocoder01:
@kodinkat Looks like all of the bugs were fixed from my last review. Thanks! I'd still love to see a cleanup of the CSS important stuff.

A few more things I just found:
- I can't seem to edit the icon color of an app added from the ML plugin. For example, a post connections template I created. It seems to be consistent across the other templates I've used from there. I select a color and save, but it doesn't save. It fails for templates Post Connections, List Sub-assigned, and Single Record but succeeds for Create Record
- I think there's some layout/padding cleanup to do: (see image below)
  - There is body padding as well as an extra 2rem padding on each section. I think we may be able to remove the body padding if we're padding each child section
  - The welcome section doesn't line up with the Apps/Link sections. Header is 1.5rem padding and apps are 2rem. Those should be consistent and line up
  - I think we should be able to fit 4 icons wide on a large mobile device (~425px). Minimum recommended touch target size is about 1cm, which comes to about 50px. 90px is really comfortable for a touch target, but maybe we can decrease that to fit another icon. If we knocked overall padding to 1.5rem, that leaves 380px in the center. If we count on 1rem gap (currently 1.5rem) between 4 items, that's 335px for the icons, or 83.75px per icon. So we could go to 80px per icon to fit 4 wide on a large mobile device. If we went to 70px, we could fit 4 wide down to a medium mobile device too (375px). So I think it would be great to reduce the icon size to somewhere between 70-80px.

image

kodinkat:
@cairocoder01 thank you for the feedback - pls see latest updates and let me know if anything. :)

Screenshot 2025-12-11 at 13 47 10

Screenshot 2025-12-11 at 13 58 10

corsacca:
Thanks @kodinkat 

home-srceen.css has 2000+ lines of css, with 300+ uses of !important, and some duplication.
I think we can simplify and reduce.

One win might be to add `$allowed_css = []`; This would mean we don't have to fight the theme CSS. 

image

Do that for the js too and we'll just be loading what this Magic Link needs.

dt-apps/dt-home/assets/css/home-screen.css
```diff
@@ -0,0 +1,1315 @@
...
+        font-size: 1.75rem;
+    }
+    .home-screen-header p {
+        font-size: 0.85rem;
+    }
+    .theme-toggle-button,
+    .menu-toggle-button {
+        width: 36px;
+        height: 36px;
+    }
```
cairocoder01: This won't take effect unless the selector matches the one further above: `.home-screen-container .theme-toggle-button, .home-screen-container .menu-toggle-button`. This one will take precedence then because it is further down the file.

Issue #2728, opened 2025-08-14
Customizations Beta sets Roles to not editable


cairocoder01:
**Describe the bug**
When the Customizations (D.T) Beta interface is used to edit roles, the subsequent values saved to the database mark every role as `is_editable:false`. If a plugin adds a role in the wrong way, this could lock their roles from being edited properly.

**To Reproduce**
1. Go to 'Customizations (D.T) Beta'
2. Select a post type and go to Roles tab
3. Scroll to bottom and click Update
4. Open database table wp_options and find value for dt_custom_roles
5. Look at `is_editable` value for all roles and see that it's set to false `s:11:"is_editable";b:0;` (`b:0` means binary value of 0, thus false)

**Expected behavior**
Customizations admin should not set `is_editable` to false. It should remain the value given by the plugin or theme that defined the role. In most cases, this is undefined.

**Disciple.Tools Theme:**
- version: 1.72.0
- setup: local and single site production

**Additional context**
This was found with the Teams Module v1.1.2. It was resolved in 1.1.3. Using the former version, after saving the Roles in Customizations, the Teams roles were shown as Locked in the main Roles admin (`/wp-admin/admin.php?page=dt_options&tab=roles`)

PR #2786, opened 2025-10-30
#2728 - Customizations Beta Sets Roles To Not Editable

kodinkat:
- fixes: #2728 
---

@cairocoder01 - `is_editable` (and `custom`) properties, are now only updated, if already present within existing roles and permissions; allowing 3rd parties to update accordingly.

Changes are now tracked; enabling targeted role updates.


corsacca:
Thanks @kodinkat @cairocoder01!

PR #2799, opened 2025-11-25
#2794 - See GenMapper On Every Group Record

### LINKED ISSUE #2794: On every group, see genmapper
On the group record see the genmapper for the group  and the group's children

kodinkat:
- fixes: #2794 
---

Screenshot 2025-11-25 at 17 16 40


corsacca:
Nice work @kodinkat!

Here are some Notes/questions. These aren't decisions yet, just things we need to connect on as a group.
- in the 2 column mode, the rest of the tiles don't distribute. they are all in the same column.
- We might consider disabling scrolling, It can make the page navigation harder. 
- All groups are created as a pregroup. We might want more fields when adding a child group.
- We always display the root group as gen zero, This might be confusing, As we might be dealing with a second generation group.
 - Do we want to display any health icons on the node? 
 
 @cairocoder01 your thoughts or more questions welcome

corsacca:
Thanks @cairocoder01 
Consensus todos:
- default single column
- upgrade icon
- explore zoom controls
- remove gen numbers

Zoom. I've seen this on another project: scroll could be disabled until the user clicks into the canvas. This reduces a bit of the frustration when trying to scroll past the GenMapper, and still keeps the scroll functionality to go in and out. 

Connecting to an existing group. Would this be from clicking on a record with a separate button? or would the add model have the option to connect to an existing group? 

More discussion needed on group health icons and displaying the group type. What happens with custom group health option and custom group types?

single-template.php
```diff
@@ -303,9 +303,13 @@ function dt_display_tile( $tile, $post ): bool {
                     <div class="cell small-12">
                         <div class="grid-x grid-margin-x grid-margin-y grid">
                             <?php
+                            // Render genmap first if it exists
```
corsacca: @kodinkat this hook will now be called twice. Any plugin that uses this hook will have content rendered twice.
What is the catch with registering the tile the normal way and hooking into dt_details_additional_section for displaying the content?

Registering the tile the normal way will also let admin hide the tile if they want. And we can provide a description shown when the user click the help icon.

corsacca (changes requested):

single-template.php
```diff
@@ -377,8 +373,7 @@ function dt_display_tile( $tile, $post ): bool {
                                         </div>
                                     </div>
                                 </section>
-                            <?php }
-                            do_action( 'dt_record_bottom_after_tiles', $post_type, $dt_post ); ?>
+                            <?php } ?>
```
corsacca: @kodinkat don't deleted the original dt_record_bottom_after_tiles

corsacca:
Thanks @kodinkat! And thanks for the reviews @cairocoder01!

PR #2802, opened 2025-11-27
fix: csv import to allow 0 items of sample data and field options

squigglybob:
## Fixes
* issues with any key selects having 0 as a key.
* issues with CSV importer and 0's in data column and mapping to key selects with 0's as keys


corsacca:
Thanks @squigglybob!

PR #2803, opened 2025-12-02
Fix/zeros to work in key selects

squigglybob:
@corsacca I had thought I had this in the other PR, but it was on a seperate one.


corsacca:
Thanks @squigglybob

Issue #2806, opened 2025-12-04
Mapbox key warning


corsacca:
When mapbox key is limited to a particular url, then in the mapping admin this message is shown:

<img width="498" height="59" alt="Image" src="https://github.com/user-attachments/assets/d8e52d02-65db-4d5d-b5cc-30b6f4c1aca1" />

PR #2808, opened 2025-12-05
#2806 - Mapbox Key Warning

kodinkat:
- fixes: #2806 
---

Screenshot 2025-12-05 at 14 20 39


kodinkat:
@corsacca see latest updates; which provides additional information within error messages and now also includes the site url within a **Referer** header; which mapbox required to successfully verify against url restriction list.

Screenshot 2025-12-10 at 11 17 24

PR #2815, opened 2025-12-12
dt-toggle component

brady-lamansky-gtt:
@cairocoder01 dt-toggle theme changes, allows the new dt-toggle web component to be used in the theme


cairocoder01:
Spacing is updated now. It is set to a minimum of 2.5rem to match the height of all other components. This way, when in a row with other components, the match the default height of each other.

image
image

corsacca:
Thanks!

PR #2818, opened 2025-12-19
Location js cleanup

cairocoder01:
This removes the usage of the typeahead location fields and widget js files in the theme. I wanted to delete them completely, but there are still plugins that make use of them. Specifically, it replaces the location fields on the profile and user management screens, and then removes the scripts from the detail, list, and create pages that have already been replaced and aren't in use any more.

I still need to push some updates to the component library but don't want to do that right before taking off for 2 weeks of holiday, so I'm creating this but it shouldn't be merged yet.


cairocoder01:
@corsacca This should be ready for your review. It affects the location fields on the profile and user screens. I added deprecation messages to the class methods that are no longer used in the theme but are still used by plugins.

PR #2819, opened 2026-01-05
List layout bugs

cairocoder01:
Fixes two bugs on list screen as reported in Discord:

- When result list is shorter than bulk edit panel, bulk edit panel gets cut off
- In bulk edit panel, typeaheads aren't visible


corsacca:
Thanks @cairocoder01!

PR #2820, opened 2026-01-05
#2812 - Connection Fields Squished Layout

### LINKED ISSUE #2812: Connection fields squished layout
On smaller screens:
Image

kodinkat:
- fixes #2812 
---


kodinkat:
> @kodinkat based on the amount of !important and css needed, It looks like the change might need to be done on the components themselves instead of from the theme CSS?

@corsacca hmm, agreed - I think that may be a better approach.

@cairocoder01 would you also agree? I don't think I've had a chance with working directly with the dt-components codebase as yet - any pointers re how best to test changes locally?

cairocoder01:
@kodinkat You might check out this recent issue/PR that was completed as it might be related to this or even fix it already:

https://github.com/DiscipleTools/disciple-tools-web-components/pull/174

That should have been released in v0.8.5. So if that's the fix, maybe the version of the components just needs to be updated. Or that fix needs to be adapted to go even smaller.

kodinkat:
@corsacca I can confirm the issue has been resolved within the stated fix above; as I also had to run `npm install` in order to update versions within my local environment.

Thanks again @cairocoder01 for the assist. :)

I guess this PR can be closed.

corsacca:
Thanks Guys!

Issue #2684, opened 2025-05-26
Move new release notifications.


corsacca:
When a new release is available, a modal is shown and users dismiss it right away.
Instead of auto showing the modal, lets add a notification about the new release notes in the nav bar.
Users click on the notification and the modal is show.

PR #2821, opened 2026-01-05
#2684 - Move New Release Notifications

kodinkat:
- fixes: #2684 
---

Screenshot 2026-01-05 at 17 27 44


corsacca:
@kodinkat, I'd like the new release notification icon to be separate from the existing notifications icon. Clicking the icon opens the modal, and then the icon disappears.

kodinkat:
> @kodinkat, I'd like the new release notification icon to be separate from the existing notifications icon. Clicking the icon opens the modal, and then the icon disappears.

Screenshot 2026-01-06 at 11 05 10

Issue #2832, opened 2026-01-09
Home screen. App vs Link


corsacca:
When adding a new app, the user can choose between App and Link.
We need to communicate what the difference is and why the user would choose one over the other.

PR #2840, opened 2026-01-13
#2832 - Home screen. App vs Link

kodinkat:
- fixes: #2832 
---

Screenshot 2026-01-13 at 15 14 05

Screenshot 2026-01-13 at 15 14 39

### DIFF
--- a/dt-apps/dt-home/assets/css/admin.css
+++ b/dt-apps/dt-home/assets/css/admin.css
@@ -24,8 +24,7 @@
     margin-top: 15px;
 }
 
-.apps-list,
-.training-list {
+.apps-list {
     margin-top: 20px;
 }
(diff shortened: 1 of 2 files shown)


kodinkat:
> @kodinkat is the only difference if the app is opened in a new tab?
> 
> Would it be simpler to just have a checkbox "Open in new tab?"

@corsacca yep, as it appears to be a legacy requirement. 

We could present it as a checkbox within the admin frontend; but to minimize code refactoring, the checkbox selection will simply translate back into a `link` or `app` type - agreed?

Also, another thought - keeping it as a dropdown does give us the flexibility to support different types in the future.

kodinkat:
@corsacca review ready.

Issue #2834, opened 2026-01-09
Home Screen - role based access setting


corsacca:
<img width="751" height="111" alt="Image" src="https://github.com/user-attachments/assets/4aa896f9-f3be-4db2-a1fc-6b8188e2947b" />

Let's just have this enabled by default. We can remove this checkbox. 

On each magic link, the default is that all rules can access the magic link. if the admin wants they can I restrict it to just some users.

PR #2842, opened 2026-01-15
#2834 - Home screen - Role Based Access Settings

kodinkat:
- fixes: #2834


kodinkat:
@corsacca - review ready....

Issue #2864, opened 2026-01-22
List bulk edit. Connection field calling 2 endpoints for Options.


corsacca:
<img width="389" height="65" alt="Image" src="https://github.com/user-attachments/assets/f96e111e-8b31-448b-b561-cafaa2a474f2" />

PR #2871, opened 2026-01-23
#2864 - List Bulk Edit - Connection Field Calling 2 Endpoints For Options

kodinkat:
- fixes: #2864


corsacca:
Thanks @kodinkat!

PR #2875, opened 2026-01-30
#2854 - Field For Uploading Files

### LINKED ISSUE #2854: Field for uploading files.
Create a field for uploading files.
These could be pictures, documents, pdfs, voice messages, etc.

Since all D.T fields are components we could make this new field a component too.

The field will hold multiple files.

kodinkat:
- fixes: #2854 
- see: https://github.com/DiscipleTools/disciple-tools-web-components/pull/193
---

Screenshot 2026-01-30 at 17 08 34

Screenshot 2026-01-30 at 17 09 14

Screenshot 2026-01-30 at 17 10 14

Screenshot 2026-01-30 at 17 13 36

Screenshot 2026-01-30 at 17 14 57

Screenshot 2026-01-30 at 17 17 31

Screenshot 2026-01-30 at 17 18 48


kodinkat:
@corsacca @cairocoder01 latest updates ready for review....

cairocoder01:
Testing just to confirm the front-end and UX:

- [x] A bit of an edge case, but if I double click to rename a file, make a change, but then hit ESC to try to cancel the change, it seems to go through and do the renaming anyways. When hitting ESC, it should revert back to the existing file name and not make any changes.
- [x] Is there a way to add a better upload progress indicator that shows some sort of percent completed? I just uploaded a 7.3MB image file that took close to 2min to complete. The spinner showed as expected, but I started questioning if it was stuck. I don't know if there's actually a way to get a progress indication back or not, but that would be ideal so people don't give up too soon and hit refresh or upload the file again. And would we want to disable the upload area while a file upload is in progress?
- [x] On the list page, if I add a file_upload field to the field list that should show in the table, nothing is displayed. To keep it simple, maybe that just shows the number of files instead of trying to display them.
- [x] On the list page, there is now way to filter by a file_upload field. Maybe this is ok for now and we do that later. It would be good to be able to filter for records with/without files saved, or maybe also the number of files.
- [x] In the Customizations admin when editing a file_upload field, the modal window is taller than my viewport and won't scroll down, so I can't properly save changes.
- [x] When creating a new record and trying to upload a file in the process, the API request to upload it fails. However, it shows the file in the list as if it was successful. Then when I hit save and go to the record, it also shows the file in the list but the thumbnail is broken because the file doesn't exist. 
  - [x] Failed API requests should cause the list/thumbs to update accordingly. 
  - [x] On the new record page, the API endpoints should not be hooked up and the file content needs to be sent as part of the form submission, where PHP can pick up the file and save it through that process.
image
image


I haven't reviewed any of the backend code for its implementation. I'll leave that to @corsacca

kodinkat:
@cairocoder01 @corsacca review ready....

Screenshot 2026-02-20 at 14 34 46

Screenshot 2026-02-20 at 14 35 27

cairocoder01:
- [x] The upload progress indicator seems to show 100% immediately and doesn't change. If there isn't a good way to show that progress, I think I'd rather not have it at all. Or maybe it just isn't working correctly. I think the server would need to send back some sort of partial response or something to show the progress. That might be hard on a Wordpress site. So if that's not possible, we can remove that idea.

kodinkat:
@cairocoder01 review ready...

cairocoder01:
This looks good to me. Ready for @corsacca 

When all looks good, I can merge the associated PR on the component library first, publish a release, and then update the version here in this PR before merging.

corsacca:
Thanks guys!

@kodinkat 
image
By default we are allowing images and pdfs. 
It is unclear how to enable:
- audio files
- videos
- documents (docx, .txt, .md, spreadsheets, etc)

I think we can enable those by default. No?
With a link to somewhere listing out other file types.


File type icons
image

It looks like we are getting the icons working by default:
image

I don't think we need the extra input for making it different.

corsacca:
Otherwise the upload field is looking good! I don't think the above changes require changing the component, so we might be good there. @cairocoder01

cairocoder01:
@corsacca Sounds good. I merged the component request and created release v0.8.10 with this new component in it. Available from NPM now.

kodinkat:
@corsacca review ready.....

Screenshot 2026-03-04 at 14 43 45

corsacca:
Can't get test to pass. Not related to changes is this PR.

corsacca:
Thanks @kodinkat!!!

PR #2886, opened 2026-02-09
Move `dt_home_screen_apps` option init so it doesn't reset value

### LINKED ISSUE #2885: Home App Admin - Unable to disable coded app if no custom apps exist
**Describe the bug**
I found on a site with only coded apps added to the home screen, when I tried to disable one of the apps in the admin, the setting wouldn't save. After the initial save, the UI would show that it is disabled, but after refreshing the page it was enabled again.

**To Reproduce**
1. Create a local site with no custom home apps - only the default coded apps from the ML plugin (user_contacts_updates and user_groups_updates)
2. In Home Screen admin, edit one of the coded apps and uncheck Enabled
3. Click Save
4. After page load, see app shows as Disabled in the list
5. Refresh the page
6. See that the app has gone back to enabled.
7. Add a custom app called Test with default values
8. See new app is added to list
9. Attempt to disable coded app again
10. See it show as disabled after page load
11. Refresh
12. See it is still showing as disabled

**Expected behavior**
Coded apps can be disabled even when there are no custom apps

**Screenshots**

**Disciple.Tools Theme:**
- version: 1.78.0
- setup: local single site

cairocoder01:
resolves #2885
If no custom apps are created and `dt_home_screen_apps` wp_option is empty, the init function seems to run too soon before coded apps are registered and causes the option value to be overwritten. This moves it after the hook for registering coded apps.

### DIFF
--- a/dt-apps/dt-home/includes/class-home-apps.php
+++ b/dt-apps/dt-home/includes/class-home-apps.php
@@ -29,9 +29,6 @@ class DT_Home_Apps {
     }
 
     public function __construct() {
-        // Initialize with default apps if none exist
-        $this->initialize_default_apps();
-
         // Register init hook to handle filter timing
         add_action( 'init', [ $this, 'on_init' ], 20 );
     }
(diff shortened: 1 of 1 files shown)


kodinkat:
@cairocoder01 - Approved — Good work. Moving initialization after the coded apps registration hooks, resolves the timing issue and prevents the option from being reset prematurely.

corsacca:
Thanks guys!

PR #2887, opened 2026-02-10
Refactor vite script compilation

cairocoder01:
Converts the bundling of scripts to generate an IIFE (standard browser compatible) style script instead of a module with a legacy version. This better supports the jquery external dependency that other site scripts rely on.

To do so, this separates the script bundling from the SCSS compilation via separate Vite config files. Both are still run on `npm run build` so it is a seamless experience. However, the IIFE build process is unable to separately package CSS files, so those things need to be separated.

Script loads in browser with no console errors, and the advanced search button successfully opens the Foundation modal - indicating Foundation is now being initialized with jQuery correctly.


corsacca:
@cairocoder01 
Is this recommended update needed?

```
"build": "vite build && vite build --config vite.config.scripts.js",
"build:css": "vite build",
"build:scripts": "vite build --config vite.config.scripts.js"
```

cairocoder01:
> @cairocoder01 Is this recommended update needed?
> 
> ```
> "build": "vite build && vite build --config vite.config.scripts.js",
> "build:css": "vite build",
> "build:scripts": "vite build --config vite.config.scripts.js"
> ```

@corsacca I wouldn't say needed, but it's another way to configure the different build tasks. It would separate out the CSS build to be able to run it on it's own apart from the scripts, so it certainly could be useful. It's assessment that devs won't get the js changes is wrong because the `postbuild` script runs automatically after the `build`. Its suggestion might be more clear though.

corsacca:
Thanks! Seem like something that can be tweaked later if confusion happens

PR #2890, opened 2026-02-16
#2878 - Group Genmapper

### LINKED ISSUE #2878: Group Genmapper
Image

Remove generation number.

When adding a new group, show more options, specifically 
- group type
and whatever the admin has configured as the default group fields, the ones that show up when creating a new group. 

Right now, all new groups are created as pre-group by default. 

Image

kodinkat:
- fixes: #2878 
---

Screenshot 2026-02-16 at 13 16 50

Screenshot 2026-02-16 at 13 30 56


kodinkat:
@corsacca review ready...

corsacca:
thanks @kodinkat!

Would you be able to check why the location dropdown get hidden?
image


Also please evaluate the following:

**1. XSS risk in mapbox token injection**
  fieldHtml = fieldHtml.replace(
    /mapbox-token="[^"]*"/,
    `mapbox-token="${window.lodash.escape(mapboxKey)}"`,
  );
  Using string .replace() on rendered HTML to inject attribute values is fragile. If renderField
  ever changes its output format (e.g., single quotes, no quotes, different attribute order), this
   silently breaks. Consider setting the attribute on the DOM element after insertion instead:
  el.setAttribute('mapbox-token', mapboxKey);

**2. alert() for validation — genmap-tile.js**
  alert('Name is required');
  This is a browser-native alert() — it blocks the thread and isn't translatable. The rest of the
  file uses window.dtGroupGenmap.strings for i18n. This string should be added to the strings
  array in PHP and referenced here.
-> make translatable

**3. No loading/disabled state on submit**
  The submit button (#group_genmap_add_child_but) is not disabled during the API call. A user can
  double-click and create duplicate child groups. The original code had the same issue, but since
  the form is now richer and slower to fill, it's more likely users will notice the lag and click
  again.
  -> make the button a loader button (like existing examples), with active class

kodinkat:
@corsacca review ready....

corsacca:
@kodinkat the location options list still is cut off. Do you get the same behavior?

image

kodinkat:
> @kodinkat the location options list still is cut off. Do you get the same behavior?
> 
> image

@corsacca no, I'm not seeing this? Anything in the console?

kodinkat:
@corsacca location options no longer cut off....

Screenshot 2026-03-09 at 10 45 25

corsacca:
Thanks @kodinkat!

PR #2893, opened 2026-02-18
Multi text groups

jlamanskygitt:
Implemented multi-text-groups changes for the theme


cairocoder01:
@jlamanskygitt It looks like you created a new field type for this component. It should actually be used for the existing `link` field type and just replace that existing UI.

dt-core/utilities/dt-components.php
```diff
@@ -321,4 +321,14 @@ public static function render_user_select( $field_key, $fields, $post, $params =
         </dt-users-connection>
         <?php
     }
+
+    public static function render_link( $field_key, $fields, $post, $params = [] ) {
+        $shared_attributes = self::shared_attributes( $field_key, $fields, $post, $params );
+        ?>
+        <dt-multi-text-groups <?php echo wp_kses_post( $shared_attributes ) ?>
+            value="<?php echo esc_attr( isset( $post[$field_key] ) ? json_encode( $post[$field_key] ) : '' ) ?>">
```
cairocoder01, line 481: This isn't passing the list of possible groups, so I'm unable to add values to different groups

dt-core/global-functions.php
```diff
@@ -834,34 +727,6 @@ class="text-input" value="<?php echo esc_html( $post[$field_key] ?? '' ) ?>" <?p
         do_action( 'dt_render_field_for_display_template', $post, $field_type, $field_key, $required_tag, $display_field_id, $custom_display, $fields );
     }
 
-    function render_link_field( $field_key, $option_key, $value, $display_field_id, $meta_id, $required_tag, $disabled ) {
-        ?>
-
-        <div class="input-group">
-            <input
-                type="text"
-                class="link-input input-group-field"
```
cairocoder01, line 846: Please do a search in the project for `link-input` to find all of the javascript that is run based on that class. We need to test that we can remove that outdated js code without affected functionality.

dt-core/utilities/dt-components.php
jlamanskygitt: Addressed in newest commit

cairocoder01:
Functionally, this looks great. Really nice work! Just waiting on the cleanup of the old code.

cairocoder01:
- [x] When attempting to merge two contacts, if one has no value for a multi-text-group and the second does have a value, when I check the box next to the second, the value does not show up in the third column that shows the updated merged value

Otherwise it seems to be looking good. There are branch conflicts, so make sure you merge the develop branch back into this branch again to get it up-to-date.

cairocoder01:
@jlamanskygitt This is almost there. It's merging the values together well. Here are some kinda edge cases I found that we should handle.

- [ ] Can't removed values from archived record
  - Selected both values for a link field.  Value merged correctly into third column. 
  - Removed a value from primary record. Success. 
  - Removed value from archived record. Did not get removed. (Do we need temp ID values added to them?)
- [ ] Handle newly added values
  - Selected both values for a link field. Value merged correctly into third column. 
  - Added new value to first group. 
  - Unchecked archived value. Just the archived values removed (as expected). 
  - Rechecked archived value. Values added back correctly. 
  - Unchecked primary record. Primary values removed along with the new value I just added (not expected)

Other than these, the merge seems to be working correctly and saves the updated value as expected.

cairocoder01:
@jlamanskygitt Look at the "Medium" issue in the AI code review above. It identifies the bug that I just noticed. Once you un-check and then re-check the field from archived record, you can no longer manually remove it from the third column field.

@corsacca Beyond this one edge case, I think this is ready for your review.

cairocoder01:
@jlamanskygitt This all looks good to me now. Thanks and nice work!
@corsacca This is all ready for you.

corsacca:
The amount of code removed by this PR is satisfying. 
Well done and thank you @jlamanskygitt

PR #2894, opened 2026-02-19
#2788 - Expiring Magic Links

### LINKED ISSUE #2788: Expiring Magic Links
**Is your feature request related to a problem? Please describe.**
When sending a magic link to a multiplier in order to update the details of a contact record, we don't want that link/URL to exist forever for the sake of security. It would be best if it could be set to expire after a chosen amount of time. This is partly available via the Magic LInk plugin and using the "Links" option to auto-generate time-specific links for users, but that doesn't work if you want a time-specific link for a contact record that isn't a user.

**Describe the solution you'd like**
In some way from the Magic Links tile on a contact record (or any record type), have the ability to select a time period (numeric quantity + period selection: days, weeks, months) that will generate a separate link that will expire after that given time. A list of those time-specific links could be listed at the bottom of the accordion.

**Describe alternatives you've considered**
Alternately, maybe you can just set an expiration for a given magic link app within the accordion. That changes the link used for all of the send features (link, copy, send by email, QR code) to have an expiration. At the time of that expiration, the link gets regenerated and goes back to a persistent link without expiration. If a link has an expiration, it should be displayed within the Magic Link tile, possibly within the title row so that it's clear even with using the link icon that it will expire.

**Additional context**

Image

kodinkat:
- fixes: #2788
---

Screenshot 2026-02-19 at 11 53 17


kodinkat:
@corsacca @cairocoder01 review ready....

cairocoder01:
I love this! Nice work!

One minor comment: the expiration date in the accordion header feels a bit crowded. Could we use a shortened date format since the full date is inside the accordion? Something like `Exp: 2/23/26 14:01`? Also, it would be good if that date was in the local time zone instead of UTC - either the WP site timezone or client-side js time zone detection.

image

Also, what is the difference between setting a link to never expire or just hitting reset to clear the expiration date? By default, they don't expire. But when you set it to not expire, it explicitly says that it never expires. I'm a bit confused at the difference, so I'm guessing users will be too.

kodinkat:
@cairocoder01 see latest updates..... Also, good spot re "Never expires"; which, on further reflection, is actually redundant; whilst the "Reset Link" functionality is present - Therefore, I've decided to remove it.

Screenshot 2026-02-24 at 13 15 12

cairocoder01:
Looks good to me. I wonder about the verbiage of the Reset label. The use of "also" for some reason feels somewhat unprofessional or something. Should it be something like "Reset link (clears expiration)" or "Reset link (and clear expiration"? And should that verbiage only show if there is a current expiration (I haven't tested myself to know if this is already the case)? So if there's no expiration set, it just says "Reset link." If there is an expiration date set, it shows the extra info about clearing expiration. I'm unsure the best route and these are just thoughts.

Regardless of that, I'm happy with this however that verbiage ends up. So on to @corsacca to review.

kodinkat:
> Looks good to me. I wonder about the verbiage of the Reset label. The use of "also" for some reason feels somewhat unprofessional or something. Should it be something like "Reset link (clears expiration)" or "Reset link (and clear expiration"? And should that verbiage only show if there is a current expiration (I haven't tested myself to know if this is already the case)? So if there's no expiration set, it just says "Reset link." If there is an expiration date set, it shows the extra info about clearing expiration. I'm unsure the best route and these are just thoughts.
> 
> Regardless of that, I'm happy with this however that verbiage ends up. So on to @corsacca to review.

Hey @cairocoder01 yep, I'll also tweak verbiage and have its display made conditional based on any existing expiration dates. :)

corsacca:
I realize i don't have the plugin updated, but so might others.

Should the relevant code be in the theme instead?

image

kodinkat:
> I realize i don't have the plugin updated, but so might others.
> 
> Should the relevant code be in the theme instead?
> 
> image

Moving those methods into the theme would:

- Duplicate core plugin logic (expiration storage, meta keys, etc.).
- Risk divergence/conflicts once the plugin is updated to a newer version that also ships these methods.

A tighter, safer fix is to feature‑detect the “new expiration API” and only use it when it’s actually available.

....review ready...

corsacca:
@kodinkat, please display the date in the user's time (pass epoch and let js in the bowser display it). 
How often does the service run that expires the links? Set set one for 1 min can access it afterwards.

kodinkat:
@corsacca  ....review ready...

Screenshot 2026-03-23 at 15 24 46

Screenshot 2026-03-23 at 15 28 20

Screenshot 2026-03-23 at 15 28 53

corsacca:
thanks @kodinkat!

PR #2899, opened 2026-03-03
#2895 - Bulk Edit - User Select Typeahead

### LINKED ISSUE #2895: Bulk Edit: User select typeahead
Make user select use the dt component and remove all typeahead code

kodinkat:
- fixes: #2895 
---

Screenshot 2026-03-03 at 12 22 18


kodinkat:
@corsacca review ready....

dt-assets/js/modular-list-bulk.js
```diff
@@ -918,6 +923,81 @@
     return Array.isArray(value) ? value : [];
   }
 
+  /**
+   * Normalize a user_select value to the "user-{id}" string format expected by
+   * the backend and conditional-removal logic. Accepts values from
+   * ComponentService.convertValue(dt-users-connection), dt-users-connection.value,
+   * or legacy shapes.
+   */
+  function normalizeUserSelectValue(rawValue) {
```
corsacca, line 926: @kodinkat this looks like a whole lot of code handling maybes.
Can we double check the actual values being passed arround?

corsacca (changes requested):

dt-assets/js/modular-list-bulk.js
corsacca: @kodinkat, my question still stands. Why handle every single possible value, instead of pulling from what the component actually returns?

PR #2904, opened 2026-04-06
Adjust home screen bottom navigation height and padding

cairocoder01:
- Increase `--launcher-bottom-nav-height` to 65px to ensure consistent layout, replacing a min-height that was not tracked in a variable, producing inconsistent bottom padding. Effectively decreased launcher bar height from 86px to 65px.
- Update wrapper padding and simplify selectors to provide better scroll clearance for the bottom navigation bar.

# Issue Description
When using a Create Contact template, I noticed there was a double scroll bar on the side. This was a template that had enough field to cause it to need to scroll.

## Before
image

## After
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
       with:
         node-version: 22.x
         cache: 'npm'
-    - run: npm install -g vite eslint @babel/eslint-parser
+    - run: npm install -g eslint @babel/eslint-parser
 
     - name: Install PHP
       uses: shivammathur/setup-php@master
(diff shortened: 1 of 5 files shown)


cairocoder01:
@corsacca Please take a look at this when you can. Fixes not just some scroll weirdness on the home screen, but also some build things that surfaced when pushing it. 

Vite has released v8 but we have v7 installed, so the CI build script had a mismatch from the globally installed version versus the one from node_modules used in `npm run build`. So the other changes outside the CSS fix that, which is also failing on other PRs.

corsacca:
Thanks @cairocoder01!

Issue #2898, opened 2026-02-24
Change in web component height doesn't update masonry layout


cairocoder01:
**Describe the bug**
When a web component that can change in height because of added items is updated so that its height changes, if there is another tile below the one it is in, the current tile increases in height and goes behind the one below it.

**To Reproduce**
1. Install plugin starter on local site so that "A Beautiful Tile" is added, or create a new tile that is first in the list that includes a multi-text/communication_channel component (or any other multi-select based component that can increase in height with multiple values selected.)
2. Ensure there is a tile directly below the new one.
3. Add values to the component until it increases in height
4. See the tile expand in height and go behind the one below it.

**Expected behavior**
Tile below adjusts position based on changed height of current tile. Masonry script should be run when any component update happens.

**Additional context**
We should be able to intercept any change events on the record detail screen and re-run the masonry init script when that happens.

corsacca:
related: https://github.com/DiscipleTools/disciple-tools-theme/issues/2857

PR #2908, opened 2026-04-09
Re-layout masonry when fields are resized

cairocoder01:
https://github.com/DiscipleTools/disciple-tools-theme/issues/2898

Related to https://github.com/DiscipleTools/disciple-tools-web-components/pull/198 which removes the resize observer inside dt-file-upload, since this branch puts in in the theme js


corsacca:
Thanks @cairocoder01!

PR #2911, opened 2026-04-13
#2846 - Replace Components User Profile

### LINKED ISSUE #2846: Replace components: user profile
On the user profile page (`/settings`), replace old form controls with the appropriate web components.

Replacements:
- Your Apps
  - [ ] Use dt-toggle for app activation toggle
- Multiplier Preferences
  - [ ] Languages selection - multi-select-button-group
  - [ ] People Groups - multi-select
- Notifications
  - [ ] Use dt-toggle for toggles
 
kodinkat:
- fixes: #2846 
---

Screenshot 2026-04-13 at 17 20 46

Screenshot 2026-04-13 at 17 21 12

Screenshot 2026-04-13 at 17 22 18


kodinkat:
@cairocoder01 @corsacca review ready....

dt-assets/js/settings.js
```diff
@@ -52,10 +58,180 @@ function app_switch(user_id = null, app_key = null) {
...
+      if (!byId.has(id)) {
+        byId.set(id, { id, label });
+      }
+    }
+    el.options = Array.from(byId.values());
+  };
+
+  const attachInputHandlers = () => {
+    const input =
+      el.shadowRoot && el.shadowRoot.querySelector('input[part=input]');
```
cairocoder01: Why are we selecting elements in the shadow root? That shouldn't be necessary from the theme. It seems the People Groups field is using a dt-multi-select but is loading records from the API. That should really be a dt-connection field then, which is meant to handle that sort of behavior. Is there a difference with this being a user record that makes that a problem?

template-settings.php
```diff
@@ -357,46 +358,60 @@
...
+                                    ];
+                                },
+                                array_keys( $contact_fields['languages']['default'] ),
+                                $contact_fields['languages']['default']
+                            );
+                            $settings_language_value = array_map( 'strval', $settings_user_languages );
+                            ?>
                             <div class="section-subheader cell" style="margin-top:30px">
                                 <img src="<?php echo esc_url( get_template_directory_uri() ) . '/dt-assets/images/languages.svg' ?>">
                                 <strong style="display: inline-block;"><?php esc_html_e( 'Languages you are comfortable speaking', 'disciple_tools' )?></strong>
```
cairocoder01: Instead of displaying these labels separately, can we use the integrated label properties of the components?

template-settings.php
```diff
@@ -357,46 +358,60 @@
...
-                                <?php endforeach; ?>
                             </div>
-
+                            <dt-multi-select-button-group
+                                id="settings-user-languages"
+                                name="languages"
+                                class="settings-user-languages-multiselect"
+                                value="<?php echo esc_attr( wp_json_encode( $settings_language_value ) ); ?>"
+                                options="<?php echo esc_attr( wp_json_encode( $settings_language_options ) ); ?>"
+                            ></dt-multi-select-button-group>
```
cairocoder01: Something might be off with this. It's not showing any options for me on my local site, though I do have language options set in the backend.

Image

Image

cairocoder01: Oh, I think the problem is the value property. It is rendering the value as an object instead of an array of strings.

Image
...instead of...
Image

template-settings.php
```diff
@@ -357,46 +358,60 @@
...
+                                            'label' => $row['post_title'],
+                                        ];
+                                    },
+                                    $settings_people_group_rows
+                                );
+                                $settings_people_groups_value = array_map( 'strval', $settings_people_group_ids );
+                                ?>
                             <div class="section-subheader cell" style="margin-top:20px">
                                 <img src="<?php echo esc_url( get_template_directory_uri() ) . '/dt-assets/images/people-group.svg' ?>">
                                 <?php esc_html_e( 'People Groups you wish to serve', 'disciple_tools' ); ?>
```
cairocoder01: Can we use the label properties of the component itself? See the location field above these two fields. It actually uses the DT_Components class to render out the appropriate HTML instead of doing it manually. That keeps things more consistent throughout the code.

dt-assets/js/settings.js
```diff
@@ -52,10 +58,180 @@ function app_switch(user_id = null, app_key = null) {
...
+      el.setAttribute('loading', true);
+      window
+        .makeRequestOnPosts('GET', 'peoplegroups/compact', { s: q || '' })
+        .done((data) => {
+          mergeCompactPosts(data.posts);
+        })
+        .always(() => {
+          el.removeAttribute('loading');
+        });
+    };
```
cairocoder01: Why are we fetching the list of values via API when they've already been set inline in the options attribute? We shouldn't need to do both. Either we load them via API or we expect them to be static in the options attribute.

cairocoder01 (changes requested):

dt-assets/js/settings.js
kodinkat: People Groups on settings is now a dt-connection with postType="peoplegroups", which is the right primitive for “search records from the API and attach them.”

The shared ComponentService dt:get-data handler filters out items whose post id equals this.postId. On user settings, this.postId is the current user id. A people group is a post; if a group’s post ID happens to match the user ID, that row was incorrectly dropped from search. We handle dt:get-data on #settings-people-groups in the capture phase, load peoplegroups/compact ourselves, and call stopImmediatePropagation() so the default handler does not run for this field only. That avoids shadow DOM hacks and fixes the ID collision edge case.

User records: This field is still “people group posts,” not “user” post type; using dt-connection to peoplegroups is appropriate. Nothing about it being on a user settings screen requires dt-multi-select or shadow queries.

dt-assets/js/settings.js
kodinkat: We’re not duplicating options and API: the template only sets value (current selections). There is no options attribute on this dt-connection. peoplegroups/compact is only for typeahead when searching for more groups to add. That differs from languages, where the working set is a small fixed list rendered as options on dt-multi-select-button-group. We still need this handler (and stopImmediatePropagation) so ComponentService doesn’t filter out people-group posts whose ID matches the logged-in user ID.

cairocoder01:
This looks good to me.

@kodinkat I updated the rendering of the components to use the shared `DT_Components` helper class for consistency across the theme.

corsacca:
Thanks @kodinkat and @cairocoder01!

PR #2912, opened 2026-04-17
support MDI icons in dt-label

cairocoder01:
Resolves 404 errors on record detail screen for fields using MDI icons. Changes were made in version 0.8.12 of the component library.

On one site using Cloudflare and WPEngine Global Edge Security, the failed requests of multiple MDI icons trying to be retrieved as image URLs caused rate limited that locked out users for a few minutes. Hopefully this should help to resolve that.


corsacca:
Thanks @cairocoder01!

PR #2913, opened 2026-04-20
Reset health circle option icon when alternate set

cairocoder01:
User reported that after creating a custom health circle icon with MDI icon and trying to change it to a custom image, the MDI icon continued to show. I found that the `font-icon` property wasn't cleared when and `icon` was added. This resolves that by clearing the other property when one of them is set.

### DIFF
--- a/dt-core/admin/admin-settings-endpoints.php
+++ b/dt-core/admin/admin-settings-endpoints.php
@@ -975,14 +975,14 @@ class Disciple_Tools_Admin_Settings_Endpoints {
                     'description' => $new_field_option_description,
                 ];
 
-            if ( $field_option_icon ){
+            if ( $field_option_icon ) {
                 $field_option_icon = strtolower( trim( $field_option_icon ) );
                 $icon_key = ( strpos( $field_option_icon, 'mdi' ) !== 0 ) ? 'icon' : 'font-icon';
-                $custom_field_options[$post_type][$field_key]['default'][$new_field_option_key][$icon_key] = $field_option_icon;
+                $null_icon_key = ( $icon_key === 'font-icon' ) ? 'icon' : 'font-icon';
 
-                if ( $icon_key == 'font-icon' ){
-                    $custom_field_options[$post_type][$field_key]['default'][$new_field_option_key]['icon'] = '';
-                }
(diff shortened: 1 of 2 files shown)


corsacca:
AI says maybe problem:

 Problem                                                                                                   
                                                                                                    
  In dt-core/admin/menu/tabs/tab-custom-fields.php at line 1039 (inside the restore_field_icon handler), the
   two icon fallbacks are inconsistent:                                                                     
                                                                                                            
  $custom_field['icon']      = $restore_icon_defaults[ $field_key ]['icon'] ?? '';                          
  $custom_field['font-icon'] = $restore_icon_defaults[ $field_key ]['font-icon'] ?? null;                   
                                                                                                            
  When the default field definition uses only a font-icon (e.g. an MDI class) and has no icon key, this     
  stores icon = '' and font-icon = 'mdi mdi-...'.                                                           
                                                                                                            
  The admin UI then renders the icon at line 437:                                            

  if ( isset( $field['icon'] ) ) {
      // render <img src="...">                                                                             
  } elseif ( isset( $field['font-icon'] ) ) {
      // render <i class="...">                                                                             
  }                                                                                                         
   
  isset() returns true for an empty string but false for null. So with icon = '', the first branch wins and 
  the UI emits <img src=""> — a broken image — and the font-icon branch is never reached. Clicking "Restore
  to Default" on a field whose default is a font-icon visually breaks the icon.                             
                                                                                             
  Fix                                                                                                       
   
  Change the icon fallback from ?? '' to ?? null so both keys behave consistently and match the null-based  
  convention the rest of the PR uses:                                                        
                                                                                                            
  $custom_field['icon']      = $restore_icon_defaults[ $field_key ]['icon'] ?? null;         
  $custom_field['font-icon'] = $restore_icon_defaults[ $field_key ]['font-icon'] ?? null;                   
   
  With null, isset( $field['icon'] ) returns false when the default has no image icon, the render falls     
  through to the font-icon branch, and the restored MDI icon displays correctly.

cairocoder01:
Good catch Claude. Fix is pushed.

corsacca:
Thanks @cairocoder01!

PR #2918, opened 2026-05-04
Components v0.9.1

cairocoder01:
- Consistent Add Item UI for multi-text (add moved to top right)
- Proper reset of loading, saved, error indicators


cairocoder01:
Claude is wrong - the disabled state is now fixed. Tested locally.

PR #2920, opened 2026-05-13
Fix `/login` 404 status code

cairocoder01:
When enabling the SSO custom login page, although the page loads correctly, it technically returns a 404 status code. The same is true of the privacy and terms pages. This resolves that by setting a 200 status for those pages.


corsacca:
Thanks @cairocoder01!

PR #2921, opened 2026-05-21
User management fields

### LINKED ISSUE #2847: Replace components: user management
On the user management page (`/user-management/users`), replace old form controls with the appropriate web components.

Replacements:
- Add User
  - [ ] Use dt-text for form controls
- Edit User
  - Status panel
    - [ ] User status: single-select
    - [ ] Workload status: single-select
    - [ ] Languages: multi-select-button-group
  - User Profile panel
    - [ ] dt-text (or appropriate component) for all fields
  - User Apps panel
    - [ ] Use dt-toggle
- Notifications
  - [ ] Use dt-toggle for toggles
 
brady-lamansky-gtt:
@cairocoder01 This resolves #2847 to replace user management fields w dt-web-components.
Please specifically look at the syntax and formatting of my changes to make sure it's what you're looking for!


dt-users/template-new-user.php
```diff
@@ -111,15 +111,48 @@
...
-                                                <dd><input type="email" class="input" id="liame" placeholder="<?php esc_html_e( 'Email', 'disciple_tools' ); ?>" required autocomplete="off" /> </dd>
+                                                <dt></dt>
+                                                <dd>
+                                                <?php DT_Components::render_text( 'nickname', [
+                                                        'nickname' => [
+                                                            'name' => __( 'Display Name', 'disciple_tools' ),
+                                                            'type'        => 'text',
+                                                            'default' => '',
+                                                        ]
+                                                ], [], [ 'placeholder' => 'Nickname (Display Name)' ] ) ?>
```
cairocoder01: This value needs to be localized so it's translated in other languages

dt-users/template-new-user.php
```diff
@@ -111,15 +111,48 @@
...
+                                                <dt></dt>
+                                                <dd>
+                                                <?php DT_Components::render_text( 'email', [
+                                                        'email' => [
+                                                            'name' => __( 'Email', 'disciple_tools' ),
+                                                            'type'        => 'text',
+                                                            'default' => '',
+                                                            'required' => true,
+                                                        ]
+                                                ], [], [ 'placeholder' => 'Email' ] ) ?>
```
cairocoder01: Localize placeholder

dt-users/template-new-user.php
```diff
@@ -111,15 +111,48 @@
...
-                                                    <dd><input type="password" class="input" id="drowssap" placeholder="<?php esc_html_e( 'Password', 'disciple_tools' ); ?>" autocomplete="off" /> </dd>
+                                                    <dt></dt>
+                                                    <dd>
+                                                    <?php DT_Components::render_text( 'username', [
+                                                        'username' => [
+                                                            'name' => __( 'Username', 'disciple_tools' ),
+                                                            'type'        => 'text',
+                                                            'default' => '',
+                                                        ]
+                                                    ], [], [ 'placeholder' => 'Username' ] ) ?>
```
cairocoder01: Localize placeholder

dt-users/template-new-user.php
```diff
@@ -111,15 +111,48 @@
...
+                                                </dd>
+                                                    <dt></dt>
+                                                    <dd>
+                                                    <?php DT_Components::render_text( 'password', [
+                                                        'password' => [
+                                                            'name' => __( 'Password', 'disciple_tools' ),
+                                                            'type'        => 'text',
+                                                            'default' => '',
+                                                        ]
+                                                    ], [], [ 'placeholder' => 'Password' ] ) ?>
```
cairocoder01: Localize placeholder

dt-users/template-new-user.php
```diff
@@ -175,23 +208,36 @@
...
-                                                        name="first_name" placeholder="<?php esc_html_e( 'First Name', 'disciple_tools' )?>" />
+                                                <dt></dt>
+                                                <dd>
+                                                    <?php DT_Components::render_text( 'first_name', [
+                                                        'first_name' => [
+                                                            'name' => __( 'First Name (optional)', 'disciple_tools' ),
+                                                            'type'        => 'text',
+                                                            'default' => '',
+                                                        ]
+                                                    ], [], [ 'placeholder' => 'First Name' ] ) ?>
```
cairocoder01: Localize placeholder

dt-users/template-new-user.php
```diff
@@ -175,23 +208,36 @@
...
-                                                            </option>
-                                                        <?php endforeach; ?>
-                                                    </select>
+                                                    <?php DT_Components::render_text( 'last_name', [
+                                                        'last_name' => [
+                                                            'name' => __( 'Last Name (optional)', 'disciple_tools' ),
+                                                            'type'        => 'text',
+                                                            'default' => '',
+                                                        ]
+                                                    ], [], [ 'placeholder' => 'Last Name' ] ) ?>
```
cairocoder01: Localize placeholder

dt-users/user-management.js
```diff
@@ -539,8 +539,8 @@ jQuery(document).ready(function ($) {
                 <a id="app_link_${app_key}" ${activated ? '' : 'disabled'} class="button" href="${url}" target="_blank" style="margin-right: 5px;">${window.SHAREDFUNCTIONS.escapeHTML(magic_link['label'])} <img class="dt-icon dt-white-icon" src="${window.SHAREDFUNCTIONS.escapeHTML(window.dt_user_management_localized?.theme_uri)}dt-assets/images/open-link.svg"/></a>
             </td>
             <td style="vertical-align: top;">
-                <input id="app_state_${app_key}"
-                        class="switch-input app-state-switches" type="checkbox"
+                <dt-toggle id="app_state_${app_key}"
+                        class="app-state-switches" type="checkbox"
```
cairocoder01: When I toggle one of these, I get a js error that the request failed. But it seems to trigger the request twice and it was the second of the two that failed. Specifically, it's when I uncheck one of these toggles that it happens.

dt-users/template-user-management.php
```diff
@@ -225,16 +229,25 @@ class="dt_multi_select empty-select-button select-button button ">
                                         <p>
                                             <?php esc_html_e( 'Display Name', 'disciple_tools' ); ?>
                                             <span id="<?php echo esc_html( 'update_display_name' ); ?>-spinner" class="loading-spinner"></span>
```
cairocoder01: We should replace these labels by setting the label attribute of the components. And the loading spinner should use the loading icon/attribute of the component

dt-users/template-new-user.php
```diff
@@ -111,15 +111,48 @@
                                             </dl>
                                             <div id="contact-result"></div>
                                             <dl>
-                                                <dt><label for="eman"><?php esc_html_e( 'Display Name', 'disciple_tools' ); ?></label></dt>
-                                                <dd><input type="text" class="input" id="eman" placeholder="<?php esc_html_e( 'Nickname (Display Name)', 'disciple_tools' ); ?>" required autocomplete="off" /> </dd>
-                                                <dt><label for="liame"><?php esc_html_e( 'Email', 'disciple_tools' ); ?></label></dt>
-                                                <dd><input type="email" class="input" id="liame" placeholder="<?php esc_html_e( 'Email', 'disciple_tools' ); ?>" required autocomplete="off" /> </dd>
+                                                <dt></dt>
+                                                <dd>
```
cairocoder01: We can get rid of this dl/dt/dd structure. A dl is a definition list for definition terms (dt) and descriptions (dd). It works well for labels and values, but those are contained within the components now. We can just put the components side-by-side (except maybe the hidden ones, which can stay in their own div).

cairocoder01 (changes requested):
This is a great start. See the specific code comments, but there are also some functional issues:

- [x] When using the new user form, the submit doesn't work because it returns an API error
- [x] When changing fields of an existing user, they don't automatically save like they used to. We need to hook up the code to capture change events and trigger the API requests
- [ ] Those will probably require js changes. Try to find the js code that was specific to the old form fields and remove it as you implement the new code for the components.

dt-users/template-new-user.php
```diff
@@ -134,8 +154,7 @@
                                                 <dt><label for="new-user-language-dropdown"><?php esc_html_e( 'Language', 'disciple_tools' ); ?></label></dt>
                                                 <dd id="new-user-language-dropdown"></dd>
                                                 <?php if ( $current_user_can_manage_users ): ?>
-                                                <dt><?php esc_html_e( 'Role', 'disciple_tools' ); ?></dt>
-                                                <dd>
+                                                <?php esc_html_e( 'Role', 'disciple_tools' ); ?>
                                                     <?php
```
brady-lamansky-gtt: Do these fields need to be changed as well? It wasn't as straightforward as changing their field type. Any suggestions?
Also, removing the <dd> on line 113 causes these two fields (Language and Role) to disappear, though there's no closing </dd> tag anywhere.

dt-users/user-management.js
```diff
@@ -567,7 +567,7 @@ jQuery(document).ready(function ($) {
     $('#magic_link_apps_spinner').hide();
   }
 
-  $(document).on('click', '.app-state-switches', function (e) {
+  $(document).on('change', '.app-state-switches', function (e) {
```
brady-lamansky-gtt, line 572: 'click' was causing the toggle to trigger twice due to the custom component being used (I can't remember the specific reason, but it was being picked up twice). Using 'change' fixed this

dt-users/user-management.js
```diff
@@ -765,27 +765,31 @@ jQuery(document).ready(function ($) {
     );
   };
 
-  $('textarea.text-input, input.text-input').change(function () {
+  $('dt-textarea, dt-text').change(function () {
```
brady-lamansky-gtt, line 770: I changed the tracked inputs for the change() function. Is this the correct way to do it? This allows the fields to autosave properly when editing a user.
However, it was also being run when creating a new user, which caused an error when calling update_user, so I added a check. Is this the correct way to handle this?

dt-users/template-new-user.php
cairocoder01: Let's try to update the Language field to the single select component. For Roles, we don't have a checkbox list component, so we can mostly leave it as-is. However, maybe we can use the dt-label component for the label to make sure that style stays consistent.

As far as those fields disappearing, they shouldn't be included in the See/Hide More Options UI. They should both always be visible. There might have been some invalid HTML because of how the dl/dt/dd were being nested with divs. But I also see there was a closing div removed after the password field. I think that's the cause of the Language and Roles fields being hidden. I think that `</div>` needs to be added back.

dt-users/user-management.js
cairocoder01: So this file is loaded on both the new user and update user pages, right? Since the loading spinner never needs to show up on the new user page, we could wrap this whole change event handler in your conditional of `window.current_user_lookup`; the new user page doesn't need to even handle the change event.

You'll need to do the same thing for the single-select code after this since the loading spinner is showing on the new user page.

And instead of handling each component type separately, you could copy this code from the component service and handle them all together. You can just supply your own change event handler.

```js
enableAutoSave(selector) {
    const allElements = document.querySelectorAll(
      selector || this.autoSaveComponents.join(','),
    );
    if (allElements) {
      allElements.forEach(el => {
        el.addEventListener('change', this.handleChangeEvent.bind(this));
      });
    }
  }
```

cairocoder01: Also, if we're going to show the loading icons for the update user page, we should also show the saved icon when it succeeds and catch errors to show the errors.

cairocoder01:
- [ ] We've lost the spacing in between the fields and the following label. See image below. Green is good, red is bad and needs to match green.
image
image

dt-users/user-management.js
```diff
@@ -1083,12 +1087,12 @@ jQuery(document).ready(function ($) {
...
 
     let create_user = (corresponds_to_contact, archive_comments = false) => {
-      let name = $('#eman').val();
-      let email = $('#liame').val();
+      let name = $('#nickname').val();
+      let email = $('#email').val();
       let locale = $('#locale').val();
 
-      const username = $('#emanresu').val();
-      const password = $('#drowssap').val();
```
cairocoder01, line 1091: I know these probably looked like typos, but I believe they were intentional. They are the field names in reverse so that auto-complete in your browser doesn't try to fill it with your own saved user/email info. Let's change these names back to their reversed versions.

dt-users/user-management.js
brady-lamansky-gtt: Sounds good! When wrapping my code in `if (window.current_user_lookup) {...}`, the loading no longer works. It looks like, when the page loads, window.current_user_lookup doesn't yet have a value. Is putting the check right inside the function alright, or do you have another suggestion?

cairocoder01: Ah, ok. That's right, it won't be there at load time because the individual user is loaded into a modal after the initial page load. You can put the check inside like you suggested. The other option is to look for the method that opens the edit user modal and initialize the event there. The thing you will have to watch for is registering the event multiple times if you click to open multiple users in one session. You can either check if an event listener is already attached or remove the previous event listener before you add a new one.

brady-lamansky-gtt: I updated the PR with the check for now! I'll look into what the other option would look like later as well. We're just about done with the project for the Japanese internship, so I'll have more capacity after it ends this coming Wednesday. Thanks for all the help

dt-users/template-new-user.php
```diff
@@ -90,57 +90,61 @@
...
+                                                <div class="subassigned details" style="margin-bottom: 16px;">
+                                                    <var id="subassigned-result-container" class="result-container subassigned-result-container"></var>
+                                                    <div id="subassigned_t" name="form-subassigned" class="scrollable-typeahead">
+                                                        <div class="typeahead__container">
+                                                            <div class="typeahead__field">
+                                                                <span class="typeahead__query">
+                                                                    <input id="subassigned[query]" class="js-typeahead-subassigned input-height"
+                                                                            name="subassigned[query]" placeholder="<?php esc_html_e( 'Search multipliers and contacts', 'disciple_tools' ); ?>"
+                                                                            autocomplete="off">
+                                                                </span>
```
cairocoder01: I just noticed that this is an old typeahead. Can we update it to a user connection field with the ability to select only a single value?

dt-users/template-new-user.php
```diff
@@ -90,57 +90,61 @@
...
                                                         </div>
                                                     </div>
-                                                </dd>
-                                            </dl>
+                                                </div>
                                             <div id="contact-result"></div>
-                                                <dd>
-                                                <?php DT_Components::render_text( 'nickname', [
-                                                        'nickname' => [
+                                                <div style="margin-bottom: 16px;">
```
cairocoder01: Instead of putting inline styles for the margin, can we just add a common class to all of the field containers (something like `user-field` or similar), and then we can update the CSS for the page to set the same margin for all of them.

cairocoder01: Also, let's use `1rem` for the margin. It comes out to roughly the same pixels since the base font size is 15px, but it means the margin will automatically adjust if someone increases or decreases the browser font size.

dt-users/template-new-user.php
```diff
@@ -191,54 +195,62 @@
...
+                                            ?>
+
+                                            <dt-textarea
+                                                class="input" id="description"
+                                                label="<?php esc_html_e( 'Biography', 'disciple_tools' )?>"
+                                                name="description"
+                                                placeholder="<?php esc_html_e( 'Biography', 'disciple_tools' )?>"
+                                                rows="5"
+                                                data-optional
+                                            ></textarea>
```
cairocoder01: This closing tag doesn't match the opening one: `<dt-textarea></textarea>`

cairocoder01 (changes requested):
Looking better. 

- [x] We still need the margin adjustments on the user management page for editing existing users.
- [x] I found a weird issue when editing users. I set the gender of a user to Female (had been blank), and then when I opened up any other user, it also showed Female even though they should have been blank. I'm having trouble reproducing it again after setting those values.
- [x] Related to that, as I was trying to reproduce it, I also had a situation where I changed another user to Male from blank, it saved, I closed to modal and then opened the same user, but the gender was blank again. If I refreshed the page, it correctly showed Male. I think whatever js is loading the modal had the original value instead of the updated one.

brady-lamansky-gtt:
Just fixed those bugs! Let me know any other suggestions. For the Gender bug, I realize it was because I had `'select_cannot_be_empty' => true` even though the field was supposed to be optional.

dt-users/template-new-user.php
```diff
@@ -326,4 +320,10 @@ class="input" id="description"
 
 </div> <!-- end #content -->
 
+<style>
+    .user-field {
+        margin-bottom: 1rem;
+    }
+</style>
+
```
cairocoder01: This should be in the main CSS file in `/dt-assets/scss/_user_management.scss`. You'll need to `npm run build` to build the CSS and see that change.

dt-users/template-new-user.php
```diff
@@ -90,23 +90,18 @@
...
-                                                    </div>
+                                                <div class="user-field">
+                                                <?php DT_Components::render_connection( 'subassigned[query]', [
+                                                        'subassigned[query]' => [
+                                                            'name' => __( 'Contact to make a user (optional)', 'disciple_tools' ),
+                                                            'default' => '',
+                                                            'type' => 'connection',
+                                                            'post_type' => 'contacts',
+                                                        ]
+                                                ], [], [ 'placeholder' => __( 'Search multipliers and contacts', 'disciple_tools' ) ] ) ?>
```
cairocoder01: This is a much better code style. However, if I select a contact in this field and save the new user, it still creates a new contact instead of using this value.

Also, we should add `'allowAdd' => false` to the params so that users can't add a new contact in the field.

dt-users/template-user-management.php
```diff
@@ -477,4 +487,10 @@ class="dt_multi_select empty-select-button select-button button ">
 
 </div> <!-- end #content -->
 
+<style>
+    .user-field {
+        margin-bottom: 1rem;
+    }
+</style>
+
```
cairocoder01: This should be in the main SCSS files

dt-users/user-management.js
```diff
@@ -1208,6 +1215,7 @@ jQuery(document).ready(function ($) {
         });
     }
 
+    // subassigned check here
     ['subassigned'].forEach((field_id) => {
       $.typeahead({
         input: `.js-typeahead-${field_id}`,
```
cairocoder01: This code won't be needed since we removed the typeahead for this.

cairocoder01 (changes requested):
I pushed one commit to reset all of the component fields before loading in new values. It was causing an issue where saved icons were still displaying on a second user modal and the gender field wasn't cleared out if the value was "none" and the previous had a value.

As I was looking in the subassigned code, I see there's a feature to check the URL for a query param and preset the subassigned field (`/user-management/add-user?contact_id=691`). Let's make sure that still works. It happens when you're on a contact, open the Admin Actions dropdown, and select the option to make that contact into a user.

dt-users/template-new-user.php
brady-lamansky-gtt: Ah I see thanks, i had it there originally but didn't know I needed `npm run build`. Thanks!

dt-core/utilities/dt-components.php
```diff
@@ -64,9 +64,14 @@ public static function shared_attributes( $field_key, $fields, $post, $params =
...
         $label_attr = $hide_label ? '' : 'label="' . esc_attr( $fields[$field_key]['name'] ) . '"';
 
+        $data_optional = isset( $params['data-optional'] ) ? 'data-optional' : '';
+
         $shared_attributes = '
               id="' . esc_attr( $display_field_id ) . '"
               name="' . esc_attr( $field_key ) . '"
+              placeholder="' . esc_attr( $params['placeholder'] ?? '' ) . '"
+              data_type="' . esc_attr( $params['data_type'] ?? '' ) . '"
+              ' . $data_optional . '
```
brady-lamansky-gtt, line 75: I originally set this up for type="password" for dt-text, but it was causing formatting errors from some native styling. So, I changed the "type" to "data_type" in my local dt-web-components file in the meantime. How do you recommend we handle this?

cairocoder01: @brady-lamansky-gtt  Can you explain more about the formatting errors it was causing?

brady-lamansky-gtt: @cairocoder01 In code, I had  the following:
``<?php DT_Components::render_text( 'ddrowssap', [
                                                        'ddrowssap' => [
                                                            'name' => __( 'Password', 'disciple_tools' ),
                                                            'type'        => 'text',
                                                            'default' => '',
                                                        ]
                                                    ], [], [ 'type' => 'password', 'placeholder' => __( 'Password', 'disciple_tools' ) ] ) ?>``

So the component was still a "text" type, but when rendering I set "type = password" using custom attributes. Below is what it looked like. It looked like this, but the text would still show as dots as if it were a password. When talking to Gemini, it said the default "type = password" styling was conflicting with our custom "type = password" functionality on the component.

Screenshot 2026-06-22 at 9 56 23 AM

cairocoder01: I took a look and the problem was some styles from the Foundation library that's in use in the project. I added some CSS to reset those styles for `dt-text` elements. This fixes the styling issue and allows us to still use the `type` attribute that the component is actually looking for.

cairocoder01:
@corsacca This looks good to me. I've reviewed all the Claude reviews and fixed what seemed important but have left a number of the medium issues. Ready for your review.

corsacca:
Thanks guys!

PR #2925, opened 2026-06-26
chore: update @disciple.tools/web-components to version 1.0.1

cairocoder01:
v1.0.1 — Patch Release

Bug Fixes

- dt-location-map: Pin now tracks the map center as you pan instead of
  requiring a click to place. (dt-map-modal.js)
- dt-location-map: canEdit now checks for metadata.lat, so a location
  with a label but no coordinates opens in editable state.
- dt-location-map: Closing a read-only map no longer emits a submit event.
- Google Geocoder: LatLng.lat()/lng() were called as properties, returning
  undefined on every geocoded result. Fixed. (componentService.js)
- Google Geocoder: reverseGeocode() now uses the Maps JS SDK Geocoder
  instead of a direct REST fetch. getPlaceDetails() returns early when
  called without a place_id. (googleGeocodeService.js)

Improvements

- dt-connection / dt-tags: Dropdown results are cached after the first
  fetch and filtered locally on reopen. Cache clears on query change.
- dt-location-map: Support for selecting location on map and reverse geocoding

Developer Experience

- dt-location-map: dt:geocode event is now observable in Storybook Actions.


corsacca:
THanks @cairocoder01!

PR #2927, opened 2026-07-02
Record list fields

### LINKED ISSUE #2845: Replace components: record list
On the contact/groups/etc list page, replace old form controls with the appropriate web components.

Replacements:
- Custom Filter Creation
  - [ ] Date fields (e.g. Baptism date)
  - [ ] Connection fields (e.g. Subassigned)
  - [ ] Multi-select fields (e.g. Sources)
  - [ ] Tags fields (e.g. Tags)
  - [ ] User-select fields (e.g. Assigned To)
  - [ ] Location grid fields
    - Both standard location_grid as well as location_grid_meta. 
    - Both currently search by grid_id, but when geocoding is enabled, the locations saved on the records (geocoded values) don't match the names of the grids in the search
    - When geocoding is enabled, use the location-map component, but still use the grid_id for the actually value to search by
- [ ] Field selection for list
  - Currently uses a custom typeahead. Convert to multi-select
- [ ] Bulk Edit
  - Ensure all field types use the appropriate component
- [ ] Show Archived toggle - use dt-toggle

brady-lamansky-gtt:
This resolves #2845 to replace record-list fields with dt-web-components. Filtering functionality was updated to accommodate the new changes & appeared to work properly, but more testing is likely needed.

I updated these things:
- Custom Filter Creation
- Bulk Edit
- Field Selection (for the displayed fields)

@cairocoder01 let me know any suggestions!

Related Components PR: https://github.com/DiscipleTools/disciple-tools-web-components/pull/208


cairocoder01:
After a brief review manually and bigger Claude review, here's some preliminary things to fix:

The issue may have been misleading about replacing all components in the filters, since some of the filter controls don't match the component normally used for that field.

# Filters
## Regressions
- [x] Date fields no longer support range selection
- [x] `key_select` fields no longer support selecting multiple values to search for
- [x] `boolean` fields no longer support selecting multiple values to search for
- [x] Filter chips display raw keys instead of labels (e.g. `Contact Status: active` instead of expected `Contact Status: Active`

## Bugs
- [x] Connection fields show the button to add a new connection. No need to add from the filter selection

cairocoder01:
Here is a full Claude review I did locally while it had access to a dev site running the code:

# Code Review: Record List Fields → Web Components (`record-list-fields`)

Reviewed the diff against `origin/develop` (source files: `archive-template.php`,
`dt-assets/js/modular-list.js`, `dt-assets/js/modular-list-bulk.js`,
`dt-core/utilities/dt-components.php`, `dt-assets/scss/_components.scss`) and
verified the runtime behavior live on `dtagent.local` (deployed via wp-deploy)
with network capture and the PHP debug log.

---

## 🔴 High

### 1. Date filters are silently broken (return unfiltered results)
`archive-template.php:994` renders a single `dt-date`, and in
`modular-list.js:1453-1469` date/datetime fields fall through into the generic
"single select" branch, producing `val = [customComponent.value]`.

The list query builder in `dt-posts/posts.php:874-884` only applies a date
constraint when the value is an object keyed by `start`/`end`. An array key (`0`)
matches neither, so **no WHERE clause is emitted — the filter is silently ignored
and all records are returned.**

Verified live — applying a Baptism Date filter posts:

```json
{"fields":[{"baptism_date":["2024-01-15"]}, ...]}
// expected: {"baptism_date":{"start":"2024-01-15"}}
```

This is also a capability regression: the old UI had **Range Start / Range End**;
the new one renders one date input, so date-range filtering is gone entirely.

**Fix:** add a dedicated date branch that emits `{ [field]: { start, end } }` and a
start/end UI (two `dt-date`s).

---

## 🟠 Medium

### 2. `render_toggle` throws PHP warnings on every list load
`dt-components.php:325` does `checked( $post[$field_key], '1', false )`, but the
new call sites pass an empty `$post` (`[]`) — `archive-template.php:466` (archived
toggle) and `archive-template.php:1005` (every boolean filter field). Confirmed in
`debug.log`:

```
PHP Warning: Undefined array key "archivedToggle" in .../dt-components.php on line 325
PHP Warning: Undefined array key "favorite" ...
PHP Warning: Undefined array key "requires_update" ...
```

**Fix:** `checked( $post[$field_key] ?? '', '1', false )` in `render_toggle`.

### 3. `key_select` and `boolean` filters regressed to single-value
`archive-template.php:964` renders `key_select` as `dt-single-select` and
`boolean` as `dt-toggle`. Previously these were multi-checkbox groups.
Consequences:

- **key_select** (e.g. Contact Status): can no longer filter for multiple statuses
  at once, and the explicit **"None Set"** option is gone. The backend still
  supports value arrays (`dt-posts/posts.php:886`) — only the UI regressed.
- **boolean**: the old Yes/No checkboxes are replaced by one toggle, so filtering
  for "No"/unset (`0`) is no longer expressible.

### 4. Field-selection save persists stale `-` delete markers
`modular-list.js:890-898` reads `$('#field_search_input').val()` directly. When a
user removes a column, `dt-multi-select` marks it as `-fieldkey` rather than
removing it. Verified live — after removing "Name", the saved value is:

```json
["-name","last_modified","favorite","overall_status","assigned_to","milestones","groups","seeker_path"]
```

That `-name` gets written to the `fields_to_show_in_table` cookie and sent as
`fields_to_return`. The filter-change handler already strips these
(`modular-list.js:1449-1452`); `save_column_choices` should do the same
(`.filter(k => k.charAt(0) !== '-')`). The now-dead `try/catch` around it (no JSON
parsing anymore) should also be removed.

### 5. Filter chips display raw keys instead of labels
For key_select/multi_select/tags, `create_name_value_label`
(`modular-list.js:1735`) interpolates the value directly. Verified live the chips
read **"Contact Status: active"** and **"Sources: web,phone"** instead of the
translated option labels ("Active", "Web, Phone"). Connections/users/locations are
fine (they use `item.label`). The change handler should resolve option keys to
labels before building the chip text.

---

## 🟡 Low / cleanup
(the issue explicitly asks to remove now-unused code)

- **Dead field-picker code** left behind after the template rewrite: the
  `.remove-field-btn` handler and `#selected_fields_input` /
  `#enabled_fields_container` references (`modular-list.js:866-888`) and in
  `reset_column_choices` (`modular-list.js:908`) — none of those elements exist
  anymore.
- **Dead `.text-comms-filter-input` keyup handler** (`modular-list.js:1774`) — that
  class is no longer rendered (fields are now `dt-text`), so the old "auto-select
  the default radio while typing" behavior is silently gone. Text filtering still
  works, but confirm this UX loss is intended.
- **Dead `.clear-date-picker` handler** (`modular-list.js:2323`) — those buttons
  were removed from the date template.
- **Duplicate/mismatched selector** in the modal-close reset
  (`modular-list.js:2115`): `dt-location` is listed twice, and `dt-location-map` is
  in the reset branch's tag list but not in the query selector.
- `dt-assets/scss/_components.scss` has no trailing newline (line 17) —
  PHPCS/prettier nitpick.

---

## What I verified vs. inferred
Live-confirmed on `dtagent.local`: date query shape (#1), PHP warnings (#2), stale
`-` markers (#4), raw-key chips (#5), single-select rendering (#3). The
double-labeling I initially suspected for text/comms fields turned out **not** to
be a bug — the radio handler correctly replaces the change-handler chip (net one
chip). Connection/user/location/multi-select/tags query building looks correct.

archive-template.php
```diff
@@ -1123,12 +1083,51 @@ class="filter-by-file-upload-option"
...
+                                if ( $field_options[$field]['type'] === 'connection' ) : ?>
+                                    <p>
+                                        <label><?php echo esc_html( sprintf( _x( 'All %1$s with %2$s', 'All Contacts with Is Coaching', 'disciple_tools' ), $post_settings['label_plural'], $field_options[$field]['name'] ) ) ?>
+                                            <input class="all-connections" type="checkbox" value="all-connections" />
+                                        </label>
+                                        <label><?php echo esc_html( sprintf( _x( 'All %1$s without %2$s', 'All Contacts without Is Coaching', 'disciple_tools' ), $post_settings['label_plural'], $field_options[$field]['name'] ) ) ?>
+                                            <input class="all-without-connections" type="checkbox" value="all-without-connections" />
+                                        </label>
+                                    </p>
+                                <?php endif;?>
```
cairocoder01: Since the switch above is on type, we can just move this up into the connection case, right?

archive-template.php
```diff
@@ -1123,12 +1083,51 @@ class="filter-by-file-upload-option"
...
+                                            <?php echo esc_html( sprintf( _x( 'All %1$s with any value in %2$s', 'All Contacts with any value in field', 'disciple_tools' ), $post_settings['label_plural'], $field_options[$field]['name'] ) ) ?>
+                                        </label>
+                                        <label>
+                                            <input name="filter_by_text_comms_option_<?php echo esc_attr( $field ); ?>" class="filter-by-text-comms-option" type="radio" value="all-without-set-value" data-field="<?php echo esc_html( $field ) ?>" />
+                                            <?php echo esc_html( sprintf( _x( 'All %1$s with no value in %2$s', 'All Contacts with no value in field', 'disciple_tools' ), $post_settings['label_plural'], $field_options[$field]['name'] ) ) ?>
+                                        </label>
+                                    </p>
+                                <?php endif;
```
cairocoder01: This can be moved up into the appropriate case statement, right?

archive-template.php
```diff
@@ -1123,12 +1083,51 @@ class="filter-by-file-upload-option"
...
+                                        </label>
+                                    </p>
+                                <?php endif;?>
+                                <?php if ( $field === 'subassigned' ): ?>
+                                    <p>
+                                        <label><?php esc_html_e( 'Filter for subassigned OR Assigned To', 'disciple_tools' ) ?>
+                                            <input id="combine_subassigned" type="checkbox" value="combine_subassigned" />
+                                        </label>
+                                    </p>
+                                <?php endif;?>
```
cairocoder01: We know subassigned has type connection, so we can move this up into the case statement above.

dt-assets/js/modular-list.js
```diff
@@ -2157,9 +1878,16 @@
...
+            `dt-multi-text#${field}, dt-text#${field}`,
+          ).value;
+          if (Array.isArray(filtered_value)) {
+            filtered_value = filtered_value
+              .filter((label) => label.value.length > 0)
+              .map((item) => item.value);
+          }
           newLabel['name'] =
-            filterName = `${esc(list_settings.post_type_settings.fields[field] ? list_settings.post_type_settings.fields[field].name : '')}: ${esc(filtered_value)}`;
+            filterName = `${esc(list_settings.post_type_settings.fields[field] ? list_settings.post_type_settings.fields[field].name : '')}: ${filtered_value}`;
```
cairocoder01: Why did we remove the `esc(...)` wrapper? That helps prevent HTML injection. For example, put a filter value of `<button>click!</button>` in a text/comm-channel field. It will show correctly on the record list, but in the filter modal header it will actually render a button element that just says click. We need that escape function in there.

dt-assets/js/modular-list.js
```diff
@@ -1575,129 +1393,77 @@
...
+              [field]: adjust_search_query_filter_states(field, type, val),
+            });
           }
-          case 'all-with-filtered-value': {
-            value = filter;
-            break;
+        } else if (type === 'text' || type === 'communication_channel') {
+          val = customComponent.value;
+
+          switch ($('.filter-by-text-comms-option:checked').val()) {
```
cairocoder01: This looks like an existing bug, but let's quick address it. This jquery selector will select the first checked option across ALL fields. So if you're setting filters on both email and phone, it will use the selection for the first one it finds. We need to scope this selector to the field container. Each tab-panel has an id of the field_id. We can follow the pattern of the connections field logic just before this that uses: ```$(`#${field} .all-connections`)``` to do something more like ```$(`#${field} .filter-by-text-comms-option:checked`)```. This makes sure we're selecting the options within the field we're actually processing at this point.

dt-assets/js/modular-list.js
```diff
@@ -861,221 +883,17 @@
...
-
   $('#save_column_choices').on('click', function () {
     let selectedFields;
     try {
-      selectedFields = JSON.parse($('#selected_fields_input').val() || '[]');
+      selectedFields = $('#field_search_input').val() || '[]';
     } catch (e) {
       console.error('Error parsing selected fields JSON:', e);
       selectedFields = [];
     }
```
cairocoder01: We're no longer trying to parse a JSON string, so I think we can remove this try/catch.

dt-assets/js/modular-list.js
```diff
@@ -1575,129 +1393,77 @@
...
-        switch ($('.filter-by-text-comms-option:checked').val()) {
-          case 'all-with-set-value': {
-            value = '*';
-            break;
-          }
-          case 'all-without-set-value': {
-            value = null;
-            break;
+      let customComponent = document.querySelector(
+        `dt-toggle[name="${field}"], dt-tags[name="${field}"], dt-date[name="${field}"], dt-location[name="${field}"], dt-users-connection[name="${field}"], dt-connection[name="${field}"], dt-single-select[name="${field}"], dt-multi-select[name="${field}"], dt-multi-select-button-group[name="${field}"], dt-multi-text[name="${field}"], dt-text[name="${field}"]`,
```
cairocoder01: If you want to select any component with that name, you can just use `[name="${field}]` without the actual element name and it will select any element with a name attribute that matches.

dt-assets/js/modular-list.js
```diff
@@ -1575,129 +1393,77 @@
...
-          case 'communication_channel': {
-            search_query.push({ [field]: value !== null ? [value] : [] });
-            break;
+          if (val && val.length > 0) {
+            // replace any '<' or '>' characters to save properly
+            val = val.map((v) =>
+              typeof v === 'string'
+                ? v.replace(/</g, '＜').replace(/>/g, '＞')
+                : v,
+            );
```
cairocoder01: According to Claude, this is only needed if the tags field has allowAdd. Since we're removing that, I think we can remove this, right? Is there another reason it's here?

dt-assets/js/modular-list.js
```diff
@@ -1971,31 +1747,26 @@
       `post_type_settings.fields.${field}.name`,
       field,
     );
+    if (
+      typeof value === 'string' &&
+      (value.includes('<') || value.includes('>'))
+    ) {
+      value = value.replace(/</g, '＜').replace(/>/g, '＞');
```
cairocoder01: According to Claude, this is only needed if the tags field has allowAdd. Since we're removing that, I think we can remove this, right? Is there another reason it's here?

dt-assets/js/modular-list.js
```diff
@@ -2609,15 +2098,44 @@
...
+  $('#filter-modal').on('closed.zf.reveal', function () {
+    $(this)
+      .find(
+        'dt-location, dt-toggle, dt-tags, dt-date, dt-users-connection, dt-connection, dt-single-select, dt-multi-select, dt-multi-select-button-group, dt-multi-text, dt-text',
+      )
+      .each(function () {
+        const tagName = this.tagName.toLowerCase();
+
+        if (tagName === 'dt-date') {
+          this.reset();
```
cairocoder01: Why can't we use reset for all components? That is supposed to work for every component.

cairocoder01:
## Regressions
- [x] key_select fields no longer support selecting "none set" in order to filter by records with no value (screenshot from current theme version before these changes)

| old | new |
| --- | --- |
| image | image |

- [x] Connection and users-connections filter previously had a placeholder "Type to search" that we can add back in. Also, we are now showing the field label above that is duplicated and not needed

| old | new |
| --- | --- |
| image | image |

- [x] Connections filter previously disabled the input field when the all or none filter was selected. We should re-implement that. That happened on 1691 with the line `typeahead.prop('disabled', true);` which no longer does anything since there is no typeahead.
- [x] Location or Address shows not unsupported error. Previously showed the standard location grid filter.

| old | new |
| --- | --- |
| image | image |

- [x] multi_select previously displayed a list selection instead of a button group. We should match the display type of the field, whether it should show as a standard list select or a button group.

| old | new |
| --- | --- |
| image | image |

- [ ] Tags filter is showing the option to add a new tag:
image

- [x] Communication channel and text are missing placeholders
- [x] If I create a date range filter and save it as a custom filter, when I click the edit icon next to the saved custom filter, the filter modal doesn't have the values of the saved filter. The same seems to be true of other saved filters. Claude eventually found it and gave this detail:
> [modular-list.js:2152 edit_saved_filter](vscode-webview://1vofadbn3km439os5l1gdjgim37vcncdfaemjrv1sonmrko8sdeh/dt-assets/js/modular-list.js#L2152) still runs on the old typeahead machinery:
>
> - Line 2081 declares let typeaheads_loaded = null; and the code that used to assign it (typeaheads_loaded = load_multi_select_typeaheads()…) was deleted in this PR. It's never reassigned, so [line 2154](vscode-webview://1vofadbn3km439os5l1gdjgim37vcncdfaemjrv1sonmrko8sdeh/dt-assets/js/modular-list.js#L2154) typeaheads_loaded.then(…) throws TypeError: Cannot read properties of null (reading 'then') the instant you click the edit (pencil) icon on a saved custom filter. The modal opens blank and nothing populates.
> - Even if you guard that, the callback body still calls removed APIs: .datepicker('setDate', …) at [2184](vscode-webview://1vofadbn3km439os5l1gdjgim37vcncdfaemjrv1sonmrko8sdeh/dt-assets/js/modular-list.js#L2184) and window.Typeahead['.js-typeahead-…'].addMultiselectItemLayout(…) at [2198–2213](vscode-webview://1vofadbn3km439os5l1gdjgim37vcncdfaemjrv1sonmrko8sdeh/dt-assets/js/modular-list.js#L2198) — the typeaheads are never created now, so those are undefined.addMultiselectItemLayout() → another throw.
> 
> So edit_saved_filter needs to be reworked to populate the dt-* components (set .value), not just cleaned up. This is in-scope for "Custom Filter Creation" and is a genuine regression — worth adding to the review as High.

cairocoder01 (changes requested):
Looking pretty good. A few things to clean up yet.

cairocoder01:
Some additional dead code that can be removed after you confirm it is for the old UI and not needed anymore:

## Dead code, safe to remove
- [906–909](/dt-assets/js/modular-list.js#L906) — #reset_column_choices sets #selected_fields_input and #enabled_fields_container; both were removed from the template (picker is now dt-multi-select). Reset still works via the cookie clear + reload below it.
- [2081](/dt-assets/js/modular-list.js#L2081) — let typeaheads_loaded = null; (only referenced by the broken line above; drop it as part of fixing edit_saved_filter).
- [2085–2087](/dt-assets/js/modular-list.js#L2085) — $('#filter-modal input.dt_date_picker').each(...); no .dt_date_picker inputs exist anymore (dates are dt-date, reset in the closed.zf.reveal handler).
- [2094–2096](/dt-assets/js/modular-list.js#L2094) — .typeahead__query.disabled removeClass loop; no such elements remain.
- [2098–2101](/dt-assets/js/modular-list.js#L2098) — $('.typeahead__query input').each(function () { let typeahead = …; }); assigns a local and does nothing with it, and matches nothing.
- [2246–2247](/dt-assets/js/modular-list.js#L2246) — window.Typeahead['.js-typeahead-${field}'].adjustInputSize() in the #filter-tabs change handler; guarded so it won't throw, but it's dead.

dt-assets/js/modular-list.js
brady-lamansky-gtt: I see! I set up the other ones before, just trying to manually set the value to be empty. I handled date separately at the end, since it had both a "Start" and "End", I couldn't just reset one single value. I then found the .reset() function and didn't check if the others had it. Thanks!

dt-assets/js/modular-list.js
brady-lamansky-gtt: I initially was mistakenly rendering the single-select labels to show the value itself: "Under 18 years old" would display: "<18" instead, so I had to handle the special characters in order to render them properly. When I changed the labels to show the name correctly, I didn't realize that was no longer needed. Thanks!

dt-assets/js/modular-list.js
brady-lamansky-gtt: Same as the above comment. I initially was handled special characters "<", but that was when I was rendering the value in the label rather than the name.

dt-assets/js/modular-list.js
brady-lamansky-gtt: Sounds good! Those options are rendered outside of the field itself, so I added them to a div of id="filter_by_text_comms_option_<?php echo esc_attr( $field ); ?>" which I believe should resolve it! Thanks for noticing this, too. The switch now checks for an option under the element with the id from the new div. Let me know if you have a better suggestion.

brady-lamansky-gtt:
@cairocoder01 I just updated this PR and it's ready for review again. Thanks!

cairocoder01:
Looks like everything from the previous reviews has been addressed except for tags fields still showing the add new option.

## New Issues
### High

- Saved filters for `text`, `communication_channel`, `textarea`, or `number` aren't populating when trying to edit:
  -  [dt-assets/js/modular-list.js:2108] `edit_saved_filter` was reworked to fix the `typeaheads_loaded` crash (confirmed — that regression is gone), but its per-label restore loop only handles `key_select`/`boolean`/`file_upload`, `date`/`datetime`, connection-type fields, `multi_select`/`tags`, and `user_select`. There's no branch for `text`, `communication_channel`, `textarea`, or `number` (see the corresponding filter cases at [archive-template.php:1072-1107]). Opening the edit (pencil) icon on a saved filter that includes one of those field types shows the correct chip, but never sets the underlying `dt-text`/`dt-textarea`/`dt-number` component's `.value`. Since `get_custom_filter_search_query()` ([modular-list.js:1368]) reads directly from the live DOM component (not from `filter.labels`), clicking "Save Filter Edits" ([modular-list.js:2202]) — even just to rename the filter — silently pushes an empty value for that field and permanently overwrites the saved filter's real criterion. Fix: add a branch in `edit_saved_filter` that sets `.value` on the matching `dt-text`/`dt-textarea`/`dt-number` component for these types (and the `filter_by_text_comms_option_*` radio state, mirroring the file_upload/key_select handling already there).

brady-lamansky-gtt:
@cairocoder01 I just made those changes! I also fixed an issue where Input fields weren't being disabled properly when selected "All With" or "All Without" checkboxes on text fields.

cairocoder01:
Adding tags issue is resolved. 

- [x] However, I'm still unable to edit a text or comm-channel filter. For comm-channel, it seems to set the radio buttons but not the text value.
- [x] Also, the Name field shows a required field indicator because it is marked as required in the backend. We need to remove that `required` attribute for the filter field

cairocoder01:
- When saving a name filter and going to edit, there is no remove icon in the tag at the top of the filter modal (though it does now populate the name filter correctly). Same for comm-channels. These may have been existing bugs comparing to a site without this updated. So I think we might be able to leave it as-is and call this PR complete
image

cairocoder01:
@corsacca Ready for your review

cairocoder01 (approved):

corsacca:
Hey Guys!

Nice work on these upgrades!

## Modal
If i add a location and then remove it, it does not get removed from the added filters:
image

Same for Assigned to, connection fields (sometimes).

Got 
 Warning: Undefined array key "icon" in /[path]/wp-content/themes/disciple-tools-theme/archive-template.php on line 1036

## Field picker
- "reset to default" does not work
- Dropdown is cut off. (tho i notice this is an existing issue on the bulk edit view)
We can remove "Add fields" since it is add and remove now.
image

## Other notes
1. Sanitization scope regression in save_user_filter — (wp-content/themes/disciple-tools-theme/dt-users/users.php#L766-L798)
Replacing deprecated FILTER_SANITIZE_STRING was needed, but the old call recursively tag-stripped every string in the structure (I verified this on PHP 8.4 — nested query, labels, and unknown keys all got stripped). The new code sanitizes only name/type/tab/ID and labels' name/field/id; the entire query subtree and any other keys (labels[].date, arbitrary client-added keys) are now stored raw in user meta. Current render paths escape output (escapeHTML), so I found no direct XSS, but it's a real defense-in-depth loss. dt_recursive_sanitize_array() — already used a few hundred lines up in (wp-content/themes/disciple-tools-theme/archive-template.php#L483) — would preserve the old guarantee with less code.

brady-lamansky-gtt:
@corsacca Thank you for the review!

For removing location/connection custom filters still showing the labels:
- I made a previous commit to the web-components to fix this issue, adding `bubbles: true` to the change events, since they weren't being received on _remove(). This worked for me when I updated my local web components! Let me know if anything else is needed: https://github.com/DiscipleTools/disciple-tools-web-components/commit/913c0a9481b1d2ab5e923778a01e291ec7be026f

For sanitization, I updated it originally because the sanitization in `filter_var_array( $filter, FILTER_SANITIZE_STRING );` was stripping special characters, like the '<' character, out of '<19' for the Age filter (and the other Age options as well). If I save an "Age <19" filter, I see the correct records. When I refresh the page, no records are displayed. "No records found matching your filter." The same happens with dt_recursive_sanitize_array().
- I went with Gemini's recommended fix back then, but it sounds like that's not what we want. How do you recommend we handle this? Gemini said this, but bypassing sanitization for specific pieces may not be what we want: "To fix both issues, I ran your dt_recursive_sanitize_array() function against the whole payload to secure arbitrary keys, but explicitly bypassed tag-stripping for the query and labels[id] system arrays, enforcing scalar types and unslashing instead." What do you think?

I fixed the other issues and will push the changes once sanitization is figured out!

corsacca:
Thanks @brady-lamansky-gtt!
I regret that we allowed `<` in the keys for the age field. It has caused headaches and would be hard to migrate from. 

Did some back and forth with claude and i think we can update save_user_filter as so:

Proposed replacement:

```
public static function save_user_filter( $filter, $post_type ){
    $current_user_id = get_current_user_id();
    if ( $current_user_id && isset( $filter['ID'] ) ){

        // Filter values are structured query data, not HTML: values like the '<19' age
        // key must survive. SQL safety comes from prepare/esc_sql at query time and XSS
        // safety from escaping at render, so only byte-validity is enforced here.
        $filter = map_deep( $filter, 'wp_check_invalid_utf8' );
        $filter['ID']   = is_scalar( $filter['ID'] ) ? (string) $filter['ID'] : '';
        $filter['type'] = sanitize_key( $filter['type'] ?? '' );
        $filter['tab']  = sanitize_key( $filter['tab'] ?? '' );
        if ( ! is_array( $filter['query'] ?? null ) ) {
            $filter['query'] = [];
        }

        $filters = get_user_option( 'saved_filters', $current_user_id );
        ...
```
`$filter = map_deep( $filter, 'wp_check_invalid_utf8' );` Removes anything that is not UTF8, so "<19" will be kept as kept in the filter key and label.

cairocoder01:
> For removing location/connection custom filters still showing the labels:
> 
> * I made a previous commit to the web-components to fix this issue, adding `bubbles: true` to the change events, since they weren't being received on _remove(). This worked for me when I updated my local web components! Let me know if anything else is needed: [DiscipleTools/disciple-tools-web-components@913c0a9](https://github.com/DiscipleTools/disciple-tools-web-components/commit/913c0a9481b1d2ab5e923778a01e291ec7be026f)

@brady-lamansky-gtt Since that component update has been released, you should update the component library version in `package.json` of the theme to v1.0.2 and `npm run build` to get those updates included in this branch.

brady-lamansky-gtt:
@corsacca @cairocoder01 thank you! I Just made those changes

corsacca:
@brady-lamansky-gtt, i'm not seeing the commit here. Try pushing again?

brady-lamansky-gtt:
> @brady-lamansky-gtt, i'm not seeing the commit here. Try pushing again?

Sorry I committed my changes but forgot to push!

cairocoder01 (approved):
This looks good to me. The 2 regressions that Claude found are fixed with the latest commit. Pending Claude finding anything more, I approve merging this.

corsacca:
Thanks Guys!

